PR #77, opened 2020-10-28
Create pr-build.yaml

wyTrivail:


.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,231 @@
...
+      with:
+        go-version: ^1.13
+
+    - uses: actions/checkout@v2
+
+    - name: Checkout dev opentelemetry-collector-contrib
+      if: github.event_name == 'repository_dispatch'
+      uses: actions/checkout@v2
+      with:
+        repository: ${{ github.repository_owner }}/opentelemetry-collector-contrib
```
mxiamxia, line 30: do we want to clean up the same step in CI workflow?

PR #89, opened 2020-11-03
enable people to build image with dockerfile directly

wyTrivail:
**Description:** 

Modify the dockerfile, so that people can directly use it to build image with building binary in front.

**Testing:** 

Tested on local

### DIFF
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -1,10 +1,44 @@
-FROM alpine:latest as certs
+################################
+#	Building Stage         #	
+#			       #	
+################################
+FROM alpine:latest as build 
+
+# update cert
 RUN apk --update add ca-certificates
 
+# install libs
+RUN apk add --update bash git make
+
+# config go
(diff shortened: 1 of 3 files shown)


anuraaga (approved):

cmd/awscollector/Dockerfile
```diff
@@ -1,10 +1,44 @@
...
+# copy artifacts
+WORKDIR /workspace
+COPY cmd  /workspace/cmd
+COPY config.yaml /workspace/config.yaml
+COPY pkg /workspace/pkg
+COPY tools /workspace/tools
+COPY go.mod /workspace/go.mod
+COPY go.sum /workspace/go.sum
+COPY VERSION /workspace/VERSION
+COPY .git /workspace/.git
```
mxiamxia, line 26: is it needed?

mxiamxia (approved):

cmd/awscollector/Dockerfile
wyTrivail: yes, it's needed, we need .git to get latest commit hash, version to get version

PR #95, opened 2020-11-05
changed prometheus exporter in readme

amanbrar1999:
**Description:**

Current reference to `prometheusexporter` is misleading, this exporter defines a Prometheus pull configuration, whereas we wish to support a Prometheus push configuration using the `awsprometheusremotewriteexporter`

This component will be added to the `opentelemetry-collector-contrib` repo [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/1497), before being moved into this repo

-- cc @alolita


mxiamxia:
If we have decided to not support the pull mode `prometheusexporter` in aws-otel-collector. Can you also remove the `prometheusexporter` component support in the code below in this PR?

https://github.com/aws-observability/aws-otel-collector/blob/main/pkg/defaultcomponents/defaults.go#L70

wyTrivail:
i guess we will support pulling mode in receiver instead?

amanbrar1999:
Yes, metrics will be received with a scrape/pull model, but we only plan to support exporting using a push model / remote write endpoints

amanbrar1999:
Not sure why some tests seemingly related to traces are failing after removing prometheus exporter. Any ideas @wyTrivail

amanbrar1999:
Will close this and update with the new exporter is brought in

PR #100, opened 2020-11-09
Fix docker shared creds

dradetsky:
**Description:** 

Since the docker-compose file in the example mounts the user's aws directory, I presume the intention was to allow the use of shared credentials in addition to environment credentials. In my own case, using shared credentials was desirable since we use aws sso so I've got a big session key as well as access & secret keys, and in any case I'd prefer that myself & other internal users have to do a lot of copy-pasting to test this out.

However, shared credentials didn't work. I eventually realized that because the collector was a golang from-scratch image, `$HOME` was undefined, and I was able to fix loading shared credentials by adding

```
      - HOME=/root
```

to my own compose file. However, it seems like it would be better to just define that in the dockerfile, so that's what I've done here.

I've also added a candidate rewrite of the example compose file. It's not required; could be cherry-picked out.

**Testing:** I have not tested it in any way

**Documentation:** None besides the comments

### DIFF
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -43,6 +43,7 @@ COPY --from=build /workspace/build/linux/aoc_linux_x86_64 /awscollector
 COPY --from=build /workspace/config.yaml /etc/otel-config.yaml
 
 ENV RUN_IN_CONTAINER="True"
+ENV HOME=/root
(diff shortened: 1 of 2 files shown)


mxiamxia (approved):
Thanks for catching it and made the fix. We'll update the doc.

PR #132, opened 2020-11-18
Update CI.yml

wyTrivail:
**Description:**

remove the dependency of candidate release and integration test so that we can test soaking test,
will add it back once the soaking test debugging is finished


shaochengwang (approved):
LGTM!

PR #153, opened 2020-11-22
Add commit id as a dimension to soaking metric

wyTrivail:
**Description:** 

so that we can track a series of soaking test using commit id


shaochengwang (approved):
LGTM!

PR #173, opened 2020-11-25
Fix Performance model template

kohrapha:
**Description:** 
Fix broken Markdown template for Performance model.


wyTrivail (approved):
thanks!

PR #198, opened 2020-12-01
Bump dynatrace exporter version

dyladan:
During testing it was discovered that the timestamp format used by the dynatrace exporter was incorrect in a previous version.


wyTrivail:
could you take a look, seems the building was failing.

dyladan:
Looks like the whole dep tree was already bumped to 0.16 anyways

PR #227, opened 2020-12-08
use dispatch instead of listening on tagging

wyTrivail:
to decouple the release logic with the actual release code, so that the release logic will always be in main branch, otherwise, the release logic will come from the tag.  if there's anything wrong in the tag for the release logic, we had to retag, and run CI to upload candidate packages again, which is time-consuming


mxiamxia (approved):
/lgtm

PR #239, opened 2020-12-09
fix the ecr public pushing

wyTrivail:
1. the region needs to be hard-coded to us-east-1
2. aws-cli 2.0 is needed for ecr-public subcommand


mxiamxia (approved):
Thanks

PR #242, opened 2020-12-10
Add build and push workflow for aws-observability public ecr images

amanbrar1999:
**Description:** 
This is a generic workflow for manually triggering builds and pushes for ECR public aws-observability. 

For whatever images we want to push, we can specify the github repo where the code exists, the ecr repo within the aws-observability ECR public project, the version we want to deploy and the path to the dockerfile from within the github repo. 

Assumptions:
- This will eventually be moved into the unified build repo
- The ECR public repo has already been previously set up (guidelines will be made more clear when eventually moved into unified build repo)
- The code we want to build and push is open source

### DIFF
--- a/.github/workflows/build-and-push-public-ecr.yaml
+++ b/.github/workflows/build-and-push-public-ecr.yaml
@@ -74,7 +74,7 @@ jobs:
         with:
           aws-access-key-id: ${{ secrets[env.KEY_ID] }}
           aws-secret-access-key: ${{ secrets[env.SECRET_ID] }}
-          aws-region: us-east-1
+          aws-region: us-west-2
       
       - name: Clone Image Code Repo
         uses: actions/checkout@master
--- a/.github/workflows/build-and-push-public-ecr.yaml
+++ b/.github/workflows/build-and-push-public-ecr.yaml
@@ -37,11 +37,25 @@ jobs:
   build-and-push-image:
     runs-on: ubuntu-latest
     steps:
+      - name: Validate Inputs
+        run: |
+          if [[ ${{ github.event.inputs.ecrRepo }} != ${{ github.event.inputs.githubRepo }} ]]
+          then
+            echo "ecr repository name (${{ github.event.inputs.ecrRepo }}) must match github repository name (${{ github.event.inputs.githubRepo }})"
+            exit 1
+          fi
+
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-push-public-ecr.yaml
```diff
@@ -0,0 +1,78 @@
...
+env:
+  ECR_REPO: aws-observability/${{ github.event.inputs.ecrRepo }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create Secret Names From ECR Repo Name
+        run: |
+          NEW_ECR_ID=`echo "${{ github.event.inputs.ecrRepo }}" | sed "s/-/_/g" | sed "s/\(.*\)/\U\1/"`
```
wyTrivail, line 42: can we add a check in case NEW_ECR_ID is a empty string, which would set the name to RELEASE_KEY_ID, which is the default credential

.github/workflows/build-and-push-public-ecr.yaml
```diff
@@ -0,0 +1,78 @@
...
+      pathToDockerfile:
+        description: 'the path within the repository to the folder containing the dockerfile'
+        required: true
+        default: "."
+  
+env:
+  ECR_REPO: aws-observability/${{ github.event.inputs.ecrRepo }}
+
+jobs:
+  build-and-push-image:
```
wyTrivail, line 37: do we check if the githubrepo name is same as the ecrrepo name?

wyTrivail (changes requested):

.github/workflows/build-and-push-public-ecr.yaml
amanbrar1999: I will confirm with Rohan first whether this is the case for him, if so I will add this in

.github/workflows/build-and-push-public-ecr.yaml
amanbrar1999: Will do

amanbrar1999:
closing to avoid spam while testing

amanbrar1999:
Comments have been addressed

wyTrivail (approved):

PR #244, opened 2020-12-11
Add Getting Started Templates for the Prometheus Pipeline

JasonXZLiu:
**Description:**

Adding getting started templates for the Prometheus pipeline. It consists of a sample app deployment and an AOC deployment (as a DaemonSet). AOC will use the Prometheus Receiver and AWS Prometheus Remote Write Exporter.

### DIFF
--- a/examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
+++ b/examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
@@ -49,5 +49,4 @@ spec:
     protocol: TCP
   selector:
     app: prometheus-sample-app
-  type: NodePort
 ---
--- a/examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
+++ b/examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
@@ -26,7 +26,8 @@ spec:
     spec:
       containers:
       - name: prometheus-sample-app
-        image: {{PUBLIC_SAMPLE_APP_IMAGE}}
+        image: "" # change to your local image
+        command: ["/bin/main", "-listen_address=0.0.0.0:8080", "-metric_count=10"]
         ports:
         - name: web
           containerPort: 8080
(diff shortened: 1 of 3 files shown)


examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
```diff
@@ -0,0 +1,53 @@
...
+    matchLabels:
+      app: prometheus-sample-app
+  template:
+    metadata:
+      labels:
+        app: prometheus-sample-app
+    spec:
+      containers:
+      - name: prometheus-sample-app
+        image: "" # change to your local image
```
wyTrivail, line 29: can't it be the sample-app we are going to push?

examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
```diff
@@ -0,0 +1,53 @@
...
+    scrape: "true"
+spec:
+  ports:
+  - name: web
+    port: 8080
+    targetPort: 8080
+    protocol: TCP
+  selector:
+    app: prometheus-sample-app
+  type: NodePort
```
wyTrivail, line 52: why we need it to be nodeport?

examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
JasonXZLiu: We actually realized we don't have enough time to push the sample app. We'll have to build and push it locally.

examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
JasonXZLiu: Hmm, we don't actually need a nodeport. We just need to expose the application's port so that the receiver can auto-discover and scrape the endpoint. I believe it's just the easiest way for us to do this.

wyTrivail: is this the recommended way?

wyTrivail (approved):

examples/eks/prometheus-pipeline/prometheus-sample-app.yaml
JasonXZLiu: Hmm, looks like Kubernetes uses ClusterIP by default. I'll remove the NodePort

PR #246, opened 2020-12-11
chore: Add gpg linux package signing and RPM FIPS support

haojhcwa:
**Description:** 
Adding GPG package signing for Linux packages. For rpm packages, rpmsign is used, where as for deb packages, dpkg-sig is used. Both use the same GPG key for signing, and the public key can be downloaded from: https://aws-otel-collector.s3.amazonaws.com/aws-otel-collector.gpg

**Testing:**
The signing workflow has been tested on my own fork using my own key pair: https://github.com/haojhcwa/aws-otel-collector/actions/runs/414275830

**Documentation:**
There is no usage change for the Linux packages, how to use the public key to verify the packages would come later in a different PR.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -205,7 +205,7 @@ jobs:
           gpg --list-secret-keys
           shred -fuvz pkg_sign_private.key
 
-      - name: Sign RPM Pakcage
+      - name: Sign DEB Pakcage
         run: |
           sudo apt install -y dpkg-sig
           dpkg-sig --sign origin -k "aws-otel-collector@amazon.com" build/packages/debian/*/*.deb
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,17 +26,18 @@ on:
 env:
   IMAGE_NAME: aws-otel-collector
   PACKAGING_ROOT: build/packages
-  ECR_REPO: aws/aws-otel-collector  
+  ECR_REPO: aws/aws-otel-collector
   TF_VAR_aws_access_key_id: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
-  TF_VAR_aws_secret_access_key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }} 
+  TF_VAR_aws_secret_access_key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
+  PKG_SIGN_PRIVATE_KEY_NAME: aoc-linux-pkg-signing-gpg-key
 jobs:
 
- 
+
   build:
     runs-on: ubuntu-latest
 
     steps:
-    # Set up building environment, patch the dev repo code on dispatch events.  
+    # Set up building environment, patch the dev repo code on dispatch events.
     - name: Set up Go 1.x
       uses: actions/setup-go@v2
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -139,6 +140,33 @@ jobs:
           ARCH=x86_64 DEST=build/packages/linux/amd64 tools/packaging/linux/create_rpm.sh
           ARCH=aarch64 DEST=build/packages/linux/arm64 tools/packaging/linux/create_rpm.sh
 
+      - name: Download Package Signing GPG key
```
wyTrivail, line 143: shall we sign deb with gpg in the near future? if yes, it will be better to put signing as a separate job.

tools/packaging/linux/create_rpm.sh
```diff
@@ -77,7 +77,7 @@ mv ${RPM_NAME}-${VERSION}.tar.gz ${BUILD_ROOT}/SOURCES/
 rm -rf ${WORK_DIR}
 
 echo "Creating the rpm package"
-rpmbuild --define "VERSION $VERSION" --define "RPM_NAME $RPM_NAME" --define "_topdir ${BUILD_ROOT}" -bb -v --clean ${SPEC_FILE} --target ${ARCH}-linux
+rpmbuild --define "VERSION $VERSION" --define "RPM_NAME $RPM_NAME" --define "_topdir ${BUILD_ROOT}" --define "_source_filedigest_algorithm 8" --define "_binary_filedigest_algorithm 8" -bb -v --clean ${SPEC_FILE} --target ${ARCH}-linux
```
wyTrivail, line 80: if customers use this script to build rpm without signing, will the rpm be able to be installed ?

.github/workflows/CI.yml
```diff
@@ -163,6 +191,30 @@ jobs:
...
+          gpg --import pkg_sign_private.key
+          gpg --list-secret-keys
+          shred -fuvz pkg_sign_private.key
+
+      - name: Sign RPM Pakcage
+        run: |
+          sudo apt install -y dpkg-sig
+          dpkg-sig --sign origin -k "aws-otel-collector@amazon.com" build/packages/debian/*/*.deb
+
+      - name: Remove Package Signing GPG Key from local GPG Key Ring
```
wyTrivail, line 213: why do we need this step? we don't want to leak it i guess?

tools/packaging/linux/create_rpm.sh
haojhcwa: Yes, this is not related to signing itself, but related to RHEL 8.2 FIPS check. Customer can still build and run their own RPM without any issue.

.github/workflows/CI.yml
haojhcwa: Yes, this cleanly remove any possible secret left on the running device disk and memory in case the content is not securely wiped after the run.

.github/workflows/CI.yml
haojhcwa: The reason not to have the GPG signing key operations as a separate job is to minimize the locations where the key has been stored. If we add it as a separate job, we have to copy the key to cache and recover it from a different job, which added another place where the key had been stored, adding risk of leaking the signing key.

.github/workflows/CI.yml
```diff
@@ -163,6 +191,30 @@ jobs:
...
+          AWS_ACCESS_KEY_ID: ${{ secrets.SIGN_PKG_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.SIGN_PKG_AWS_SECRET_ACCESS_KEY }}
+
+      - name: Import Package Signing GPG Key
+        run: |
+          gpg --import pkg_sign_private.key
+          gpg --list-secret-keys
+          shred -fuvz pkg_sign_private.key
+
+      - name: Sign RPM Pakcage
```
mxiamxia, line 208: Typo- `Sign Deb Package` ?

wyTrivail, line 208: sign deb?

.github/workflows/CI.yml
```diff
@@ -163,6 +191,30 @@ jobs:
...
+          md5sum pkg_sign_private.key
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.SIGN_PKG_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.SIGN_PKG_AWS_SECRET_ACCESS_KEY }}
+
+      - name: Import Package Signing GPG Key
+        run: |
+          gpg --import pkg_sign_private.key
+          gpg --list-secret-keys
+          shred -fuvz pkg_sign_private.key
```
wyTrivail, line 206: the import step is different for signing deb and rpm?

.github/workflows/CI.yml
wyTrivail: can't we get the key in that separate job?

.github/workflows/CI.yml
haojhcwa: Yes, RPM maintains its own public key internally, while using GPG for private keys, so for rpm, the private key needs to be imported to gpg, then export the public key to be imported to RPM. Whereas dpkg-sig use gpg directly, so no other fiddling is needed.

.github/workflows/CI.yml
haojhcwa: Good catch, fixed.

haojhcwa: Good catch, fixed.

wyTrivail (approved):

.github/workflows/CI.yml
haojhcwa: Yes, that's not a bad idea, I'll look into it to see how this could be optimized, especially when signing packages for other platforms are considered.

PR #252, opened 2020-12-11
Add CodeQL Security Scanning Workflow for AOC

JasonXZLiu:
**Description:** 

This PR adds basic CodeQL scanning into the CI and PR-Build workflows to perform security scanning on the AWS-OTel-Collector. CodeQL will create alerts based on it's analysis which will be found in the security tab: https://github.com/aws-observability/aws-otel-collector/security.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -99,18 +99,13 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
-      matrix:
-        language: [ 'go' ]
-    
     steps:
       - uses: actions/checkout@v2
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v1
         with:
-          languages: ${{ matrix.language }}
+          languages: "go"
         
       - name: Download built artifacts
         uses: actions/download-artifact@v2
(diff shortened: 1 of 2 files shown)


alolita (approved):
+1

amanbrar1999 (approved):

.github/workflows/CI.yml
```diff
@@ -94,7 +93,32 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v2
 
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: "go"
+        
+      - name: Download built artifacts
```
wyTrivail, line 110: Why do we need to download the artifacts?

.github/workflows/PR-build.yml
```diff
@@ -61,6 +61,31 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: "go"
+        
+      - name: Download built artifacts
```
wyTrivail, line 77: the same question, i thought it's just a source code scan?

.github/workflows/CI.yml
JasonXZLiu: Hmm, I think by default, some languages must be built and compiled for the code scanning (more specifically for languages like Java and C++). I wasn't sure if we needed it for Golang (but I haven't found any documentation on it). I'll try testing without it and seeing if the code scanning still runs

JasonXZLiu: Seems like we actually don't need it. I'll remove this and push again.

wyTrivail (approved):

PR #264, opened 2020-12-14
Change Logging File Permission from 0755 to 0750

JasonXZLiu:
**Description:** 

Change log file permissions from 0755 to 0750. This will remove read and execute requests for "others".


mxiamxia (approved):
Thanks!

PR #266, opened 2020-12-15
add AWS profile customization support in extra config file

mxiamxia:
**Description:**
add AWS profile customization support in extra config file

**Link to tracking Issue:** 
N/A

**Testing:** 
Unit Test


hdj630 (approved):

pkg/extraconfig/extraconfig.go
```diff
@@ -64,8 +67,16 @@ func GetExtraConfig() (*ExtraConfig, error) {
...
 	// build extraconfig object
 	extraConfig := ExtraConfig{}
-	if val, ok := extraConfigMap["loggingLevel"]; ok {
-		extraConfig.LoggingLevel = val
+	for key, val := range extraConfigMap {
+		switch key {
+		case "loggingLevel":
+			extraConfig.LoggingLevel = val
+		case "awsProfile":
+			os.Setenv("AWS_PROFILE", val)
```
wyTrivail, line 75: i feel we need to define the vars in extraconfig instead of set it as env var here. we can get it in the upper layer and set it as env var.

mxiamxia: I had the same thought but the style will be different with `loggingLevel` and we had the release for this change yesterday

PR #273, opened 2020-12-16
Update Collector to v0.17.0

tigrannajaryan:
- Updated all Collector module references to v0.17.0.
- Removed k8sclusterreceiver reference from go.mod since it is no longer a required
  dependency.
- Run `go mod tidy` in the root.

Tested with `make build`.


mxiamxia (approved):
Thanks!

PR #275, opened 2020-12-17
Fix getting started template with IRSA

JasonXZLiu:
**Description:**

Change the service account to be the one created when setting up the IAM Roles for service accounts tutorial. This IAM Role will be attached with the proper permissions needed.

### DIFF
--- a/e2etest/get-performance-model-table.py
+++ b/e2etest/get-performance-model-table.py
@@ -60,7 +60,7 @@ def flatten_performance_models(models):
 if __name__ == "__main__":
     from jinja2 import Environment, PackageLoader, select_autoescape
     templateLoader = jinja2.FileSystemLoader(searchpath="e2etest/templates/")
-    env = Environment(autoescape=select_autoescape(['html', 'xml', 'tpl']), default_for_string=True, loader=templateLoader)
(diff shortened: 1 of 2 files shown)


examples/eks/prometheus-pipeline/eks-prometheus-sidecar.yaml
```diff
@@ -71,8 +71,10 @@ data:
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: adotcol-admin
+  name: amp-iamproxy-ingest-service-account
   namespace: adot-col
+  annotations:
+    eks.amazonaws.com/role-arn: arn:aws:iam::YOUR_ACCOUNT_ID:role/amp-iamproxy-ingest-role
```
mxiamxia, line 77: Can we use `<>` for the variables in the template?

mxiamxia:
We named the example file `examples/eks/prometheus-pipeline/eks-prometheus-sidecar.yaml`. But in the template,  we deploy the collector as service from DaemonSet mode which is not fitting the meaning of the file name. This example seems misleading.

examples/eks/prometheus-pipeline/eks-prometheus-sidecar.yaml
JasonXZLiu: I'll change that throughout our template

mxiamxia (approved):
Thanks!

PR #278, opened 2020-12-21
Add ECS template configurations for prometheus pipeline

JasonXZLiu:
**Description:**

Add ECS template configurations for prometheus pipeline. This will be used in our getting started documentation. 

For both EC2 and Fargate, we include:

* ADOT Collector configurations
* Task definitions for the ADOT Collector and sample app


amanbrar1999 (approved):

mxiamxia (commented):
Do we have the user guide docs for running collector with Prometheus in this example?

mxiamxia (approved):
Thanks

PR #280, opened 2020-12-22
Fix an issue for the ECS template

JohnWu20:
**Description:** 
Changed the sample application essential value from true to false both in the task definition and cloudformation templates.

https://github.com/aws-otel/aws-otel.github.io/issues/52

**Testing:** 
Tested in local.


mxiamxia (commented):
Any reason why sample app stopped? Wonder why we can't stop the collector if sample app is shutdown?

JohnWu20:
> Any reason why sample app stopped? Wonder why we can't stop the collector if sample app is shutdown?

The sample app will generate trace data when there is any http request, I wrote a loop to generate the http request. After the loop ends, the application stoped.

JohnWu20:
> Any reason why sample app stopped? Wonder why we can't stop the collector if sample app is shutdown?

After discussing with our team, we decided to update the sample image. I will cancel this ptr. Thanks!

PR #286, opened 2021-01-07
Exit 0 when configuration is invalid

hdj630:
**Description:** 
Exit 0 when configuration is invalid.

### DIFF
--- a/pkg/config/config_factory.go
+++ b/pkg/config/config_factory.go
@@ -18,8 +18,10 @@ package config
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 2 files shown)


mxiamxia (approved):
Thanks

anuraaga:
@mxiamxia Why do we want to exit 0? If it's a failure, we want to return a failure status code right?

mxiamxia:
> @mxiamxia Why do we want to exit 0? If it's a failure, we want to return a failure status code right?

@anuraaga , Sorry I missed this one :( .

PR #324, opened 2021-01-28
Update the template to make the other containers not essensial

JohnWu20:
**Description:** 
Update the sample containers to mark them as no-essensial, because the stopped containers shouldn't shut down the task.

### DIFF
--- a/deployment-template/ecs/aws-otel-ec2-sidecar-deployment-cfn.yaml
+++ b/deployment-template/ecs/aws-otel-ec2-sidecar-deployment-cfn.yaml
@@ -93,12 +93,13 @@ Resources:
               condition: START
           image: >-
             611364707713.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-otel-goxray-sample-app:v0.0.1
-          essential: true
+          essential: false
           links:
             - aws-collector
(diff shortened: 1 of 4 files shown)


deployment-template/ecs/aws-otel-ec2-sidecar-deployment-cfn.yaml
```diff
@@ -93,12 +93,13 @@ Resources:
               condition: START
           image: >-
             611364707713.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-otel-goxray-sample-app:v0.0.1
-          essential: true
+          essential: false
```
mxiamxia, line 96: In this template/example, we want to run otel collector together with sample apps. is it our expectation to put the sample app `essential` to false?

JohnWu20: One of the reason is that the container receiver should be able to collect the stopped container's metadata. We don't want to make an example that when the customer use AOC, they should make their container essential, because even one of the containers dead, the task shouldn't be dead.

mxiamxia (approved):
Thanks

PR #325, opened 2021-01-28
Test the ecs container receiver v0.19.0

JohnWu20:
**Description:** 
Do another test for the new version 0.19.0


wyTrivail:
How do we plan to do the test? Would it be possible that we can add the test cases into integration-test instead of manually testing it so that we don't lose track of it?

JohnWu20:
> How do we plan to do the test? Would it be possible that we can add the test cases into integration-test instead of manually testing it so that we don't lose track of it?

Yes, I think that's a good point. We should include the test cases into integration test.

PR #331, opened 2021-01-30
increase the workflow jobs parallel number to 10

mxiamxia:
**Description:** 
increase the workflow jobs parallel number to 10

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -133,7 +133,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [get-testing-suites, release-to-s3]
     strategy:
-      max-parallel: 5
+      max-parallel: 10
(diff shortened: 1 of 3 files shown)


wyTrivail (commented):
why would increasing parallelism help?

mxiamxia:
> why would increasing parallelism help?

I may not help the fix or worsen the instability issue form GithubAction, but I want to try to test it out. Do you think we should merge it after enhance the VPC and subnet IP limits? @wyTrivail

wyTrivail:
> > why would increasing parallelism help?
> 
> I may not help the fix or worsen the instability issue form GithubAction, but I want to try to test it out. Do you think we should merge it after enhance the VPC and subnet IP limits? @wyTrivail

yes, probably wait a bit for the large vpc setting up.

wyTrivail (approved):
let's see...

PR #332, opened 2021-01-30
Fix CD workflow validation

wyTrivail:
change -ne to !- as -ne expect number instead of string

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -60,8 +60,8 @@ jobs:
           # check commit sha and version
           version_in_candidate=`cat build/packages/VERSION`
           sha_in_candidate=`cat build/packages/GITHUB_SHA`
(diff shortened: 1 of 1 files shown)


.github/workflows/CD.yml
```diff
@@ -60,8 +60,8 @@ jobs:
           # check commit sha and version
           version_in_candidate=`cat build/packages/VERSION`
           sha_in_candidate=`cat build/packages/GITHUB_SHA`
-          [ $version_in_candidate -ne ${{ github.event.inputs.version }} ] && echo "wrong version is detected: $version_in_candidate != ${{ github.event.inputs.version }}" && exit 1
-          [ $sha_in_candidate -ne ${{ github.event.inputs.sha }} ] && echo "wrong sha is detected: $sha_in_candidate != ${{ github.event.inputs.sha }}" && exit 1
```
mxiamxia, line 64: any reason for this change? `-ne` works from the testing

mxiamxia (approved):

PR #338, opened 2021-02-04
Return config read failure to collector entrypoint instead of force exiting.

anuraaga:
Fixes bad status code on config error

**Link to tracking Issue:** #340 

### DIFF
--- a/pkg/config/config_factory.go
+++ b/pkg/config/config_factory.go
@@ -43,8 +43,7 @@ func GetCfgFactory() func(otelViper *viper.Viper, cmd *cobra.Command, f componen
 		// use OTel yaml config from input
 		otelCfg, err := service.FileLoaderConfigFactory(otelViper, cmd, f)
 		if err != nil {
-			log.Printf("Config file is missing or invalid, %s", err)
-			os.Exit(0)
+			log.Fatalf("Config file is missing or invalid: %w", err)
 		}
 		return otelCfg, nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/config/config_factory.go
```diff
@@ -43,8 +43,7 @@ func GetCfgFactory() func(otelViper *viper.Viper, cmd *cobra.Command, f componen
 		// use OTel yaml config from input
 		otelCfg, err := service.FileLoaderConfigFactory(otelViper, cmd, f)
 		if err != nil {
-			log.Printf("Config file is missing or invalid, %s", err)
-			os.Exit(0)
+			log.Fatalf("Config file is missing or invalid: %w", err)
```
pingleig, line 46: well ... just found this PR after spent one afternoon wondering why there is no error when the collector is not starting .. It should fix #340  though I think we should just return the error directly, which is the behaviour of otel and otelconrtib. The error will get printed by caller.

```go
return service.FileLoaderConfigFactory(otelViper, cmd, f)
```

anuraaga: Thanks for the suggestion, sounds good

mxiamxia: We probably have to keep `Exit(0)` for fixing bug in windows and Linux with Upstart service. I tried to leave the reason in this https://github.com/aws-observability/aws-otel-collector/pull/286#issuecomment-769512414. @anuraaga

pingleig: I think we should try to fix the control script instead of returning 0 on error, and maybe we should publish a new release because it is causing more confusion like #350

pingleig:
Created https://github.com/aws/amazon-cloudwatch-agent/issues/205 to track the long term the fix on composite agent. Currently, it should return 1 when it's not running on windows (i.e. Linux EC2 or containers).

PR #366, opened 2021-02-17
[WIP] Add default config for prometheus

bjrara:
**Why do we need it?**
1. Add default configuration for aws-otel-collector using prometheus receiver.
2. Inject attribute `ClusterName` from environment variable `OTEL_RESOURCE` and `TaskId` from `service.name` for emf exporter to use in metric declarations.

### DIFF
--- a/cmd/awscollector-prometheus/Dockerfile
+++ b/cmd/awscollector-prometheus/Dockerfile
@@ -1,0 +1,54 @@
+################################
+#	Building Stage         #	
+#			       #	
+################################
+FROM alpine:latest as build 
+
+# update cert
+RUN apk --update add ca-certificates
+
+# install libs
+RUN apk add --update bash git make build-base
+
+# config go
+COPY --from=golang:1.15.3-alpine /usr/local/go/ /usr/local/go/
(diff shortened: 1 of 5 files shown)


cmd/awscollector-prometheus/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+################################
+#	Final Stage            #	
+#			       #	
+################################
+FROM scratch
+
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=build /workspace/build/linux/aoc_linux_x86_64 /awscollector
+COPY --from=build /workspace/config-prometheus.yaml /etc/otel-config.yaml
+COPY --from=build /workspace/config/ecs/container-insights/otel-task-metrics-config.yaml /etc/ecs/container-insights/otel-task-metrics-config.yaml
```
mxiamxia, line 45: Do we need to create a new Docker image for Prometheus support only? Why can't we just add the prometheus default config into the exiting Dockerfile?

PR #377, opened 2021-02-25
Use Setuid/Setgid implemented in go1.16

vastin:
**Description:** 
Use syscall Setuid/Setgid to change user when built by go1.16 and later.
https://golang.org/doc/go1.16#minor_library_changes
On Linux, Setgid, Setuid, and related calls are now implemented. Previously, they returned an syscall.EOPNOTSUPP error. 

**Testing:**
local test

### DIFF
--- a/pkg/userutils/set_uid_gid_1_15.go
+++ b/pkg/userutils/set_uid_gid_1_15.go
@@ -1,0 +1,34 @@
+// +build linux
+// +build !go1.16
+
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package userutils
+
+import (
+	"github.com/opencontainers/runc/libcontainer/system"
+)
+
+// go1.15 and before: use Setgid/Setuid from 3rd party library.
+// go1.16 and later: use Setgid/Setuid implemented in go syscall.
+
+func setUid(uid int) (err error) {
+	return system.Setuid(uid)
+}
+
+func setGid(gid int) (err error) {
+	return system.Setgid(gid)
+}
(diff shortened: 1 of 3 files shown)


pkg/userutils/set_uid_gid_1_15.go
```diff
@@ -0,0 +1,34 @@
+// +build linux
```
mxiamxia, line 1: `set_uid_gid_1_15.go ` -> `set_uid_gid_lower_go_ver.go ' or something?

anuraaga (commented):
I think we should just change this to 1.16

https://github.com/aws-observability/aws-otel-collector/blob/main/go.mod#L3

And say we don't support building below it (we plan to release the next version with 1.16 right?). Since this is a binary, not a library, I don't think there's any reason to support older go versions when we use a feature from a newer one.

pkg/userutils/set_uid_gid_1_15.go
vastin: It follows the file naming convention like:
https://github.com/aws/aws-sdk-go/blob/master/aws/url.go
https://github.com/aws/aws-sdk-go/blob/master/aws/url_1_7.go

There is another style in golang code. We could change to it if preferred.
https://github.com/golang/go/blob/master/src/sort/slice_go14.go
https://github.com/golang/go/blob/master/src/sort/slice_go18.go
https://github.com/golang/go/blob/master/src/sort/slice_go113.go

vastin:
@anuraaga [go directive](https://golang.org/ref/mod#go-mod-file-go) won't prevent developer to build with older golang compiler. Since go1.16 is just released, it is more convenient to keep older golang compiler working in a transition period.

mxiamxia (approved):
Thanks

anuraaga:
Yes but they can easily install the new version, which we also say is required in the docs. Instead we've added significant complexity to the code and made it harder to have reproducible builds. So I think this isn't a good change and recommend reverting. 

To continue to support older version of go it's better to just stick to the external library rather than adding spaghetti to the code since it still works. That would be ok too.

anuraaga:
@vastin @mxiamxia Will you be addressing this? I think it's obvious the problem here given even our CI isn't testing the new code

https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CI.yml#L48

While it could be updated to support testing on multiple go versions, for distributed binaries (not libraries), it's less complex to just require the latest version of Go for development, it's trivial to install a new version and a lot of work maintaining a branching build / code.

anuraaga:
Talked with @vastin offline and it sounds like we do have a build currently we have to maintain that's on an older Go - it should be updated relatively soon when we can remove the branching.

In the meantime, let's at least update our CI to use 1.16 so the new code is getting tested.

vastin:
Thank you @anuraaga ! The CI will pick up the latest golang release which is 1.16 now. So it should be good.

PR #393, opened 2021-03-11
Introduce resourcedetectionprocessor for cluster name injection

bjrara:
**Why do we need it?**
This PR adds resource detection processor to allow injecting metric attributes through environment variables.


config.yaml
```diff
@@ -22,6 +26,8 @@ processors:
 exporters:
   awsxray:
   awsemf:
+    resource_to_telemetry_conversion:
+      enabled: true
```
mxiamxia, line 30: we can't enable it in the master default config. It will generate lots of unwanted dimensions for customers. Should it be applied to Prometheus Config file in your case?

bjrara: `ClusterName` is inserted as a resource label by `resourcedetectionprocessor`. If `resource_to_telemetry_conversion` is configured as `false`, metrics will not contain any labels injected by `resourcedetectionprocessor`, and that will cause `ClusterName` missing from each grouped metric as well. Reckoning that ClusterName is a key dimension used by ContainerInsight dashboard, I'm afraid this option needs to be set to `true` not only in the case of Prometheus but any case that `ClusterName` is injected in this way.

mxiamxia: Yes, in AWS OTel Collector docker image, we'll build-in a few configuration files to support different features [here](https://github.com/aws-observability/aws-otel-collector/tree/main/config). So in this config folder, we'll create one config for `Prometheus`(in your case) and another one for `eks/containerinsights`. All these feature related configuration files, we can enable `resourcedetectionprocessor` and `resource_to_telemetry_conversion` per our needs. But for the default config file(the one you modified), we'll leave the minimal configuration w/ only AWS xray and emf enabled.

bjrara: Done

bjrara:
bump for more activity

anuraaga:
IIUC the PR title should be something like "Add resourcedetectionprocessor as a configurable collector".

config.yaml
```diff
@@ -13,6 +13,10 @@ receivers:
     transport: udp
 
 processors:
+  resourcedetection/ec2:
+    detectors: [ env ]
+    timeout: 2s
+    override: false
```
mxiamxia, line 19: can we remove it since it is not used by the pipeline?

mxiamxia (commented):
Please address the comments and I'll approve and merge. Thx!

bjrara:
> IIUC the PR title should be something like "Add resourcedetectionprocessor as a configurable collector".

Updated the title. Thanks for the suggestion.

config.yaml
bjrara: Removed.

anuraaga (approved):

mxiamxia (approved):
Thx

PR #394, opened 2021-03-11
Add default configurations for Container Insights on EKS with Prometheus

bjrara:
**Why do we need it?**
This is a follow-up PR of https://github.com/aws-observability/aws-otel-collector/pull/393.

This PR adds configurations for demo workloads including AppMesh, HAProxy, Nginx, Java/Jmx, Memcached, and Redis. It also contains a one-for-all config to collect multiple resources in the cluster.

Metrics should show up automatically from the out-of-box container insight dashboard if user follows the official document and deploys the workloads properly.

PR #420, opened 2021-03-31
Revert "Add default configurations for Container Insights on EKS with Prometheus"

mxiamxia:
Reverts aws-observability/aws-otel-collector#394


bjrara (approved):
Thank!

PR #423, opened 2021-04-02
Support the partner components from upstream

mxiamxia:
**Description:**
What's new
- zipkinreceiver
- jaegerreceiver
- logzioexporter

**Link to tracking Issue:** 
https://github.com/aws-observability/aws-otel-collector/issues/292

**Testing:** 
https://github.com/aws-observability/aws-otel-test-framework/pull/261

PR #426, opened 2021-04-07
Remove duplicated test case

bjrara:
This PR removes an duplicated case that was mistakenly added in https://github.com/aws-observability/aws-otel-collector/pull/423


pingleig (approved):
I think we don't have /assign bot ... I am not sure about the policy of using bot though ...

PR #430, opened 2021-04-08
Add more regex to match new haproxy backend format

bjrara:
Same issue as https://github.com/aws/amazon-cloudwatch-agent/pull/197


mxiamxia (approved):
Thanks!

PR #433, opened 2021-04-09
Add an event type to trigger CI workflow

vastin:
**Description:**
Allow trigger CI workflow via repository_dispatch.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,7 +22,7 @@ on:
 
   # from collector and contrib repo
   repository_dispatch:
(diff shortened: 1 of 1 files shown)


pingleig (commented):
Actually I am not very familiar with the logic here, so we want to avoid doing go mod replace unless it's dependency build?

vastin:
> 
> 
> Actually I am not very familiar with the logic here, so we want to avoid doing go mod replace unless it's dependency build?
Yes. I want to add a way to trigger CI workflow for testing without dependency build.

PR #436, opened 2021-04-12
Add SSM build and speed up CI re-run

vastin:
**Description:**
Create and test ADOT Collector SSM package in CI workflow.
Make CI re-run faster by using cache hit flag to skip successful jobs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -98,6 +98,13 @@ jobs:
         cat go.mod
         ls pkg
 
+    - name: Cache binaries
+      id: cached_binaries
+      uses: actions/cache@v2
+      with:
+        key: "cached_binaries_${{ github.run_id }}"
+        path: build
+
     # Unit Test and attach test coverage badge
     - name: Unit Test
       run: make test
@@ -109,13 +116,8 @@ jobs:
 
     # Build and archive binaries into cache.
     - name: Build Binaries
+      if: steps.cached_binaries.outputs.cache-hit != 'true'
       run: make build
-
-    - name: Cache binaries
-      uses: actions/cache@v2
-      with:
-        key: "cached_binaries_${{ github.run_id }}"
-        path: build
 
     # upload the binaries to artifact as well because cache@v2 hasn't support windows
     - name: Upload
@@ -214,7 +216,15 @@ jobs:
       # Build and archive rpms into cache.
       - uses: actions/checkout@v2
 
+      - name: Cache rpms
+        id: cached_rpms
+        uses: actions/cache@v2
+        with:
+          key: "cached_rpms_${{ github.run_id }}"
+          path: build/packages
+
       - name: restore cached binaries
+        if: steps.cached_rpms.outputs.cache-hit != 'true'
         uses: actions/cache@v2
         with:
           key: "cached_binaries_${{ github.run_id }}"
@@ -224,6 +234,7 @@ jobs:
         run: ls -R
 
       - name: Build RPM
+        if: steps.cached_rpms.outputs.cache-hit != 'true'
         run: |
           ARCH=x86_64 DEST=build/packages/linux/amd64 tools/packaging/linux/create_rpm.sh
           ARCH=aarch64 DEST=build/packages/linux/arm64 tools/packaging/linux/create_rpm.sh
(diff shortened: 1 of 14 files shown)


.github/workflows/CI.yml
```diff
@@ -458,242 +547,326 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
 
+      - name: Cache if success
+        id: e2etest-ec2-2
+        uses: actions/cache@v2
+        with:
+          path: |
+            VERSION
+          key: e2etest-ec2-2-${{ github.run_id }}-${{ matrix.testcase }}-${{ matrix.testing_ami }}
```
pingleig, line 617: So if the cache does not exists, it only got created when the job is success? 

https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action

.github/workflows/CI.yml
```diff
@@ -360,40 +434,55 @@ jobs:
...
           key: "cached_packages_${{ github.run_id }}"
 
+      - name: Create SSM package
+        if: steps.e2etest-release.outputs.cache-hit != 'true'
+        run: |
+          ssm_pkg_version=`cat build/packages/VERSION`
+          ssm_pkg_version="${ssm_pkg_version:1}"
+          python3 tools/ssm/ssm_manifest.py ${ssm_pkg_version}
+          aws s3 cp build/packages/ssm s3://aws-otel-collector-test/ssm --recursive
+          bash tools/ssm/ssm_create.sh testAWSDistroOTel-Collector ${ssm_pkg_version} aws-otel-collector-test/ssm
```
pingleig, line 474: We might need to clean up the document though, the default quota for ssm accounts is not very high. Or we can remove the test ssm document if all the ec2 test has passed.

tools/ssm/linux-amd64-deb/install.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
pingleig, line 1: Use `#!/usr/bin/env bash` https://github.com/aws/amazon-cloudwatch-agent/blob/53f8c7cd6c2e1977bc45ea55f70298b0d7dce942/Tools/src/create_darwin.sh#L1 or `sh` https://github.com/aws/amazon-cloudwatch-agent/blob/53f8c7cd6c2e1977bc45ea55f70298b0d7dce942/packaging/dependencies/amazon-cloudwatch-agent-ctl#L1

Though I think all of them should work on most vms and bms.

.github/workflows/CI.yml
vastin: Yes, it is only created when job return success.

.github/workflows/CI.yml
vastin: Agree. I need to verify what happen when reaching SSM quota. We could add the clean up logic in  resource-clean.yml later.

tools/ssm/linux-amd64-deb/install.sh
vastin: Agree. I would prefer to `#!/usr/bin/env bash` and verify in all supported AMIs.

tools/ssm/ssm_create.sh
```diff
@@ -0,0 +1,47 @@
...
+
+set -eu
+
+PKG_NAME=$1
+REL_VERSION=$2
+S3_BUCKET=$3
+
+output=$(aws ssm list-documents --filter Key=Owner,Values=Self Key=DocumentType,Values=Package Key=Name,Values=${PKG_NAME})
+echo ${output}
+status=$(echo ${output} | python3 -c "import sys, json; print('empty') if len(json.load(sys.stdin)['DocumentIdentifiers']) == 0 else print('exist')")
```
mxiamxia, line 23: Q: do we need to check or pre-install if `python3 -c ` is not available on the target host?

tools/ssm/ssm_create.sh
```diff
@@ -0,0 +1,47 @@
...
+if [[ ${status} == "empty" ]]
```
mxiamxia, line 24: Excuse me if has asked this before. For `aws ssm create-document` in CI workflow, does it just create/update the document in our Test account for integration tests. And the Prod version of document will be created in CD workflow?

tools/ssm/ssm_create.sh
vastin: The python3 is listed in installed software of ubuntu-latest test runner as below.
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#installed-software

tools/ssm/ssm_create.sh
vastin: Exactly. The pro version of document will be created in CD workflow with same artifacts which passes CI workflow as release candidate.

mxiamxia (approved):
LGTM. Pls resolve the remaining comments before merging.

.github/workflows/CI.yml
vastin: The SSM quota issue will be addressed in a separate PR.

tools/ssm/linux-amd64-deb/install.sh
vastin: Fixed

PR #447, opened 2021-04-16
bug fix for container insight

pxaws:
**Description:** 
* Fix inconsistent metric units
* Use gopsutil to get cpu/mem capacity instead
   * cadvisor is used previously but it only work for linux
   * gopsutil works for linux, windows and mac os
* Detect cluster name and autoscaling group in receiver
  * we previously rely on the upstream `resourcedetectionprocessor` to fetch ec2 tags to find out cluster name, but that method does not always work. 
  * we now use the same method as cloudwatch agent as it is guaranteed to work
* Add delay at the start of metric collection to avoid missing metric data point intermittently 
* Resolve some regex warning

### DIFF
--- a/deployment-template/eks/eks-ci-daemonset.yaml
+++ b/deployment-template/eks/eks-ci-daemonset.yaml
@@ -1,0 +1,252 @@
+# create namespace
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: aws-otel-eks
+  labels:
+    name: aws-otel-eks
+
+---
+# create aoc-agent service account and role binding
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: aws-otel-sa
+  namespace: aws-otel-eks
+
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: aoc-agent-role
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "nodes", "endpoints"]
+    verbs: ["list", "watch"]
+  - apiGroups: ["apps"]
+    resources: ["replicasets"]
+    verbs: ["list", "watch"]
+  - apiGroups: ["batch"]
+    resources: ["jobs"]
+    verbs: ["list", "watch"]
+  - apiGroups: [""]
+    resources: ["nodes/proxy"]
+    verbs: ["get"]
+  - apiGroups: [""]
+    resources: ["nodes/stats", "configmaps", "events"]
+    verbs: ["create", "get"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["aoc-clusterleader"]
+    verbs: ["get","update"]
+
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: aoc-agent-role-binding
+subjects:
+  - kind: ServiceAccount
+    name: aws-otel-sa
+    namespace: aws-otel-eks
+roleRef:
+  kind: ClusterRole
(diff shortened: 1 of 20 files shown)


internal/aws/containerinsightcommon/const.go
```diff
@@ -5,10 +5,11 @@ const (
 	InstanceType       = "InstanceType"
 	GoPSUtilProcDirEnv = "HOST_PROC"
 
-	MinTimeDiff    = 50 * 1000 // We assume 50 micro-seconds is the minimal gap between two collected data sample to be valid to calculate delta
-	ClusterNameKey = "ClusterName"
-	NodeNameKey    = "NodeName"
-	Version        = "Version"
+	MinTimeDiff             = 50 * 1000 // We assume 50 micro-seconds is the minimal gap between two collected data sample to be valid to calculate delta
```
pingleig, line 8: Can we use unit from time package like `50 * time.Second`

internal/aws/containerinsightcommon/const.go
```diff
@@ -107,11 +116,15 @@ var metricToUnitMap = make(map[string]string)
...
-	metricToUnitMap[CpuSystem] = UnitNanoSecond
-	metricToUnitMap[CpuLimit] = UnitCount
+	//The following metrics are reported in unit of millicores, but cloudwatch doesn't support it
+	//  CpuTotal
+	//	CpuUser
+	//	CpuSystem
+	//	CpuLimit
+	//  CpuRequest
 	metricToUnitMap[CpuUtilization] = UnitPercent
+	metricToUnitMap[CpuReservedCapacity] = UnitPercent
```
pingleig, line 126: You can merge all the map assignment into one.

```
var metricsToUniMap ....

func init() {
   t := map[xxx]xxx{
        a: b,
        c: d,
   }
   metricsToUnitMap = t
}
```

internal/receiver/awscontainerinsightreceiver/host/nodeCapacity.go
```diff
@@ -0,0 +1,51 @@
...
+
+type NodeCapacity struct {
+	MemCapacity int64
+	CPUCapacity int64
+	logger      *zap.Logger
+}
+
+func NewNodeCapacity(logger *zap.Logger) *NodeCapacity {
+	if _, err := os.Lstat("/rootfs/proc"); os.IsNotExist(err) {
+		panic("/rootfs/proc doesn't exists")
```
pingleig, line 23: Can we return error instead of panic here

pingleig:
btw: can fix those two regex warnings in https://github.com/aws-observability/aws-otel-collector/security/code-scanning

internal/receiver/awscontainerinsightreceiver/host/nodeCapacity.go
pxaws: Sure.

internal/aws/containerinsightcommon/const.go
pxaws: Good suggestion.

internal/aws/containerinsightcommon/const.go
pxaws: Sure.

deployment-template/eks/eks-ci-daemonset.yaml
```diff
@@ -0,0 +1,252 @@
+# create namespace
```
mxiamxia, line 1: Lets align the CI deployment template files location with PR https://github.com/aws-observability/aws-otel-collector/pull/453/files

mxiamxia (approved):
LGTM!

PR #449, opened 2021-04-16
Add SSM package release support in CD workflow

vastin:
**Description:**
Create and publish AWSDistroOTel-Collector SSM package in multiple commercial regions.


.github/workflows/ssm-release.yml
```diff
@@ -0,0 +1,249 @@
...
+  SSM_S3_BICKET_CN: aws-otel-collector-cn
+  RC_S3_BUCKET: aws-otel-collector-release-candidate
+
+jobs:
+  release-ssm-package:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the sha
+        uses: actions/checkout@v2
+        # with:
```
pingleig, line 46: Are we commenting out the sha on purpose?

.github/workflows/ssm-release.yml
```diff
@@ -0,0 +1,249 @@
...
+              ${{ env.SSM_S3_BICKET_BAH }}/ssmfile/${{ github.event.inputs.version }} ${region}
+          done
+
+      - name: Publish SSM packages to MXP
+        if: github.event.inputs.public == 'true'
+        run: |
+          REGIONS="eu-south-1"
+          for region in ${REGIONS}; do
+            echo "Share package to public in ${region}"
+            aws --region ${region} ssm modify-document-permission \
```
pingleig, line 213: I think we can merge the shared logic into common shell/python scripts, e.g. HKG, MXP etc. have very similar steps

.github/workflows/ssm-release.yml
```diff
@@ -0,0 +1,249 @@
...
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+name: release SSM package
+
+on:
+  workflow_dispatch:
```
pingleig, line 17: It is possible to make region part of input, so only some region get updated? I am not sure if we will have use cases like that though.

tools/ssm/ssm_create.sh
```diff
@@ -17,31 +17,37 @@ set -eu
...
         --version-name ${REL_VERSION} \
         --document-type Package
 else
-    output=$(aws ssm update-document \
-        --name "${PKG_NAME}" \
-        --content file://build/packages/ssm/manifest.json \
-        --attachments Key="SourceUrl",Values="https://s3.amazonaws.com/${S3_BUCKET}" \
-        --version-name ${REL_VERSION} \
-        --document-version "\$LATEST")
+    len=$(aws --region ${REGION} ssm list-document-versions --name "${PKG_NAME}" | python3 -c "import sys, json; print(len([ver for ver in json.load(sys.stdin)['DocumentVersions'] if ver['VersionName'] == '${REL_VERSION}']))")
```
pingleig, line 34: Since we already have the python script, can we use python for ssm_create as well, it is shelling out to python3 for json anyway (btw: you can also use `jq` but I think using python for all of them reduce maintain workload.  You can install boto like [this link](https://github.community/t/help-with-aws-cli/17054/2)

.github/workflows/ssm-release.yml
vastin: It is commented for testing and I will fix it.

.github/workflows/ssm-release.yml
vastin: Each step has few commands and steps in Github workflow could provide a clear view of progress. If merge all steps into one script, it may lose some observability.

.github/workflows/ssm-release.yml
vastin: Yes. It is a valid use case when deployment fails in some region. The ssm_create.sh will skip regions which have already created SSM package with same version. It is safe to re-run the workflow to continue the failed deployment. So the region name is not required as input.

tools/ssm/ssm_create.sh
vastin: Agree. The logic of ssm_create.sh is simple and it is trivial to rewrite in python. However I'd like to use pre-installed AWS CLI instead of installing boto since AWS CLI is enough to handle the situation.

vastin: Rewrite script in python with boto3 as you suggested.

pingleig (approved):
LGTM

.github/workflows/ssm-release.yml
pingleig: ok, I was thinking if we can use it to test release on specific region, but in that case we can just run the python script locally (against test accounts)

mxiamxia (approved):
Thx

Issue #340, opened 2021-02-04
[bug] Config error should return exit code 1


pingleig:
## Behaviour

If you give an invalid config (e.g. malformed json), the container exit with 0 and  without any error in standard output due to  #339 .
This is a very frustrating experience util you switch to otelcol or otelcontribcol .... 

## Reproduce

```yaml
version: "2"
services:
  aoc:
    image: public.ecr.aws/aws-observability/aws-otel-collector:latest
    command: [ "--config=/etc/otel-agent-config.yaml", "--log-level=DEBUG" ]
    environment:
      - AWS_REGION=us-west-2
    volumes:
      - ./otel-invalid.yaml:/etc/otel-agent-config.yaml
```

```yaml
receivers:
  prometheus:
    config:
      scrape_configs:
        - job_name: "static"
            # NOTE: the indent here is WRONG
            static_configs:
              - targets: [ 'localhost:8888' ] # self metric
                labels:
                  foo: bar
        - job_name: "task_def_1"
          file_sd_configs:
            - files:
                - '/etc/ecs_sd_targets.yaml'
exporters:
  awsemf:

service:
  pipelines:
    metrics:
    receivers: [ prometheus ]
    exporters: [ awsemf ]
```

```text
docker-compose up
Recreating config-error_aoc_1 ... done
Attaching to config-error_aoc_1
aoc_1  | AWS OTel Collector version: v0.7.0
aoc_1  | 2021/02/04 04:37:34 find no extra config, skip it, err: open /opt/aws/aws-otel-collector/etc/extracfg.txt: no such file or directory
aoc_1  | 2021-02-04T04:37:34.983Z	INFO	service/service.go:411	Starting AWS OTel Collector...	{"Version": "v0.7.0", "GitHash": "14a0d2cf11fe4ed2cd254241acdda17c54a0e59e", "NumCPU": 12}
aoc_1  | 2021-02-04T04:37:34.984Z	INFO	service/service.go:255	Setting up own telemetry...
aoc_1  | 2021-02-04T04:37:34.984Z	INFO	service/telemetry.go:102	Serving Prometheus metrics	{"address": "localhost:8888", "level": 0, "service.instance.id": "f1c1104e-fe11-419e-b1c0-c502299c2694"}
aoc_1  | 2021-02-04T04:37:34.985Z	INFO	service/service.go:292	Loading configuration...
```

## Expected

- exit with 1
- print error out

```text
[...]

PR #450, opened 2021-04-17
fix the collector fail to start issue after updated the new config deps

mxiamxia:
**Description:**
Fix the new config interface error and taken care of windows exit code(0) issue

### DIFF
--- a/cmd/awscollector/main_windows.go
+++ b/cmd/awscollector/main_windows.go
@@ -18,6 +18,9 @@
 package main
 
 import (
+	"log"
+	"os"
+
 	"github.com/pkg/errors"
 	"golang.org/x/sys/windows/svc"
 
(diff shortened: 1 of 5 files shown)


pingleig (approved):
Can you remove the config.TryConfig as well (and its usage in config factory test).

cmd/awscollector/main_windows.go
```diff
@@ -45,3 +47,11 @@ func runService(params service.Parameters) error {
 
 	return nil
 }
+
+// Has to set exit code 0 for the fatal errors when run the collector as service in Windows
```
pingleig, line 51: This is a new level of hack 😅 for #340

vastin (approved):
LGTM

cmd/awscollector/main_windows.go
mxiamxia: any better idea? I think it seems better than previous one :)

mxiamxia:
> Can you remove the config.TryConfig as well (and its usage in config factory test).

done, thx

cmd/awscollector/main_windows.go
pingleig: yeah, it's better than previous one (which did even work ...

PR #456, opened 2021-04-20
Config cleansing after job and instance label is added in prometheus receiver

bjrara:
**Why do we need it?**
We don't need to add `job` and `instance` label explicitly since https://github.com/open-telemetry/opentelemetry-collector/pull/2897 is merged.


mxiamxia (approved):
Thanks!

PR #457, opened 2021-04-20
remove unused IMDS v1 test func from packaging script

mxiamxia:
**Description:** 
removed the outdated info


pingleig (approved):
LGTM

Though I think the PR title should be remove unused IMDS v1 call

PR #461, opened 2021-04-22
[In Process]Container Insights receiver for ECS EC2 instance level metrics

JohnWu20:
**Description:** 

This pr will add new feature to awscontainerinsightsreceiver to make it collect ECS EC2 instance level metrics.

There are some changes made for the original awscontainerinsightsreceiver. One of the main change is:
Add region for the config of Ec2 tag and EBS volume session config

There are some new feature added to this receiver:
1,  Call [Amazon ECS container agent introspection](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-introspection.html) end point to get task and cluster information.
2,  Get cpu and mem capacity based on the ECS info
3,  Add resource data to the receiver.

Test:
End2end test.

Unit test is still in process.


pxaws (commented):
Please remove this file `internal/receiver/awscontainerinsightreceiver/ecs/.DS_Store`

internal/receiver/awscontainerinsightreceiver/host/ebsvolume.go
```diff
@@ -28,7 +28,7 @@ type EbsVolume struct {
 	shutdownC  chan bool
 }
 
-func NewEbsVolume(instanceId string, refreshInterval time.Duration, logger *zap.Logger) *EbsVolume {
+func NewEbsVolume(instanceId string, refreshInterval time.Duration, logger *zap.Logger, region string) *EbsVolume {
```
pxaws, line 31: Could you put `region` after the `instanceId`?

internal/receiver/awscontainerinsightreceiver/host/ebsvolume.go
```diff
@@ -43,7 +43,7 @@ func NewEbsVolume(instanceId string, refreshInterval time.Duration, logger *zap.
 
 	//add some sleep jitter to prevent a large number of receivers calling the ec2 api at the same time
 	time.Sleep(hostJitter(3 * time.Second))
-	ebsVolume.refresh()
+	ebsVolume.refresh(region)
```
pxaws, line 46: Could you make `region` as a field for ebsVolume so that you don't need to pass the parameter in `refresh()` function?

internal/receiver/awscontainerinsightreceiver/host/ebsvolume.go
```diff
@@ -66,13 +66,13 @@ func (e *EbsVolume) Shutdown() {
...
 
-func (e *EbsVolume) refresh() {
+func (e *EbsVolume) refresh(region string) {
 	if e.instanceId == "" {
 		return
 	}
 
 	e.logger.Info("Fetch ebs volumes from ec2 api")
-	sess, err := session.NewSession(&aws.Config{})
+	sess, err := session.NewSession(&aws.Config{Region: aws.String(region)})
```
pxaws, line 75: Create the config conditionally?

* if region is empty, use the default one
* otherwise, use `&aws.Config{Region: aws.String(region)}`

internal/receiver/awscontainerinsightreceiver/ecs/ecs_util.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"fmt"
+	common "github.com/aws-observability/aws-otel-collector/internal/aws/containerinsightcommon"
+	"go.opentelemetry.io/collector/consumer/pdata"
+	"go.uber.org/zap"
+	"reflect"
+)
+
+func insertMetrics(key string, value interface{}, ilms pdata.InstrumentationLibraryMetricsSlice, metricType string, timestamp pdata.Timestamp, logger *zap.Logger) {
```
pxaws, line 11: Looks like there is a lot of duplicated code with `aws/containerinsightcommon/utils.go`. Can you create the ecs metrics not in OpenTelemetry format? You can just use map[string]interface{} to represent metrics values and then use map[string]string to represent the labels. In my opinion, I think it is easy to do metric decoration with those simple structures.

internal/receiver/awscontainerinsightreceiver/ecs/ecs_util.go
```diff
@@ -0,0 +1,127 @@
...
+		ilms.Append(common.DoubleGauge(key, unit, float64(t), timestamp))
+	case float64:
+		ilms.Append(common.DoubleGauge(key, unit, t, timestamp))
+	default:
+		valueType := fmt.Sprintf("%T", value)
+		logger.Warn("Detected unexpected field", zap.String("key", key), zap.Any("value", value), zap.String("value type", valueType))
+	}
+}
+
+func getMetricValue(metricName string, rms pdata.ResourceMetrics) (bool, interface{}, pdata.Timestamp) {
```
pxaws, line 39: There is a lot of duplicate code here. See https://github.com/aws-observability/aws-otel-collector/blob/main/internal/processor/awscontainerinsightprocessor/stores/utils.go#L189-L210

As you can see, it is very cumbersome to get a metric value from openTelemetry format. I would suggest to not use it for raw metrics. You can use simple maps to represent the metrics and do processing and then convert them to openTelemetry format at the final stage.

pxaws (commented):
I think the folder name `ecs` is quite confusing as it indicates all ecs metrics are from this folder. You could use `ecsinstance` instead?

alolita:
Old PR. Closing.

PR #471, opened 2021-04-23
lock the tf version to fix aws resource destroy error

mxiamxia:
lock the tf version to fix aws resource destroy error caused by TF bug
https://github.com/hashicorp/terraform-provider-aws/issues/8040#issuecomment-822774525

locally


pingleig (approved):
LGTM

PR #479, opened 2021-04-26
remove container insight components

pxaws:
**Description:** 
Remove container insight components. We are going to upstream the code.


mxiamxia (approved):
Thanks!

Issue #451, opened 2021-04-19
aws-otel-collector when run in container gives a permission denied error to create directory


AnuPrabath:
We are able to run aws-otel-collector in Cloud foundry but binary still tries to create a directory /opt/aws – Will there be any update to suppress this error when RUN_IN_CONTAINER env variable is set ?
ERR 2021-03-24 20:34:51.659421625 +0000 UTC m=+0.053561495 write error: can't make directories for new logfile: mkdir /opt/aws: permission denied

mxiamxia:
Can you pls tell which version of AWS OTel Collector you were running? After v0.8.0 it shouldn't ask you to create directly for logs.

https://github.com/aws-observability/aws-otel-collector/blob/main/pkg/logger/logger.go#L64

AnuPrabath:
@mxiamxia  - We are using v0.8.0 Collector with RUN_IN_CONTAINER flag set to True. I still see directory creation errors both in Cloud foundry and in EKS, we do run applications as non root users and I believe awscollector runs with awsotel user in v0.8.0 image not sure whether this might be the issue.

pingleig:
@AnuPrabath we found the problem,  my bad in https://github.com/aws-observability/aws-otel-collector/pull/341 there are two redirect to file logic in our distro, one is standard library's `log` package, another is a hook for `zap` (used by all otel components). I only disabled the redirect for first one, so the collector still tries to create the folder and failed when running as non root user.

https://github.com/aws-observability/aws-otel-collector/blob/5e1bcb2a6979144c87a6afe8abc7b8ed00552b2a/pkg/logger/logger.go#L36-L44

PR #341, opened 2021-02-05
log: Write to stderr in container Fix #339

pingleig:
**Description:**

Fix #339 

**Link to tracking Issue:** <Issue number if applicable>

- #339 

**Testing:**

Using example in #340 no the config error shows up when running via docker-compose (NOTE: the exit 0 is still 0 until #338 is merged)

```text
aoc_1  | AWS OTel Collector version: v0.7.0
aoc_1  | 2021/02/05 00:55:04 find no extra config, skip it, err: open /opt/aws/aws-otel-collector/etc/extracfg.txt: no such file or directory
aoc_1  | 2021-02-05T00:55:04.100Z	INFO	service/service.go:411	Starting AWS OTel Collector...	{"Version": "v0.7.0", "GitHash": "94015373a37231d8678d115fe05e90d662c555a7", "NumCPU": 6}
aoc_1  | 2021-02-05T00:55:04.101Z	INFO	service/service.go:255	Setting up own telemetry...
aoc_1  | 2021-02-05T00:55:04.105Z	INFO	service/telemetry.go:102	Serving Prometheus metrics	{"address": "localhost:8888", "level": 0, "service.instance.id": "bbb29b0d-6f5b-4bb0-b129-67030d3149dd"}
aoc_1  | 2021-02-05T00:55:04.105Z	INFO	service/service.go:292	Loading configuration...
aoc_1  | Config file is missing or invalid, error loading config file "/etc/otel-agent-config.yaml": While parsing config: yaml: line 6: did not find expected key
config-error_aoc_1 exited with code 0
```

**Documentation:** 

Log to stdout/stderr is the expected the behaviour the collector container.

PR #484, opened 2021-04-28
Fix log init fail when running as nonroot

bjrara:
**Why do we need it?**
Fixes https://github.com/aws-observability/aws-otel-collector/issues/451

This is a follow-up PR of https://github.com/aws-observability/aws-otel-collector/pull/341. We skip log hook initialization when running inside containers.


mxiamxia (approved):
Thanks!

PR #486, opened 2021-04-29
fix DockerHub release account username in CD WF

mxiamxia:
**Description:** 
fix DockerHub release account username in CD WF


vastin (approved):
LGTM

PR #487, opened 2021-04-29
terraform_version: 0.14.10

mxiamxia:
use fixed terraform version 0.14.10 before fixing the compatible issue in test framework. The `list` func deprecated in tf v0.15.0

```
Call to function "list" failed: the "list" function was deprecated in
│ Terraform v0.12 and is no longer available; use tolist([ ... ]) syntax to
│ write a literal list.
```


vastin (approved):
LGTM

PR #489, opened 2021-04-29
Fix the link issue for AOC task defination

JohnWu20:
**Description:** 
[AWS Distro for OpenTelemetry for the ECS customers](https://aws-otel.github.io/docs/setup/ecs/task-definition-for-ecs-ec2) use the following link to download the task definition:
https://github.com/aws-observability/aws-otel-collector/blob/main/examples/ecs/ecs-ec2-sidecar.json
https://github.com/aws-observability/aws-otel-collector/blob/main/examples/ecs/ecs-fargate-sidecar.json
They can't get the right template after the[ layout updating](https://github.com/aws-observability/aws-otel-collector/commit/f854aa0cd7aa6aab1f74e0c4853a6bef12a07aba).

So this PR will change the layout for these two task definations.


mxiamxia:
Hi John, we changed the example template folder structure on purpose. We want to separate out default config (cloudwatch), aws managed prometheus and ECS metadata examples to diff folders. Could you help to make a change to our user guide docs instead of change it back? Pls let me know if it makes sense to you.

pingleig:
In the long term we should use version prefix for both the doc website and the link to manifest in aoc repo. As long as the tag is immutable the link is valid.

For the doc website, [gatsby can use versioning with source-git plugin](https://www.gatsbyjs.com/blog/2019-07-03-using-themes-for-distributed-docs/#:~:text=versioning%20with%20gatsby-source-git). For adot repo, github tag would work out of box.

hossain-rayhan:
The new folder structure makes sense to me. From a full product perspective, I think the new structure would be good. We may also support AppMesh and GuardDuty in future. If we need more config to support them, this structure would be more acceptable. 

If we don't have any breaking concern, we can update our doc to reflect the new paths. Also, I like the idea about prefixing with version.

JohnWu20:
Close this pr, and I will change the link in the instruction. Thanks everyone.

PR #490, opened 2021-04-29
Disable fail-fast in CD workflow

vastin:
**Description:**. 
Disable fail-fast in CD workflow to make it run more tests when failed. It will make re-run faster.


mxiamxia:
ah. that's why some tests were cancelled because another test in the same job matrix got failed.

mxiamxia (approved):
Thanks

PR #501, opened 2021-05-06
Update CI job after terraform test case update

bjrara:
**Why do we need it?**
This is a follow-up PR of https://github.com/aws-observability/aws-otel-test-framework/pull/281. 

This PR cleanse CI jobs after container insight test cases are merged into EKS.


pingleig (approved):
LGTM

Issue #419, opened 2021-03-30
failed to shutdown exporter


sammy22:
I ran the Otel-collector image locally. 
docker run --rm -p 4317:4317 -p 55680:55680 -p 8889:8888       -e AWS_REGION=us-west-2   -v ${PWD}/config.yaml:/otel-local-config.yaml -v ~/.aws:/root/.aws  --name awscollector public.ecr.aws/aws-observability/aws-otel-collector:v0.8.0   --config otel-local-config.yaml;

when I just stop the container 
I get the error 
2021-03-30T07:11:05.119Z	INFO	service/service.go:469	Shutdown complete.
Error: failed to shutdown pipelines: failed to shutdown exporters: sync /dev/stderr: invalid argument

I get the same error when I run this in ecs task as well.

pingleig:
I think the error might come from the logger when flushing https://github.com/uber-go/zap/issues/328

sammy22:
This error occurs only when i use x-ray exporter. the error is because of logger.Sync(). according to the discussion here uber-go/zap#328. sync errors can be safely ignored. I did the same and recreated the image . Can this be fixed in x-ray exporter ? 
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/21914fa0deb8fcb87b6eb5f72825feb54d1fce49/exporter/awsxrayexporter/awsxray.go#L93

pingleig:
@sammy22 sure, we can fix it in the contrib repo cc @mxiamxia @awssandra

PR #503, opened 2021-05-07
Add new error reporting feature in AOC for ECS

skyduo:
**Description:** 
This is the new error reporting feature for ECS. Except sending all error logs to the original destinations, we need to also send error or higher level log messages to another single error log file (the file path configurable later) and keep it less than 1 KB, so that ECS agent could mount this path and read the error log file. When the error log file exceed the max size (1KB), we want to delete the oldest log message to release the space. Also we want to rotate the logs by repeatedly delete the expired logs (the expired time is also configurable later). 

**Testing:** 
`make build` and run the executable file with a config file with some errors, and succeed with 2 config errors write into the error log file.

**Documentation:**


pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+	"time"
+)
+
+
+var UnixErrorLogPath = "/opt/aws/aws-otel-collector/logs/aws-otel-collector-error.log"
+
+var WindowsErrorLogPath = "C:\\ProgramData\\Amazon\\AWSOTelCollector\\Logs\\aws-otel-collector-error.log"
```
gavindoudou, line 17: May we use const here?

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+					}
+				} else {
+					queue.Remove(queue.Front())
+				}
+			}
+			err := file.Close()
+			if err != nil {
+				RotateError <- err
+			}
+		case <- quit:
```
gavindoudou, line 139: If we only use quit once here, I wonder what's the purpose of it?

PettitWesley:
@hossain-rayhan Can you help with review on this too

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+
+var WindowsErrorLogPath = "C:\\ProgramData\\Amazon\\AWSOTelCollector\\Logs\\aws-otel-collector-error.log"
+
+var errorLogfile = getErrorLogFilePath()
+
+var errorLogMaxAge = 10
+
+var queue = list.New()
+var errorLogChannel = make(chan zapcore.Entry, 10)
+var ticker = time.NewTicker(10 * time.Minute)
```
PettitWesley, line 25: Use env var STATUS_MESSAGE_TTL to set the frequency of the ticker. It should either be the same as STATUS_MESSAGE_TTL or should be less

cmd/awscollector/main.go
```diff
@@ -54,6 +54,9 @@ func main() {
 	// init lumberFunc for zap logger
 	lumberHook := logger.GetLumberHook()
 
+	// init a gorountine to rotate the error logs
+	go logger.RotateErrorLog()
```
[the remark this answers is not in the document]
hossain-rayhan: jsut to make sure this `STATUS_MESSAGE_FILE_PATH` will always be there, right? Either by managed telemetry default option or by customer.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+
+	}
+	return byteNum, err
+}
+
+var RotateError = make(chan error)
+
+// RotateErrorLog func help select the different case between either write error log into file or
+// rotate and delete old logs in file based on TTL
+func RotateErrorLog() {
```
PettitWesley, line 108: I still don't think this is an ideal function name. This function might as well be called `ECSErrorReporter` or something. Its like the main function for this feature. It does more than just rotation.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/crossdock/crossdock-go v0.0.0-20160816171116-049aabb0122b
+	github.com/golangci/golangci-lint v1.38.0 // indirect
```
hossain-rayhan, line 7: ignore my knowledge here- why the comment // indirect here?

pkg/logger/ecs_error_log.go
hossain-rayhan, line 17: Out of curosity, did we test it on windows?

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+					if err != nil {
+						RotateError <- err
+					}
+				} else {
+					queue.Remove(queue.Front())
+				}
+			}
+			err := file.Close()
+			if err != nil {
+				RotateError <- err
```
PettitWesley, line 137: what's the point of writing to this error channel? Does anything read from it? If you write to a channel and nothing is reading from it, then the channel will block your programs execution.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+)
+
+
+var UnixErrorLogPath = "/opt/aws/aws-otel-collector/logs/aws-otel-collector-error.log"
+
+var WindowsErrorLogPath = "C:\\ProgramData\\Amazon\\AWSOTelCollector\\Logs\\aws-otel-collector-error.log"
+
+var errorLogfile = getErrorLogFilePath()
+
+var errorLogMaxAge = 10
```
hossain-rayhan, line 21: Maybe rename the variabl postfixing unit, like `errorLogMaxAgeInMinutes`?

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	err = errorFile.Truncate(0)
+	if err != nil {
+		return byteNum, err
+	}
+	// iterate errors in queue, if the old error's message same as the new error's message, skip it,
+	// otherwise write it into the file and queue
+	var backupQueue = list.New()
+	for queue.Len() > 0 {
+		e := queue.Front().Value.(zapcore.Entry)
```
PettitWesley, line 46: I don't think you need to create two queues and dequeue from one into the other. That seems very inefficient. 

This is not how the documentation for list suggests you iterate over it: https://golang.org/pkg/container/list/

Also, more importantly, in go we generally use slices instead of lists. The core list type is only really useful if you need a collection that can hold multiple types of data I think. See: https://stackoverflow.com/questions/21326109/why-are-lists-used-infrequently-in-go

I'd prefer to see this code re-written to use a slice.

pkg/logger/ecs_error_log.go
PettitWesley: Its okay if it doesn't work on Windows yet. Only linux is required.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+		queue.Remove(queue.Front())
+	}
+	backupQueue.PushBack(newErrorLog)
+	queue = backupQueue
+	n, err = errorFile.Write([]byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+		newErrorLog.Time, newErrorLog.Level, newErrorLog.Caller, newErrorLog.Message)))
+	if err != nil {
+		return byteNum, err
+	}
+	byteNum += n
```
hossain-rayhan, line 65: Wondering rather than doing a dequeue and enqueuing again, keeping a map and list wouldn't b e efficient. map for lookup and list for deleting the item in order?

pkg/logger/ecs_error_log.go
hossain-rayhan: Then maybe we should remove it for now. We will introduce them later when we will add them after testing.

pingleig (changes requested):

pkg/logger/ecs_error_log_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+var testErrorLogFilePath = "/opt/aws/aws-otel-collector/logs/aws-otel-collector-error-test.log"
+
+func TestWriteError(t *testing.T) {
+	defer os.Remove(testErrorLogFilePath)
+	openExistingOrNewErrorFile = OpenExistOrNew
+	entry := zapcore.Entry{
+		Message: "test",
+	}
+	_, error := WriteError(entry, testErrorLogFilePath)
```
JohnWu20, line 24: Why this test case only cover one output of `WirteError` function?

pkg/logger/ecs_error_log_test.go
```diff
@@ -0,0 +1,94 @@
...
+	queue.PushBack(e1)
+	queue.PushBack(e2)
+	byteNum := len(content1) + len(content2)
+	n, err := rotate(byteNum, testErrorLogFilePath)
+	require.NoError(t, err)
+	assert.Equal(t, len(content2), n)
+	queue.Remove(queue.Back())
+}
+
+func OpenExistOrNew() error {
```
JohnWu20, line 74: This function seems unnecessary to me. For unit test, it's better for you to use the original function which you used in your code instead of creating a new function here.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+			currentTime := time.Now()
+			file, _ := os.OpenFile(errorLogfile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0777)
+			_ = file.Truncate(0)
+			for queue.Len() > 0 {
+				e := queue.Front().Value.(zapcore.Entry)
+				if e.Time.Add(time.Duration(errorLogMaxAge) * time.Hour).After(currentTime) {
+					_, err := file.Write([]byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+						e.Time, e.Level, e.Caller, e.Message)))
+					if err != nil {
+						RotateError <- err
```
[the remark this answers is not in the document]
PettitWesley: Could this function use the zap logger to log errors itself? And then those errors would be printed to stdout and sent to the log file as normal.

pkg/logger/ecs_error_log.go
skyduo: yeah, will change that, didn't attend the meeting last week which decide the final interface.

pkg/logger/ecs_error_log.go
skyduo: good idea, will change that

pkg/logger/ecs_error_log.go
skyduo: sure

pkg/logger/ecs_error_log.go
skyduo: i forget to handle this channel, will resolve it

pkg/logger/ecs_error_log.go
skyduo: sure, i will take a look

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+func rotate(byteNum int, errorLogfile string) (n int, err error) {
+	file, err := os.OpenFile(errorLogfile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0777)
+	if err != nil {
+		return byteNum, err
+	}
+	err = file.Truncate(0)
+	if err != nil {
+		return byteNum, err
+	}
+	errorNum := queue.Len()
```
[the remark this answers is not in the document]
skyduo: i think it sometimes can auto inline, don't know why it's not this time, will check that

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
...
+				if e.Time.Add(time.Duration(errorLogMaxAge) * time.Hour).After(currentTime) {
+					_, err := file.Write([]byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+						e.Time, e.Level, e.Caller, e.Message)))
+					if err != nil {
+						RotateError <- err
+					}
+				} else {
+					queue.Remove(queue.Front())
+				}
+			}
```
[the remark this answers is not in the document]
skyduo: the error will contains a timestamp, and i think this goroutine will keep running with the AOC, so if there is some error shows up, it will get this error channel, which doesn't related to the rotate case

pkg/logger/ecs_error_log_test.go
skyduo: oh yeah, the test part still not all done yet, will add some more test case later

pkg/logger/ecs_error_log.go
skyduo: i think this is required? point me out if i'm wrong but based on investigation, we need to contain a exit case when using goroutine with ticker

pkg/logger/logger.go
```diff
@@ -50,6 +49,9 @@ func GetLumberHook() func(e zapcore.Entry) error {
 		_, err := lumberjackLogger.Write(
 			[]byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v, Stack:%+v}\r\n",
 				e.Time, e.Level, e.Caller, e.Message, e.Stack)))
+		if e.Level >= zapcore.ErrorLevel {
```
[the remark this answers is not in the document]
skyduo: 1. for `RotateError`, it's my mistake that didn't handle it, will resolve it
2. okay i see your change, will modify my part
3. will add a logic to outside put error in channel
4. sure, will add more comments there

pkg/logger/ecs_error_log.go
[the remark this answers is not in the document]
skyduo: that's also i'm concerning about, like what happened if our error reporting logic cause some errors

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,201 @@
+package logger
+
+import (
+	"container/list"
```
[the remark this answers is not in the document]
pingleig: You don't need to, I cloned your branch and run `go mod tidy` that line should be gone. I think you should rebase the main branch before addressing review comments.

> This branch is 1 commit ahead, 68 commits behind aws-observability:main.

pkg/logger/ecs_error_log.go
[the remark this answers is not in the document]
pingleig: I suggest just ignore it, there was similar issue like https://github.com/aws-observability/aws-otel-collector/issues/419

docs/releases/v0.9.0.md
```diff
@@ -1,18 +1,24 @@
 # Changelog
 
-## [v0.9.0](https://github.com/aws-observability/aws-otel-collector/tree/v0.9.0) (2021-04-27)
+## [v0.9.0](https://github.com/aws-observability/aws-otel-collector/tree/v0.9.0) (2021-04-28)
```
gavindoudou, line 3: Can you rebase the code? This PR now includes Min's change. It's hard to review.

skyduo: yeah i submit a new pr and merged the correct version from mainline, problem resolved

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+	"time"
+)
+
+var ErrorFilePath = os.Getenv("STATUS_MESSAGE_FILE_PATH")
+
+var errorLogMaxAge = os.Getenv("STATUS_MESSAGE_TTL")
+var errorLogMaxAgeInHours, _ = strconv.Atoi(errorLogMaxAge)
+
+var size = os.Getenv("STATUS_MESSAGE_MAX_BYTE_LENGTH")
+var errorFileSize, _ = strconv.Atoi(size)
```
PettitWesley, line 18: we need to be able to handle errors in this code, including failure to convert the strings to ints.

I think this code may be should be added inside of a function.

What if you had a structure to store all of these:

```
type ECSErrorLogger struct {
ErrorFilePath string
errorLogMaxAge string
// put all of the error config options here
}

// then there is a create method which gets the env vars and sets everything:
func createECSErrorLogger() ECSErrorLogger {
os.getenv...
// set up the struct and return it
}
```

And then your other functions are methods in the structure:

```
// the methods can access the member fields on ECSErrorLogger structure
func (logger ECSErrorLogger) ECSErrorReporter() {}
func (logger ECSErrorLogger) WriteError() {}
```

PettitWesley: @pingleig @gavindoudou What do you think of this approach?

PettitWesley: If you go with this approach, move all of the global variables to that structure if possible. I don't like global variables. I think wrapping the behavior in a structure is better organization and cleaner.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+
+var queue []zapcore.Entry
+var errorLogChannel = make(chan zapcore.Entry, 10)
+var ticker = time.NewTicker(10 * time.Minute)
+var errorFile *os.File
```
PettitWesley, line 23: Why is this a global variable? I think this is again an argument for my idea of putting everything in a new structure. 

Also, it looks like you're keeping this file open between calls. I am not sure if that is ideal. @mxiamxia @pingleig Is that bad? Keeping the file open between write calls? See the code in `func WriteError`; it keeps the file pointer open between calls.

pkg/logger/ecs_error_log.go
JohnWu20: Agree, that will increase the readability and make the code more organized.

pkg/logger/ecs_error_log.go
skyduo: yeah, if we want to build all the things in an ECSErrorLogger type, this will be more clear, those variables will belongs to the type, and so does that error log file, just like everything set in the lumberjacklogger

pkg/logger/ecs_error_log.go
skyduo: i'm ok with this change

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+					queue = queue[1:]
+				}
+			}
+			_ = file.Close()
+		}
+	}
+}
+
+// WriteError func could write the new error log message into file
+func WriteError(newErrorLog zapcore.Entry, errorLogfile string) (n int, err error) {
```
JohnWu20, line 56: It seems that the return value of the function is not used anywhere. Maybe delete it?

PettitWesley, line 56: why the `n` return argument? What's it used for? I think its not used for anything.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+
+// ECSErrorReporter func help select the different case between either write error log into file or
+// rotate and delete old logs in file based on TTL
+func ECSErrorReporter() {
+	defer ticker.Stop()
+	for {
+		select {
+		// when new error add in the channel, this case will write it into file
+		case e := <-errorLogChannel:
+			_, _ = WriteError(e, ErrorFilePath)
```
PettitWesley, line 33: I don't love that we are ignoring errors from `WriteError`. I know  may be we can't write an error with the logger since this already processes logging errors and that would lead to recursive calls...

But would it be safe to print the error to stdout? That way it isn't completely ignored. If we ever have any issues with the error logger it would be impossible to debug...

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+		// this case will run every x mins (x is configurable) to delete the error which already expired, error log max age
+		// depends on STATUS_MESSAGE_TTL
+		case <-ticker.C:
```
PettitWesley, line 36: nit: for organization, and also may be to make unit testing easier, I'd prefer if you moved the code for this case to a separate function. 

Then you code is nice and simple and readable like:

```
select {		
		case e := &lt;-errorLogChannel:			
                      _, _ = WriteError(e, ErrorFilePath)
                  case <-ticker.C:
                             processTimeout()
```

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+	}
+	err = errorFile.Truncate(0)
+	if err != nil {
+		return byteNum, err
+	}
+	// iterate errors in queue, if the old error's message same as the new error's message, skip it,
+	// otherwise write it into the file and queue
+	var errorNum = len(queue)
+	for i := 0; i < errorNum; i++ {
```
PettitWesley, line 70: nit: I don't think you need `errorNum`, I think you can just use `len(queue)` in the for loop

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+		if e.Message != newErrorLog.Message {
+			n, err = errorFile.Write([]byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+				e.Time, e.Level, e.Caller, e.Message)))
+			if err != nil {
+				return byteNum, err
+			}
+			byteNum += n
+			queue = append(queue, e)
+		}
+		queue = queue[1:]
```
PettitWesley, line 81: why are you mutating the queue in this loop? I think I commented on this before. I also don't know why you are using a queue. Can you just use a normal golang slice?

PettitWesley: Sorry I just realized the queue is a slice. In that case, please iterate over it using a `range` for loop since that is standard in golang: https://tour.golang.org/moretypes/16

PettitWesley: Please do not mutate the slice in the for loop.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	queue = append(queue, newErrorLog)
```
PettitWesley, line 83: Based on the design, you are supposed to only add the new error log to the queue if it is not already in the queue. That way there are no duplicates in the queue or the file.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+		if err != nil {
+			return byteNum, err
+		}
+	}
+	return byteNum, err
+}
+
+// rotate func could delete the old logs to make sure error log file less than 1 KB
+func rotate(byteNum int, errorLogfile string) (n int, err error) {
+	file, err := os.OpenFile(errorLogfile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0777)
```
PettitWesley, line 102: why re-open the file here? In the writeError method you only re-open it if needed.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+func errorDir() string {
+	return filepath.Dir(ErrorFilePath)
+}
+
+func errorFileName() string {
+	if ErrorFilePath != "" {
+		return ErrorFilePath
+	}
+	name := filepath.Base(os.Args[0]) + "-lumberjack.log"
+	return filepath.Join(os.TempDir(), name)
```
PettitWesley, line 174: This looks like old code which should be removed? The log file name/path comes from the env var.

pkg/logger/ecs_error_log_test.go
PettitWesley: I agree with John's comment. 

Can't you use the env vars in the test? You can set the test file path as the value of the env var and then the code will use it. 

This will work if you do my suggested refactoring where you move all global variables into a structure.

pkg/logger/ecs_error_log.go
gavindoudou: Nice explanation! Very clear structure. Agree

pkg/logger/ecs_error_log.go
gavindoudou: Maybe we can write something like log.Printf

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+		case <-ticker.C:
+			currentTime := time.Now()
+			file, _ := os.OpenFile(ErrorFilePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0777)
+			_ = file.Truncate(0)
+			errorNum := len(queue)
+			for i := 0; i < errorNum; i++ {
+				e := queue[0]
+				if e.Time.Add(time.Duration(errorLogMaxAgeInHours) * time.Hour).After(currentTime) {
+					_, _ = file.Write([]byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+						e.Time, e.Level, e.Caller, e.Message)))
```
gavindoudou, line 45: A little bit confused here. If `i=0`, the time of `e` is after the current time, we will go to this if section and write the log. Then, when `i=1`, we still check `e` is after the current time? It looks like that e does not change and we write `e` to the log agin?  Can you explain a little bit here? Thanks

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+var errorFile *os.File
+
+// ECSErrorReporter func help select the different case between either write error log into file or
+// rotate and delete old logs in file based on TTL
+func ECSErrorReporter() {
+	defer ticker.Stop()
+	for {
+		select {
+		// when new error add in the channel, this case will write it into file
+		case e := <-errorLogChannel:
```
gavindoudou, line 32: There is another `e` in the ticker's case, it would be nice if we could use another name.

pkg/logger/ecs_error_log.go
skyduo: yeah, there is an issue here, will fix it

pkg/logger/ecs_error_log.go
skyduo: sure, didn't notice this part, will change

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,175 @@
...
+			}
+			_ = file.Close()
+		}
+	}
+}
+
+// WriteError func could write the new error log message into file
+func WriteError(newErrorLog zapcore.Entry, errorLogfile string) (n int, err error) {
+	var byteNum int
+	if errorFile == nil {
```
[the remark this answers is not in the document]
skyduo: no, it keep the file open, and it just check the `file == nil` like what i did here

pkg/logger/ecs_error_log.go
skyduo: i think it's just a check in case it closed by some random issue

pkg/logger/ecs_error_log.go
skyduo: yeah, will do that

pkg/logger/ecs_error_log.go
skyduo: good idea, will change it

pkg/logger/ecs_error_log.go
skyduo: yeah, will use `log.Printf`

pkg/logger/ecs_error_log.go
skyduo: i just want to track the way `file.Write()` and `lumberjacklogger.Write()` do, will change it if we don't need

pkg/logger/ecs_error_log.go
pingleig: I am not sure if you can use `log.Printf`,  because we do set the output for `log`. Though if it is running in container we should be using the default and writes to stderr.

https://github.com/aws-observability/aws-otel-collector/blob/87ebdac1fb521c2d1033538b8161edf935fd9e34/pkg/logger/logger.go#L70-L85

gavindoudou: Thanks for the callout. I found the collector checked if it is running in container: https://github.com/aws-observability/aws-otel-collector/blob/87ebdac1fb521c2d1033538b8161edf935fd9e34/pkg/logger/logger.go#L41
In this way, I think we are okay to use `log.Printf` if we are running in the container. By the way, I am not sure if running in container is our only use case here.

pkg/logger/ecs_error_log.go
skyduo: the current case is that during the iteration, the len(queue) will change because we don't want 2 queue to implement it, that's why i need to use this way

pkg/logger/ecs_error_log.go
skyduo: since this ECS error reporter will only run in ECS, i think it will always run in container

pkg/logger/ecs_error_log.go
PettitWesley: +1. I hate 1 letter variable names. I prefer descriptive names.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+			l.size += n
+			l.queue = append(l.queue, e)
+		}
+		l.queue = l.queue[1:]
+	}
+	// delete the old errors if old errors plus new error exceed the max size
+	if l.size+len(newErrorLog) > l.ErrorFileMaxSize {
+		l.rotate(newErrorLog)
+	}
```
PettitWesley, line 79: So as we talked about on Wednesday, I think it's best if you check the size before you write the file. Why write all errors to the file first and then check size second?

Why shouldn't the algorithm be:
```
create a message queue;
 when (receive a new error log message) {
    calculate all message size;
    truncate the message to 1KB first;
    
    while (all message size > 1KB) {
        remove the oldest message;
    }
    
    if (message is already in message queue) {
        remove old message
    } 
    add new message into message queue;
    
    delete the local file;
    write all messages in message queue to local file
}
```

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,188 @@
...
+	if l.errorFile == nil {
+		if err := l.openExistingOrNewErrorFile(); err != nil {
+			log.Printf("could not open error log file when write new error, err: %v", err)
+		}
+	}
+	newErrorLog := []byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+		newError.Time, newError.Level, newError.Caller, newError.Message))
+
+	if len(newErrorLog) > l.ErrorFileMaxSize {
+		log.Printf("error size exceed the max size of error file")
```
PettitWesley, line 50: In this and the other error print statements, could we add something to make it clear that these come from the error reporter?

May be something like:


```
log.Printf("[ecs error reporter] error size exceed the max size of error file")
```

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,188 @@
...
+		log.Printf("could not truncate error log file during rotation, err: %v", err)
+	}
+
+	errorNum := len(l.queue)
+	for i := 0; i < errorNum; i++ {
+		currentError := l.queue[0]
+		currentErrorLog := []byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+			currentError.Time, currentError.Level, currentError.Caller, currentError.Message))
+		if l.size+len(newErrorLog) > l.ErrorFileMaxSize {
+			l.size -= len(currentErrorLog)
```
PettitWesley, line 107: I think the logic here is wrong...

PettitWesley: I think there is copy paste mistake here with `newErrorLog` being used.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,188 @@
...
+
+	errorNum := len(l.queue)
+	for i := 0; i < errorNum; i++ {
+		e := l.queue[0]
+		if e.Time.Add(time.Duration(l.ErrorLogMaxAge) * time.Hour).After(currentTime) {
+			n, err := file.Write([]byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+				e.Time, e.Level, e.Caller, e.Message)))
+			if err != nil {
+				log.Printf("could not write error into error log file during process timeout, err: %v", err)
+			}
```
PettitWesley, line 141: One thing I am noticing in this code. There are 3 functions which write to the file. So the code for writing is duplicated 3 times. That is not ideal from an organization and maintainability standpoint. 

It would be ideal if each of these functions only manipulates the queue of messages. And then there is one separate function for writing all contents of the queue to the file. And then each function can call that function to write.

pkg/logger/ecs_error_log.go
skyduo: because there is a case that an old error is same as the new one and file size plus new error size exceed the max limit, and if we iterate the queue first, we don't need to write the old one into the file and the new one will directly add to the file.
if an old error is same as the new one and we check the total size first, it will count the same error size twice and delete another error from file.
is this logic correct?

pkg/logger/ecs_error_log.go
skyduo: why there is a logic issue? for example we iterate 3 errors in queue, when `i = 0`, we found the `l.size+len(newErrorLog)` exceed the max size, we just minus the first error length from `l.size` and don't write it into the file. when `i=1` and `i=2`, we found that now `l.size+len(newErrorLog)` smaller than max size, so we still write the rest 2 errors into the file.
is this logic correct?

pkg/logger/ecs_error_log.go
skyduo: but you can see that for those 3 functions, though they all write to the file, it has some different logic during the iteration of the queue

pkg/logger/ecs_error_log.go
skyduo: okay, will change like that

pkg/logger/ecs_error_log.go
PettitWesley: I see. Good point. But still you can/should do the write to file last.

I think you are saying it should be:

```
create a message queue;
 when (receive a new error log message) {
    calculate all message size;
    truncate the message to 1KB first;

    if (message is already in message queue) {
        remove old message
    } 

   add new message into message queue;
    
    while (all message size > 1KB) {
        remove the oldest message;
    }
    
    delete the local file;
    write all messages in message queue to local file
}
```

pkg/logger/ecs_error_log.go
PettitWesley: > but you can see that for those 3 functions, though they all write to the file, it has some different logic during the iteration of the queue

Yeah and that's exactly my point. The custom logic for iterating over the queue and manipulating it should be in each function. Each function does its job to manipulate the queue. It gets the queue in the correct final state. And then you call a common utility function to write the contents of the queue to the file.

pkg/logger/ecs_error_log.go
PettitWesley: I see. Ok I think your logic is technically correct, but I think its a bit confusing and not ideal.

It'd be simpler and easier to read if the logic was:

```
append new error log to end of queue

while size of queue is greater than max size:
     remove first log from queue (oldest log)
    
```

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,188 @@
...
+	l.queue = append(l.queue, newError)
+	n, err := l.errorFile.Write(newErrorLog)
+	if err != nil {
+		log.Printf("could not write new error into error log file, err: %v", err)
+	}
+	l.size += n
+}
+
+// rotate func could delete the old logs to make sure error log file less than 1 KB
+func (l *ECSErrorLogger) rotate(newErrorLog []byte) {
```
PettitWesley, line 90: This function doesn't actually use the newErrorLog I think? Like you just use the length- so make the argument be the length instead.

pingleig (commented):
There are too many inline comments and the latest commits are very different from the original version so I will just write it outside.

- the hook logic is not right,  we only do the lumberjack hook when it's not running in container, so when running on ECS, `lumberjackLogger ` would always be nil and there will no error send to `errorLogChannel`. I think you should define your hook and there is no performance issue as the two hooks never run at same time.

```go
// in main
var zapHooks []zap.Hook // I didn't check the right type, this is just pseudo code

if logger.ReportECSErrorMessage() {
     ecsErrorLogger := logger.NewECSErrorLogger() // put all the env logic inside logger packege
     zapHooks = append(zapHooks,  ecsErrorLogger.Hook())
    go ecsErrorLogger.Run() // assume you don't return/care about error
}

if lumberHook != nil {
      zapHooks = append(zapHoosk, lumberHook)
}

params.LoggingOptions = []zap.Options(zap.Hooks(zapHooks...))
```

- you don't need a global `errorChannel`, you can put it in your `ECSErrorLogger` struct and initialize in your `NewECSErrrorLogger` func. 
- All the env lookup logic should be in this factory func as well, also you can return an error if you want to check env value is valid, e.g. you have write permission to the error log path.
- btw: ~~Are you expecting the file in a path mounted from host into the container? Otherwise the log inside container's own write layer is not that easy to collect and ship to somewhere else.~~ the description mentioned it will be mounted by ecs agent.

```go
func NewECSErrorLogger() *ECSErrorLogger {
      // do all the os.Getenv here
     return &ECSErrorLogger{
          errorChan: make([] ....)
     }
}

func (l *ECSErrorLogger) Run() { // ECSErrorReporter
      for {
               select {
                    case newErr := <- l.errorChan:
                               l.Write(newErr)
      // ....
}
}
```

PettitWesley (approved):
I think this mostly looks good to me. I think @pingleig recent comments mostly made sense to me and are worth implementing: https://github.com/aws-observability/aws-otel-collector/pull/503#pullrequestreview-669594373

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,218 @@
...
+		case <-l.ticker.C:
+			l.processTimeout()
+		}
+	}
+}
+
+func (l *ECSErrorLogger) Write(newError zapcore.Entry) {
+	if l.errorFile == nil {
+		if err := l.openExistingOrNewErrorFile(); err != nil {
+			log.Printf("[ecs error reporter] could not open error log file when write new error, err: %v", err)
```
pingleig, line 75: If open/create new file fails, shouldn't we abort write?

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,218 @@
...
+
+	errorChannel chan zapcore.Entry
+	queue        []zapcore.Entry
+	ticker       time.Ticker
+	errorFile    *os.File
+	size         int
+}
+
+func NewECSErrorLogger() *ECSErrorLogger {
+	ecsErrorFilePath := os.Getenv("STATUS_MESSAGE_FILE_PATH")
```
pingleig, line 41: Can you make those env var key constant, also do we need a prefix like `ECS_ERROR_LOG`, the `STAUS_MESSAGE_` seems too general.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,218 @@
...
+		log.Printf("[ecs error reporter] could not write new error into error log file, err: %v", err)
+	}
+	l.size += n
+}
+
+// rotate func could delete the old logs to make sure error log file less than 1 KB
+func (l *ECSErrorLogger) rotate(newErrorLogSize int) {
+	file, err := os.OpenFile(l.ErrorFilePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0777)
+	if err != nil {
+		log.Printf("[ecs error reporter] could not open error log file during rotation, err: %v", err)
```
pingleig, line 123: If you can't open the file and the function continues, `file.Truncate` would get a nil pointer .

pkg/logger/logger.go
```diff
@@ -67,6 +67,16 @@ func GetLumberHook() func(e zapcore.Entry) error {
 	return nil
 }
 
+// put error log into a channel for ECS when higher than error level
```
pingleig, line 70: You don't need to create a new func, you can just add a new method to `ECSErrorLogger` and move it to `ecs_error_log.go`.

```go
func (e *ECSErrorLogger) GetErrrorHook() ....
````

**Review by pingleig — COMMENTED** [review]
- For error handling, after logging the error you should return to avoid nil pointer in following logic.
- Run `goimports` to sort the imports.

File: pkg/logger/ecs_error_log.go (hunk 37 above)
Comment (skyduo) [reply pkg/logger/ecs_error_log.go:75]: > * Run `goimports` to sort the imports.

which import i'm not sort? if you mean the imports in the `logger.go` file, that one originally sort in this way.

File: pkg/logger/ecs_error_log.go (hunk 38 above)
Comment (skyduo) [reply pkg/logger/ecs_error_log.go:41]: sure

File: pkg/logger/ecs_error_log.go (hunk 39 above)
Comment (skyduo) [reply pkg/logger/ecs_error_log.go:123]: yeah, i just add return if open file met some error

File: pkg/logger/logger.go (hunk 40 above)
Comment (skyduo) [reply pkg/logger/logger.go:70]: sure

File: pkg/logger/ecs_error_log.go (hunk 37 above)
Comment (skyduo) [reply pkg/logger/ecs_error_log.go:75]: i just add return if open file met error

**Comment by skyduo** [comment]
Hey @pingleig , is there any other changes we need?

File: cmd/awscollector/main.go [hunk 41]
```diff
@@ -48,8 +47,22 @@ func main() {
...
+
 	// init lumberFunc for zap logger
 	lumberHook := logger.GetLumberHook()
+	if lumberHook != nil {
+		zapHooks = append(zapHooks, zap.Hooks(lumberHook))
+	}
+
+	ecsErrorLogFilePath := os.Getenv("ECS_ERROR_LOG_FILE_PATH")
+	// init an ECS Error Logger and a go routine for error reporting when the STATUS_MESSAGE_FILE_PATH is set
+	if ecsErrorLogFilePath != "" {
```
pingleig, line 60: Since you only use `ecsErrorLogFilePath` to check if the logger is enable, you can write `if os.Getenv("ECS_ERROR_LOG_FILE_PATH") != ""` here, also it might be better to write a method in the `logger` package like `logger.ECSErrorLogEnabled()` to avoid writing the constant string `ECS_ERROR_LOG_FILE_PATH` in two places.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,238 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package logger
+
+import (
+	"fmt"
+	"go.uber.org/zap/zapcore"
```
pingleig, line 20: goimports.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,238 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+)
+
+type ECSErrorLogger struct {
```
pingleig, line 28: I think you can paste your PR description here so people would know what it's for when they look at it later on. e.g.

```go
// ECSErrorLogger writes to a file mounted into container from host by ECS Agent.
// The file has a limit of 1MB and only contains error level messages.
// A zap hook is injected to intercept error level log entries 
// if the env var `ECS_ERROR_LOG_FILE_PATH` is valid.
// ....
type ECSErrorLogger struct {
}
```

pingleig (commented):
Other than `goimports`, call helper method in `main`, LGTM.

I'll leave the approval and merge to @mxiamxia

pingleig:
btw: You might want to create a new PR with single commit, the PR diff is not large but it has a bunch of merge commits and minor fixes.

gavindoudou (approved):
LGTM, agree it's better to have a new PR.

JohnWu20 (approved):

hossain-rayhan (approved):

pingleig:
The build is failing because of https://github.com/aws-observability/aws-otel-collector/pull/534 create a new PR from latest main branch should fix it, it is caused by dependent bot.

skyduo:
Hey @mxiamxia @pingleig @gavindoudou @hossain-rayhan @PettitWesley @JohnWu20 here is the new PR for final review, https://github.com/aws-observability/aws-otel-collector/pull/538, please take a look and let me know if anything we need to change. It will be merged after the appsec review with the health check.

pingleig:
Close it for #538

PR #515, opened 2021-05-12
Fix Docker command

willarmiros:
**Documentation:** The path to the docker config no longer exists, this change updates it to the correct one: https://github.com/aws-observability/aws-otel-collector/blob/main/examples/docker/config-test.yaml


mxiamxia (approved):
Thanks!

PR #524, opened 2021-05-21
chore: use os.Chown instead of calling chown command

haojhcwa:
**Description:**
Use os.Chown instead of calling chown command when changing ownership of
the aoc config files. This would improve the future portability of the code.
Also fix the aoc running user to be `aoc` since the environment override is not 
very useful but create more complicated use cases.

**Testing:**
Unit test added and tested locally

**Documentation:**
No documentation change needed since no behavior change are expected.

### DIFF
--- a/pkg/userutils/userutil_linux.go
+++ b/pkg/userutils/userutil_linux.go
@@ -18,19 +18,22 @@
 package userutils
 
 import (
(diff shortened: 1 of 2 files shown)


mxiamxia (approved):
LGTM! Thx

PR #534, opened 2021-06-04
Remove the unexpected import in go.mod

mxiamxia:
Remove the duplicate go libs in go mod file

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,9 +22,8 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.7.0
 	go.opentelemetry.io/collector v0.27.0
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -22,9 +22,8 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.7.0
 	go.opentelemetry.io/collector v0.27.0
-  github.com/opencontainers/runc-1.0.0-rc95
```
pingleig, line 25: I remember we were. using it for run as user? https://github.com/aws-observability/aws-otel-collector/blob/main/pkg/userutils/userutil_linux.go#L27

pingleig (approved):
checked w/ Min, the run import is duplication from dependant bot

PR #538, opened 2021-06-09
Add new ECS error reporting feature in AOC

skyduo:
**Description:** 
This is the new error reporting feature for ECS. Except sending all error logs to the original destinations, we need to also send error or higher level log messages to another single error log file (the file path configurable later) and keep it less than 1 KB, so that ECS agent could mount this path and read the error log file. When the error log file exceed the max size (1KB), we want to delete the oldest log message to release the space. Also we want to rotate the logs by repeatedly delete the expired logs (the expired time is also configurable later). 

**Testing:** 
`make build` and run the executable file with a config file with some errors, and succeed with 2 config errors write into the error log file.

**Previous Pull Request:** 
https://github.com/aws-observability/aws-otel-collector/pull/503

### DIFF
--- a/cmd/awscollector/main.go
+++ b/cmd/awscollector/main.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"time"
 
 	"github.com/aws-observability/aws-otel-collector/pkg/config"
 	"github.com/aws-observability/aws-otel-collector/pkg/defaultcomponents"
@@ -46,8 +47,22 @@ func main() {
 		log.Fatalf("failed to build components: %v", err)
 	}
 
+	var zapHooks []zap.Option
+
 	// init lumberFunc for zap logger
 	lumberHook := logger.GetLumberHook()
+	if lumberHook != nil {
+		zapHooks = append(zapHooks, zap.Hooks(lumberHook))
+	}
+
+	// init an ECS Error Logger and a go routine for error reporting when the STATUS_MESSAGE_FILE_PATH is set
+	if os.Getenv("STATUS_MESSAGE_FILE_PATH") != "" {
+		ecsErrorLogger := logger.NewECSErrorLogger()
+		ecsErrorHook := ecsErrorLogger.GetErrorHook()
+		zapHooks = append(zapHooks, zap.Hooks(ecsErrorHook))
+		ecsErrorLogger.Ticker = time.NewTicker(time.Minute)
+		go ecsErrorLogger.Run()
+	}
 
 	// set the collector config from extracfg file
 	if extraConfig != nil {
@@ -64,8 +79,8 @@ func main() {
 		Factories: factories,
 		BuildInfo: info,
 	}
-	if lumberHook != nil {
-		params.LoggingOptions = []zap.Option{zap.Hooks(lumberHook)}
+	if len(zapHooks) != 0 {
+		params.LoggingOptions = zapHooks
 	}
 	if err := run(params); err != nil {
 		logFatal(err)
--- a/pkg/logger/ecs_error_log.go
+++ b/pkg/logger/ecs_error_log.go
@@ -1,0 +1,250 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package logger
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"go.uber.org/zap/zapcore"
+)
+
+// ECSErrorLogger is used to send the error level log to a separated file and rotate
+// error logs when expired, ECS agent will mount the filepath and read it periodically
+type ECSErrorLogger struct {
+	ErrorLogFilePath    string
+	ErrorLogMaxAge      int
+	ErrorLogFileMaxSize int
+
+	errorChannel chan zapcore.Entry
+	queue        []zapcore.Entry
+	Ticker       *time.Ticker
+	errorFile    *os.File
+	size         int
+}
+
+// NewECSErrorLogger could new a struct based on the 3 environment variables:
+// error file path, error log TTL and error file max size
+func NewECSErrorLogger() *ECSErrorLogger {
+	ecsErrorLogFilePath := os.Getenv("STATUS_MESSAGE_FILE_PATH")
+	errorLogMaxAge, err := strconv.Atoi(os.Getenv("STATUS_MESSAGE_TTL"))
+	if err != nil || errorLogMaxAge <= 0 {
+		log.Fatalf("failed to get STATUS_MESSAGE_TTL: %v", err)
+	}
+	errorFileSize, err := strconv.Atoi(os.Getenv("STATUS_MESSAGE_MAX_BYTE_LENGTH"))
+	if err != nil || errorFileSize <= 0 {
+		log.Fatalf("failed to get STATUS_MESSAGE_MAX_BYTE_LENGTH: %v", err)
+	}
+	if errorFileSize > 1000 {
+		log.Printf("error log file size should not exceed 1KB, will use 1000 instead")
+		errorFileSize = 1000
+	}
+	return &ECSErrorLogger{
+		ErrorLogFilePath:    ecsErrorLogFilePath,
+		ErrorLogMaxAge:      errorLogMaxAge,
+		ErrorLogFileMaxSize: errorFileSize,
+		errorChannel:        make(chan zapcore.Entry, 100),
+	}
+}
+
+// Run func help select the different case between either write error log into file or
+// rotate the old logs in file based on TTL
+func (l *ECSErrorLogger) Run() {
+	defer l.Ticker.Stop()
+	for {
+		select {
+		case newError := <-l.errorChannel:
+			l.Write(newError)
+		case <-l.Ticker.C:
+			l.processTimeout()
+		}
+	}
+}
+
+// Write function could write the new error into the file, when the file size plus new error size
+// exceed the max limit, will call rotate function to delete the old log
+func (l *ECSErrorLogger) Write(newError zapcore.Entry) {
+	if l.errorFile == nil {
+		if err := l.openExistingOrNewErrorFile(); err != nil {
+			log.Printf("[ecs error reporter] could not open error log file when write new error, err: %v", err)
+			return
+		}
+	}
+	newErrorLog := []byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+		newError.Time, newError.Level, newError.Caller, newError.Message))
+
+	if len(newErrorLog) > l.ErrorLogFileMaxSize {
+		log.Printf("[ecs error reporter] error size exceed the max size of error file")
+		return
+	}
+
+	err := l.errorFile.Truncate(0)
+	if err != nil {
+		log.Printf("[ecs error reporter] could not truncate error log file when write new error, err: %v", err)
+		return
+	}
+	l.size = 0
+
+	errorNum := len(l.queue)
+	for i := 0; i < errorNum; i++ {
+		e := l.queue[0]
+		if e.Message != newError.Message {
+			n, err := l.errorFile.Write([]byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+				e.Time, e.Level, e.Caller, e.Message)))
+			if err != nil {
+				log.Printf("[ecs error reporter] could not write error into error log file, err: %v", err)
+				return
+			}
+			l.size += n
+			l.queue = append(l.queue, e)
+		}
+		l.queue = l.queue[1:]
+	}
+
+	if l.size+len(newErrorLog) > l.ErrorLogFileMaxSize {
+		l.rotate(len(newErrorLog))
+	}
+
+	l.queue = append(l.queue, newError)
+	n, err := l.errorFile.Write(newErrorLog)
+	if err != nil {
+		log.Printf("[ecs error reporter] could not write new error into error log file, err: %v", err)
+		return
+	}
+	l.size += n
+}
+
+// rotate func could delete the old logs to make sure error log file less than max limit
+func (l *ECSErrorLogger) rotate(newErrorLogSize int) {
+	file, err := os.OpenFile(l.ErrorLogFilePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0777)
+	if err != nil {
+		log.Printf("[ecs error reporter] could not open error log file during rotation, err: %v", err)
+		return
+	}
+
+	err = file.Truncate(0)
+	if err != nil {
+		log.Printf("[ecs error reporter] could not truncate error log file during rotation, err: %v", err)
+		return
+	}
+
+	errorNum := len(l.queue)
+	for i := 0; i < errorNum; i++ {
+		currentError := l.queue[0]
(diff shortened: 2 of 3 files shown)


pingleig (changes requested):

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,251 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"go.uber.org/zap/zapcore"
+)
+
+type ECSErrorLogger struct {
```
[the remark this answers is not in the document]
skyduo: sorry maybe i miss your comment, will add it

pkg/logger/ecs_error_log_test.go
```diff
@@ -0,0 +1,154 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"go.uber.org/zap/zapcore"
+
```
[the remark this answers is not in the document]
gavindoudou: Does you mean we are following the pattern for dividing into two parts here? I found some files here only have two parts.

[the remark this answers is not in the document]
skyduo: sure i will merge this two together, it was auto formed like this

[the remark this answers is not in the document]
pingleig: We are not enforcing linter tools here, otel repo has https://github.com/pavius/impi and does not allow splitting non local deps into two groups.

pkg/logger/ecs_error_log.go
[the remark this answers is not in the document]
pingleig: I think it's better to move the doc to go struct,  it's not package level doc and won't show up in go doc https://pkg.go.dev/github.com/asakermohd/aws-otel-collector/pkg/logger

pingleig (approved):
LGTM

pingleig:
ping @mxiamxia

mxiamxia:
Hi @skyduo , since we use main branch for prod release. These changes are on critical path of running ADOT collector. Do you think we can move this changes in a feature branch and we can merge it after enough testing later?

skyduo:
hey @mxiamxia ! sure, we can do that

skyduo:
hey @pingleig! that is a new health check feature in the upstream repo which is different from this error reporting feature, also @mxiamxia can we merge this pr?

pingleig:
@skyduo I think you need to sync w/ @jhnlsn or @alolita on merging this. We (me and Min) no longer actively work on ADOT.

cmd/awscollector/main.go
```diff
@@ -17,6 +17,7 @@ package main // import "aws-observability.io/collector/cmd/awscollector"
 
 import (
 	"fmt"
+	"go.uber.org/zap"
```
Aneurysm9, line 20:
```suggestion
```

cmd/awscollector/main.go
```diff
@@ -27,7 +28,6 @@ import (
 	"github.com/aws-observability/aws-otel-collector/tools/version"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/service"
-	"go.uber.org/zap"
```
Aneurysm9, line 30: Restore this.  It doesn't belong with the stdlib imports.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+	size         int
+}
+
+// NewECSErrorLogger could new a struct based on the 3 environment variables:
+// error file path, error log TTL and error file max size
+func NewECSErrorLogger() *ECSErrorLogger {
+	ecsErrorLogFilePath := os.Getenv("STATUS_MESSAGE_FILE_PATH")
+	errorLogMaxAge, err := strconv.Atoi(os.Getenv("STATUS_MESSAGE_TTL"))
+	if err != nil || errorLogMaxAge <= 0 {
+		log.Fatalf("failed to get STATUS_MESSAGE_TTL: %v", err)
```
Aneurysm9, line 49: I don't think that failure to set up this logging handler should be fatal to the application.  Have it return an error that can be logged but don't use `log.Fatal`.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	newErrorLog := []byte(fmt.Sprintf("{%+v, Level:%+v, Caller:%+v, Message:%+v}\r\n",
+		newError.Time, newError.Level, newError.Caller, newError.Message))
+
+	if len(newErrorLog) > l.ErrorLogFileMaxSize {
+		log.Printf("[ecs error reporter] error size exceed the max size of error file")
+		return
+	}
+
+	err := l.errorFile.Truncate(0)
```
Aneurysm9, line 98: Why is this truncating the log file on every error?

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+				e.Time, e.Level, e.Caller, e.Message)))
+			if err != nil {
+				log.Printf("[ecs error reporter] could not write error into error log file, err: %v", err)
+				return
+			}
+			l.size += n
+			l.queue = append(l.queue, e)
+		}
+		l.queue = l.queue[1:]
+	}
```
Aneurysm9, line 119: What is this doing?  It looks like it is trying to write back into the error log file all of the errors that were just truncated from it.

Aneurysm9 (changes requested):
This seems dangerous and woefully underexplained.  What does it do, why is it necessary, why does it need to do what it does the way it does?

pkg/logger/ecs_error_log.go
skyduo: we need to go through all the error logs and delete the expired ones or the oldest one if the file size exceed and write all the rest of errors in the file

pkg/logger/ecs_error_log.go
skyduo: since we need to go through and rotate all the errors if expired or exceed the file size

pkg/logger/ecs_error_log.go
skyduo: since this is a requirement for the error reporting feature, if user choose to use it, they have to provide a `STATUS_MESSAGE_TTL`

cmd/awscollector/main.go
skyduo: sure, thanks!

pkg/logger/ecs_error_log.go
Aneurysm9: Or there can be a reasonable default.  This is should not be fatal to the process.

pkg/logger/ecs_error_log.go
Aneurysm9: How does this behave with a high rate of errors being logged?  What impact does it have on collector performance?  Will the consumer of this data even be likely to see any/valid/complete data if the new data aren't prepared until after the old data are removed?

pkg/logger/ecs_error_log.go
Aneurysm9: Isn't that what the `rotate()` method does? Why can't this method simply keep track of the current file size, write the error if it wouldn't exceed the file size, and call `rotate()` if it does?

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+		ErrorLogMaxAge:      errorLogMaxAge,
+		ErrorLogFileMaxSize: errorFileSize,
+		errorChannel:        make(chan zapcore.Entry, 100),
+	}
+}
+
+// Run func help select the different case between either write error log into file or
+// rotate the old logs in file based on TTL
+func (l *ECSErrorLogger) Run() {
+	defer l.ticker.Stop()
```
Aneurysm9, line 70: What starts this ticker?  I can't seem to find anything setting it running, which means that `processTimeout()` will never be invoked.

pkg/logger/ecs_error_log.go
skyduo: we want to do one thing that check the new error write into the file not the same as the old error, so if the new error is the same one, we just update the timestamp of the error stored in the queue

pkg/logger/ecs_error_log.go
skyduo: yeah good point, i didn't notice that maybe i delete the new ticker line by mistake, will add it back

pkg/logger/ecs_error_log.go
skyduo: we had a talk before about it, if customer choose to use this feature and not provide a ttl parameter, we will exit the process

pkg/logger/ecs_error_log.go
skyduo: actually customer will not directly use the error file, just like in the description, ECS agent will mount the file path and read the file, we will have a performance testing with agent team next year after they implement their code part.

cmd/awscollector/main.go
```diff
@@ -46,8 +47,22 @@ func main() {
...
+	if lumberHook != nil {
+		zapHooks = append(zapHooks, zap.Hooks(lumberHook))
+	}
+
+	// init an ECS Error Logger and a go routine for error reporting when the STATUS_MESSAGE_FILE_PATH is set
+	if os.Getenv("STATUS_MESSAGE_FILE_PATH") != "" {
+		ecsErrorLogger := logger.NewECSErrorLogger()
+		ecsErrorHook := ecsErrorLogger.GetErrorHook()
+		zapHooks = append(zapHooks, zap.Hooks(ecsErrorHook))
+		ecsErrorLogger.Ticker = time.NewTicker(time.Minute)
```
Aneurysm9, line 63: Should this interval be configurable?  Why is the main routine responsible for creating the ticker rather than the logger?  Not creating the ticker when the logger is created sounds like an incident waiting to happen.

cmd/awscollector/main.go
```diff
@@ -46,8 +47,22 @@ func main() {
...
+
 	// init lumberFunc for zap logger
 	lumberHook := logger.GetLumberHook()
+	if lumberHook != nil {
+		zapHooks = append(zapHooks, zap.Hooks(lumberHook))
+	}
+
+	// init an ECS Error Logger and a go routine for error reporting when the STATUS_MESSAGE_FILE_PATH is set
+	if os.Getenv("STATUS_MESSAGE_FILE_PATH") != "" {
+		ecsErrorLogger := logger.NewECSErrorLogger()
```
Aneurysm9, line 60: If the main routine is going to be responsible for determining whether this logger should be used based on environment variables then it should provide the entire configuration to the logger.  In other words, exactly one of the main routine and the logger should be aware of the existence of these environment variables.  I'm not sure that putting it in the main routine is the correct choice, but I am sure that splitting it between the two is the wrong choice.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+
+	"go.uber.org/zap/zapcore"
+)
+
+// ECSErrorLogger is used to send the error level log to a separated file and rotate
+// error logs when expired, ECS agent will mount the filepath and read it periodically
+type ECSErrorLogger struct {
+	ErrorLogFilePath    string
+	ErrorLogMaxAge      int
+	ErrorLogFileMaxSize int
```
Aneurysm9, line 34: Do these need to be exported?  It doesn't seem that they do.  Even if the main routine is to be responsible for extracting configuration from the environment, that should be provided to the initializer and these fields should remain unexported so that they cannot be modified externally while the process is running.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+
+	errorChannel chan zapcore.Entry
+	queue        []zapcore.Entry
+	Ticker       *time.Ticker
```
Aneurysm9, line 38: This also does not need to be exported.  The ticker should be started in the initializer or in `Run()`.

pkg/logger/ecs_error_log.go
Aneurysm9: Who is "we"? Who from ADOT approved that decision?  This is contrary to the design principles followed by most OTel components and may be surprising to users.  I don't think it is the correct decision when a default value can be provided.  How is the user to know what value to use for this TTL?  If it is something that the control plane is supposed to inject automatically then we can set a default here that mirrors what should be injected.  If it is something that we provide recommendations on through documentation then we can set the default to match that recommendation.

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+		log.Fatalf("failed to get STATUS_MESSAGE_TTL: %v", err)
+	}
+	errorFileSize, err := strconv.Atoi(os.Getenv("STATUS_MESSAGE_MAX_BYTE_LENGTH"))
+	if err != nil || errorFileSize <= 0 {
+		log.Fatalf("failed to get STATUS_MESSAGE_MAX_BYTE_LENGTH: %v", err)
+	}
+	if errorFileSize > 1000 {
+		log.Printf("error log file size should not exceed 1KB, will use 1000 instead")
+		errorFileSize = 1000
+	}
```
Aneurysm9, line 58:
```suggestion
	if errorFileSize > 1024 {
		log.Printf("error log file size should not exceed 1KB, will use 1024 instead")
		errorFileSize = 1024
	}
```

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+		ErrorLogMaxAge:      errorLogMaxAge,
+		ErrorLogFileMaxSize: errorFileSize,
+		errorChannel:        make(chan zapcore.Entry, 100),
+	}
+}
+
+// Run func help select the different case between either write error log into file or
+// rotate the old logs in file based on TTL
+func (l *ECSErrorLogger) Run() {
+	defer l.Ticker.Stop()
```
Aneurysm9, line 70: This function never returns so this deferred statement will never be executed.

pkg/logger/ecs_error_log.go
Aneurysm9: If nothing is using the data written by this logger then we should take the time to take a step back and address the concerns that the current implementation presents.  We definitely shouldn't ship it before we have assessed its performance and behavior under stress.

pkg/logger/ecs_error_log.go
Aneurysm9: Aren't there ways of achieving that objective that 1) do not truncate and rewrite the log file with every error and 2) do not duplicate (or triplicate) logic?

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+			l.size += n
+			l.queue = append(l.queue, e)
+		}
+		l.queue = l.queue[1:]
+	}
+
+	if l.size+len(newErrorLog) > l.ErrorLogFileMaxSize {
+		l.rotate(len(newErrorLog))
+	}
```
Aneurysm9, line 123: Can't we know this before we attempt to rewrite the log file above?  This can result in truncating and rewriting the log file twice per error message with messages as small as 512 bytes (smaller in practice since the formatting around the message is ~85 bytes at minimum).

pkg/logger/ecs_error_log.go
```diff
@@ -0,0 +1,250 @@
...
+
+	l.queue = append(l.queue, newError)
+	n, err := l.errorFile.Write(newErrorLog)
+	if err != nil {
+		log.Printf("[ecs error reporter] could not write new error into error log file, err: %v", err)
+		return
+	}
+	l.size += n
```
Aneurysm9, line 131:
```suggestion
	l.size += n
	if err != nil {
		log.Printf("[ecs error reporter] could not write new error into error log file, err: %v", err)
		return
	}
```

Those bytes were written to the target regardless whether there was an error.

Aneurysm9 (changes requested):
I cannot recommend that this PR be merged at this time.  The design needs to be reevaluated to minimize unnecessary rewriting of the log file.  Given the tiny amount of space we're talking about here storing some additional data could go a long way.  A set of messages rather than a list would make it easier to determine whether a given message is already in the file, and thus whether the file may or may not need to be rewritten.  A double buffer could avoid leaving a truncated file sitting around while determining what to write to it.  Restructuring the code a little bit could avoid significant duplication of code that is likely to lead to duplication of errors.

pkg/logger/ecs_error_log.go
PettitWesley: I did the high level design for this... and while the backend will always set these values, it doesn't hurt to have a default.

PR #540, opened 2021-06-11
Use examples/docker Dir In Make File For Docker Compose Up

sethAmazon:
**Description:** Changed make file to look at dir examples/docker for docker-compose file. 
Ex. Fixing a bug - Make file is looking at a dir without a docker file. 

**Testing:** docker-compose up -d triggered starting of docker containers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ docker-run:
 
 .PHONY: docker-compose
 docker-compose:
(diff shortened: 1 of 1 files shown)


pingleig (approved):
LGTM (TBH I never tried `make docker-compose` 😅

PR #550, opened 2021-06-23
Enable container insights infra metrics for EKS

pxaws:
**Description:** 
* Sync the dependency with the otel upstream
* Enable the container insight receiver from upstream (it is not ready yet and we will need to update the version later)
* Add the doc for EKS Container Insights

**Testing:** 
```
make build
```


pingleig (approved):
LGTM

PR #582, opened 2021-07-20
Update OTel deps to v0.30.0 and update sample app

mxiamxia:
Update OTel deps to v0.30.0 and update sample app


pingleig (approved):
LGTM

Issue #618, opened 2021-08-19
Allow collector to run as a different user


rafaelpereyra:
Description: While fixing #524 , the ability to use an environment variable (`AOT_RUN_USER`) was removed from the code forcing the collector to run with the default user (`aoc`).

Use case: 

- System administrators might be interested in controlling the user running this collector to match any naming convention or UID/GID patterns.

Solution: 

Partially revert the code to use the function `getCustomUser` to populate the variable `aocUserName` and default to `aoc` otherwise.

PR #620, opened 2021-08-19
Added back the ability to run the collector with a different user

rafaelpereyra:
**Description:** 
Added back the possibility to run the collector with a different user than default `aoc`.

Reused the previous environment variable from version 0.9.0 `AOT_RUN_USER`

**Link to tracking Issue:**
https://github.com/aws-observability/aws-otel-collector/issues/618

**Testing:** 
Added a test case. Validated is properly working on an Amazon Linux 2 machine (Cloud9).

**Documentation:**

In the configuration of the agent, add the option to override the default user passing the environment variable 'AOT_RUN_USER'

i.e.: 

```
export AOT_RUN_USER=root
sudo --preserve-env=AOT_RUN_USER /opt/aws/aws-otel-collector/bin/aws-otel-collector --config ./config.yaml
```

### DIFF
--- a/pkg/userutils/userutil_linux.go
+++ b/pkg/userutils/userutil_linux.go
@@ -28,7 +28,7 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-var aocUserName = "aoc"
+var defaultUser = "aoc"
 var defaultInstallPath = "/opt/aws/aws-otel-collector/"
 
 type ChownFunc func(name string, uid, gid int) error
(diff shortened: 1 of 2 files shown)


mxiamxia:
Hi, Thanks for contributing, but we disabled running as different user on purpose. It is not suggested from the our security team. 

https://github.com/aws-observability/aws-otel-collector/commit/bf8b2b0678f8d7683072f47c6c37bcd4a22cdcef

PR #625, opened 2021-08-25
bump version to v0.12.0

straussb:
bump version to v0.12.0


alolita (commented):
These release notes do not list upstream changes. Can we add a link to the upstream release notes for the Collector here. Thanks!

PR #675, opened 2021-10-19
Add Nightly Build Unit Test

sethAmazon:
**Description:**
Adding unit test nightly

**Testing:** 
Github action


anuraaga:
@sethAmazon What's the difference between the current CI, unit tests that run on every commit (we shouldn't need to duplicate them into a nightly build) and soaking tests, which seem to currently be failing but should those be fixed instead of introducing yet another scheduled build?

Issue #699, opened 2021-11-01
E2E tests have failed for distribution packages having too many versions. 


sethAmazon:
**Is your feature request related to a problem? Please describe.**
E2E tests have failed for distribution packages having too many versions. 

**Describe the solution you'd like**
github action to clean old versions 

**Describe alternatives you've considered**
None

PR #700, opened 2021-11-01
Add Clean Distributor Document Versions (#699)

sethAmazon:
**Description:** 
Can only have a certain amount of distribution document versions. Clean old versions.

**Link to tracking Issue:** 
#699 

**Testing:** 
Created a new distribution type and run the go for the new file.

### DIFF
--- a/.github/workflows/resources-clean.yml
+++ b/.github/workflows/resources-clean.yml
@@ -35,3 +35,8 @@ jobs:
       
       - name: Clean resources which are created before yesterday
         run: sh tools/workflow/clean-terraform-resources.sh
+
+      - name: Clean distributor documents
+        run: go run e2etest/clean/clean_distributor.go
+        env:
+          AWS_SDK_LOAD_CONFIG: true
--- a/e2etest/clean/clean_distributor.go
+++ b/e2etest/clean/clean_distributor.go
@@ -1,0 +1,94 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package main
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"log"
+	"strconv"
+)
+
+type documentVersionTuple struct {
+	str *string
+	num int
+}
+
+func main() {
+	// set up aws go sdk ssm client
+	testSession, err := session.NewSession()
+	if err != nil {
+		log.Fatalf("Error creating session %v", err)
+	}
+	ssmClient := ssm.New(testSession)
+
+	//get list of documents with versions
+	documentName := "testAWSDistroOTel-Collector"
+	listDocumentVersionInput := ssm.ListDocumentVersionsInput{Name: &documentName}
+	listDocumentVersionsOutput, err := ssmClient.ListDocumentVersions(&listDocumentVersionInput)
+	if err != nil {
+		log.Fatalf("Error getting document list %v", err)
+	} else if len(listDocumentVersionsOutput.DocumentVersions) <= 1 {
+		log.Print("No document versions to delete")
+		return
+	}
+
+	versionTuples := make([]documentVersionTuple, 0, len(listDocumentVersionsOutput.DocumentVersions))
+	var largestVersion *documentVersionTuple
+	for _, documentVersion := range listDocumentVersionsOutput.DocumentVersions {
+		versionTuple := toVersionTuple(documentVersion)
+		if largestVersion == nil || versionTuple.num > largestVersion.num {
+			largestVersion = &versionTuple
+		}
+		versionTuples = append(versionTuples, versionTuple)
+	}
+
+	//this is here to remove compiler warning that largest version may be nil
+	if largestVersion == nil {
+		log.Fatalf("Largest version pointer is nil. Not sure how we got here")
+	}
+
+	//do not delete latest version and make it default
+	updateDocumentDefaultVersionInput := ssm.UpdateDocumentDefaultVersionInput{Name: &documentName, DocumentVersion: largestVersion.str}
+	_, err = ssmClient.UpdateDocumentDefaultVersion(&updateDocumentDefaultVersionInput)
+	if err != nil {
+		log.Fatalf("Error updating default document %v", err)
+	}
+	log.Printf("Updated default version to %d", largestVersion.num)
+
+	for _, version := range versionTuples {
+		if largestVersion.num == version.num {
+			continue
+		}
+		deleteDocumentInput := ssm.DeleteDocumentInput{Name: &documentName, DocumentVersion: version.str}
+		_, err := ssmClient.DeleteDocument(&deleteDocumentInput)
+		if err != nil {
+			log.Fatalf("Error deleting document version %d %v", version.num, err)
+		}
+		log.Printf("Document version deleted %d", version.num)
+	}
+}
+
+func toVersionTuple(info *ssm.DocumentVersionInfo) documentVersionTuple {
+	versionNumber, err := strconv.Atoi(*info.DocumentVersion)
+	if err != nil {
+		log.Fatalf("Error parsing version number %v", err)
+	}
+	return documentVersionTuple{
+		str: info.DocumentVersion,
+		num: versionNumber,
+	}
+}
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ ALL_SRC := $(shell find . -name '*.go' \
 							-not -path './build/*' \
 							-not -path './tools/linters/*' \
 							-not -path './tools/release/*' \
+							-not -path './e2etest/clean/*' \
 							-type f | sort)
 # ALL_PKGS is the list of all packages where ALL_SRC files reside.
 ALL_PKGS := $(shell go list $(sort $(dir $(ALL_SRC))))


e2etest/clean/clean_distributor_test.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"sort"
+	"strconv"
+	"testing"
+)
+
+func TestCleanDistributor(t *testing.T) {
```
jefchien, line 27: Why is this a test?

.github/workflows/resources-clean.yml
```diff
@@ -35,3 +35,6 @@ jobs:
       
       - name: Clean resources which are created before yesterday
         run: sh tools/workflow/clean-terraform-resources.sh
+
+      - name: Clean distributor documents
+        run: go test e2etest/clean/clean_distributor_test.go -v
```
Aneurysm9, line 40: I'm not a fan of using `go test` to invoke functional scripts.  Is there any reason this needs to be done this way rather than having a `package main` and invoking it with `go run`?

e2etest/clean/clean_distributor_test.go
```diff
@@ -0,0 +1,74 @@
...
+	//set up aws go sdk config
+	awsConfig := aws.Config{
+		Region: aws.String("us-west-2"),
```
Aneurysm9, line 30: This should come from the environment.  If we set `AWS_REGION` in the workflow step's environment this line can be dropped entirely.

e2etest/clean/clean_distributor_test.go
```diff
@@ -0,0 +1,74 @@
...
+	ssmClient := ssm.New(testSession, &awsConfig)
+
+	//get list of documents with versions
+	documentName := "testAWSDistroOTel-Collector"
+	listDocumentVersionInput := ssm.ListDocumentVersionsInput{Name: &documentName}
+	listDocumentVersionsOutput, err := ssmClient.ListDocumentVersions(&listDocumentVersionInput)
+	if err != nil {
+		t.Fatalf("Error getting document list %s", err.Error())
+	}
+	t.Logf("Document Versions %s", listDocumentVersionsOutput.String())
```
Aneurysm9, line 46: Is this necessary?

e2etest/clean/clean_distributor_test.go
```diff
@@ -0,0 +1,74 @@
...
+	//get document versions
+	var versions []int
+	for i := range listDocumentVersionsOutput.DocumentVersions {
+		//can't delete the default version
+		if *listDocumentVersionsOutput.DocumentVersions[i].IsDefaultVersion {
+			continue
+		}
+		version, err := strconv.Atoi(*listDocumentVersionsOutput.DocumentVersions[i].DocumentVersion)
```
Aneurysm9, line 54:
```suggestion
	for _, docVer := range listDocumentVersionsOutput.DocumentVersions {
		//can't delete the default version
		if *docVer.IsDefaultVersion {
			continue
		}
		version, err := strconv.Atoi(*docVer.DocumentVersion)
```

Access is read-only and we don't need the index, so this is simpler.

e2etest/clean/clean_distributor_test.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			t.Fatalf("Error getting document list %s", err.Error())
+		}
+		versions = append(versions, version)
+	}
+	//sort document versions from oldest to newest
+	sort.Ints(versions[:])
```
Aneurysm9, line 61:
```suggestion
	sort.Ints(versions)
```

Is there a reason to slice this?  If so, add a comment explaining it.

e2etest/clean/clean_distributor_test.go
```diff
@@ -0,0 +1,74 @@
...
+	t.Logf("Versions %v", versions)
```
Aneurysm9, line 62:
```suggestion
```

This isn't necessary, is it?

e2etest/clean/clean_distributor_test.go
```diff
@@ -0,0 +1,74 @@
...
+	//get document versions
+	var versions []int
+	for i := range listDocumentVersionsOutput.DocumentVersions {
+		//can't delete the default version
+		if *listDocumentVersionsOutput.DocumentVersions[i].IsDefaultVersion {
+			continue
+		}
+		version, err := strconv.Atoi(*listDocumentVersionsOutput.DocumentVersions[i].DocumentVersion)
+		if err != nil {
+			t.Fatalf("Error getting document list %s", err.Error())
```
jefchien, line 56:
```suggestion
			t.Fatalf("Error getting document list %v", err)
```

e2etest/clean/clean_distributor_test.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		versions = append(versions, version)
+	}
+	//sort document versions from oldest to newest
+	sort.Ints(versions[:])
+	t.Logf("Versions %v", versions)
+	//delete all documents except the 5 latest
+	for i := 0; i < len(versions)-5; i++ {
+		version := strconv.Itoa(versions[i])
```
Aneurysm9, line 65:
```suggestion
	for _, v := range versions[:len(versions)-5] {
		version := strconv.Itoa(v)
```

Here slicing does make sense and also avoids any confusion about what the index variable is for.

e2etest/clean/clean_distributor_test.go
jefchien, line 54: Why are you converting from string to int and then back to string? Is it just for the sorting?

sethAmazon: Yes

e2etest/clean/clean_distributor_test.go
sethAmazon: I get a compilation error

Aneurysm9: What error?  It looks like it is possible for `versions` to be a nil slice, though I wouldn't expect a compile error for that.  In any event, it's easy to avoid.  Instead of declaring it like this:

```go
var versions []int
```

do this:

```go
versions := make([]int, 0, len(listDocumentVersionsOutput.DocumentVersions))
```

That makes a new slice of ints with length 0 and capacity to hold as many items as are in the list of document versions.

sethAmazon: Cannot use 'versions[]' (type int) as the type []int

sethAmazon: Same issue

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"log"
+	"sort"
+	"strconv"
+)
+
+func main() {
+	//set up aws go sdk config
+	awsConfig := aws.Config{}
+	testSession, err := session.NewSession(&awsConfig)
```
jefchien, line 30: If the `aws.Config` is empty, then you don't need it.

https://docs.aws.amazon.com/sdk-for-go/api/aws/session/
> Creating a Session without additional options will load credentials region, and profile loaded from the environment and shared config automatically. See, "Environment Variables" section for information on environment variables used by Session.

```
// Create Session
sess, err := session.NewSession()
```

e2etest/clean/clean_distributor_test.go
sethAmazon: versions := make([]int, 0, len(listDocumentVersionsOutput.DocumentVersions)) creates a slice if I'm not mistaken since len is not allowed in go array size like it is in java.

sethAmazon: It would make runtime faster but lets be honest not by much since max size is 25

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		log.Fatalf("Error creating session %v", err)
+	}
+	// set up aws go sdk ssm client
+	ssmClient := ssm.New(testSession, &awsConfig)
```
jefchien, line 35: Same for this client. https://docs.aws.amazon.com/sdk-for-go/api/service/ssm/#New

```
// Create a SSM client from just a session.
svc := ssm.New(mySession)
```

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,71 @@
...
+			continue
+		}
+		version, err := strconv.Atoi(*docVer.DocumentVersion)
+		if err != nil {
+			log.Fatalf("Error getting document list %v", err)
+		}
+		versions = append(versions, version)
+	}
+	//sort document versions from oldest to newest
+	//sort.Ints(versions[:]) gives compilation error "Cannot use 'versions[]' (type int) as the type []int"
```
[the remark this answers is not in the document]
sethAmazon: lol I ment to remove the : mb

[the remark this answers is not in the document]
jefchien: Shouldn't it just be `sort.Ints(versions)`?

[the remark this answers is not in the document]
sethAmazon: In the comment yes I have changed it.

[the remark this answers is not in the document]
jefchien: The comment got changed to `sort.Ints(versions[])` instead of `sort.Ints(versions)`.

e2etest/clean/clean_distributor_test.go
sethAmazon: ignore what I said earlier this proper way of doing this.

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,80 @@
...
+	ssmClient := ssm.New(testSession)
+
+	//get list of documents with versions
+	documentName := "testAWSDistroOTel-Collector"
+	listDocumentVersionInput := ssm.ListDocumentVersionsInput{Name: &documentName}
+	listDocumentVersionsOutput, err := ssmClient.ListDocumentVersions(&listDocumentVersionInput)
+	if err != nil {
+		log.Fatalf("Error getting document list %v", err)
+	}
+	versions := make([]*string, 0, len(listDocumentVersionsOutput.DocumentVersions))
```
jefchien, line 41: Maybe make a struct to hold both the string and int values so you don't have to convert back and forth.

```go
type documentTuple struct {
	str string
	num int
}
```
Then, you would just add all of them to the versions slice and only need to do two loops. One to get the largest `documentTuple` and append them to the slice and one to set the default/delete all the others.

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"log"
+	"strconv"
+)
+
+func main() {
+	//set up aws go sdk config
```
jefchien, line 26: nitpick: This comment seems like it's a holdover from a previous commit.

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,80 @@
...
+		deleteDocumentInput := ssm.DeleteDocumentInput{Name: &documentName, DocumentVersion: version}
+		_, err := ssmClient.DeleteDocument(&deleteDocumentInput)
+		if err != nil {
+			log.Fatalf("Error deleting document version %s %v", *version, err)
+		}
+		log.Printf("Document version deleted %s", *version)
+	}
+}
+
+func convertVersionToInteger(version *string) int  {
```
jefchien, line 74: Something like 
```go
func toVersionTuple(info *ssm.DocumentVersionInfo) *versionTuple  {
	versionNumber, err := strconv.Atoi(*info.DocumentVersion)
	if err != nil {
		log.Fatalf("Error parsing version number %v", err)
	}
	return &versionTuple{
		str: *info.DocumentVersion,
		num: versionNumber,
	}
}
```

alolita (approved):

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,89 @@
...
+	listDocumentVersionInput := ssm.ListDocumentVersionsInput{Name: &documentName}
+	listDocumentVersionsOutput, err := ssmClient.ListDocumentVersions(&listDocumentVersionInput)
+	if err != nil {
+		log.Fatalf("Error getting document list %v", err)
+	} else if len(listDocumentVersionsOutput.DocumentVersions) < 1{
+		log.Fatalf("At least one document version must be found")
+	}
+
+	versionTuples := make([]documentVersionTuple, 0, len(listDocumentVersionsOutput.DocumentVersions))
+	largestVersion := documentVersionTuple{num: -1}
```
jefchien, line 49: nitpick: This could be a pointer so you would just point it to the largest version instead of creating another one, but this works as well.

Something like
```go
var largestVersion *documentVersionTuple

if largestVersion == nil || versionTuple.num > largestVersion.num {
  largestVersion = &versionTuple
}
```

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	ssmClient := ssm.New(testSession)
+
+	//get list of documents with versions
+	documentName := "testAWSDistroOTel-Collector"
+	listDocumentVersionInput := ssm.ListDocumentVersionsInput{Name: &documentName}
+	listDocumentVersionsOutput, err := ssmClient.ListDocumentVersions(&listDocumentVersionInput)
+	if err != nil {
+		log.Fatalf("Error getting document list %v", err)
+	} else if len(listDocumentVersionsOutput.DocumentVersions) < 1{
```
jefchien, line 44:
```suggestion
	} else if len(listDocumentVersionsOutput.DocumentVersions) <= 1{
```
This doesn't need to fail, but it should exit early if there's only 1 or less, which will be the case often if this runs hourly. It shouldn't be a failure if we don't need to clean it up.

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,94 @@
...
+		log.Print("No document versions to delete")
+		return
+	}
+
+	versionTuples := make([]documentVersionTuple, 0, len(listDocumentVersionsOutput.DocumentVersions))
+	var largestVersion *documentVersionTuple
+	for _, documentVersion := range listDocumentVersionsOutput.DocumentVersions {
+		versionTuple := toVersionTuple(documentVersion)
+		if largestVersion == nil || versionTuple.num > largestVersion.num {
+			largestVersion = &documentVersionTuple{num: versionTuple.num, str: versionTuple.str}
```
jefchien, line 54: Can't you just point it to the existing object?
```suggestion
			largestVersion = &versionTuple
```

e2etest/clean/clean_distributor.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		versionTuples = append(versionTuples, versionTuple)
+	}
+
+	//this is here to remove compiler warning that largest version may be nil
+	if largestVersion == nil {
+		log.Fatalf("Largest version pointer is nill not sure how we go here")
```
jefchien, line 61:
```suggestion
		log.Fatalf("Largest version pointer is nil. Not sure how we got here")
```

jefchien (approved):
LGTM

Issue #56, opened 2020-10-21
Don't trigger workflow for readme changes


wyTrivail:
we don't need to trigger workflow everytime for any changes, will need to filter the "code" files in the trigger condition of ci workflow

PR #709, opened 2021-11-03
Added path-ignore to PR-build and CI workflows.

jefchien:
**Description:** Fixes https://github.com/aws-observability/aws-otel-collector/issues/56 by preventing the PR-build and CI workflows from triggering on document and workflow changes. If the workflows (PR-build and CI) are not changed and the code is not changed, then the workflows don't need to be run.


sethAmazon (approved):

anuraaga:
Generally this setting doesn't work since required checks won't be run on docs PRs and the PRs become unmergable.

I notice this repo doesn't have required checks but it probably should per best practice. I think there are other solutions that work on the step level that could work in conjunction with that.

.github/workflows/PR-build.yml
```diff
@@ -24,116 +24,172 @@ env:
   PACKAGING_ROOT: build/packages 
 
 jobs:
+  changes:
```
sethAmazon, line 27: Why was paths-ignore not enough?

.github/workflows/PR-build.yml
```diff
@@ -24,116 +24,172 @@ env:
...
+              - '.github/workflows/PR-build.yml'
+              - 'cmd/**'
+              - 'config/**'
+              - 'deployment-template/**'
+              - 'e2etest/**'
+              - 'pkg/**'
+              - 'tools/**'
+              - 'Makefile'
+              - 'go.mod'
+              - 'go.sum'
```
aateeqi, line 47: Nitpick but would it make sense to make these two `go.**`?

.github/workflows/PR-build.yml
jefchien: Per @anuraaga up above:
> Generally this setting doesn't work since required checks won't be run on docs PRs and the PRs become unmergable.
> 
> I notice this repo doesn't have required checks but it probably should per best practice. I think there are other solutions that work on the step level that could work in conjunction with that.

If it's done on a paths-ignore level, then none of the jobs will run. This way, all of the jobs run, but they exit quickly. See [PR Build #805](https://github.com/aws-observability/aws-otel-collector/actions/runs/1423228261)

.github/workflows/PR-build.yml
jefchien: Yeah doesn't hurt. I guess I technically don't even need to check go.mod since any go.mod change will result in a go.sum change as well.

khanhntd:
Would you **consider job-level-condition** instead of step-level-condition, find interesting blog post here that supports that: https://github.blog/changelog/2019-10-01-github-actions-new-workflow-syntax-features/? So we only need to change one lines instead of changing the if condition for each steps.

jefchien:
> Would you **consider job-level-condition** instead of step-level-condition, find interesting blog post here that supports that: https://github.blog/changelog/2019-10-01-github-actions-new-workflow-syntax-features/? So we only need to change one lines instead of changing the if condition for each steps.

I tried that earlier. With job level conditions, the jobs themselves are cancelled rather than run.

khanhntd:
> > Would you **consider job-level-condition** instead of step-level-condition, find interesting blog post here that supports that: https://github.blog/changelog/2019-10-01-github-actions-new-workflow-syntax-features/? So we only need to change one lines instead of changing the if condition for each steps.
> 
> I tried that earlier. With job level conditions, the jobs themselves are cancelled rather than run.



> > Would you **consider job-level-condition** instead of step-level-condition, find interesting blog post here that supports that: https://github.blog/changelog/2019-10-01-github-actions-new-workflow-syntax-features/? So we only need to change one lines instead of changing the if condition for each steps.
> 
> I tried that earlier. With job level conditions, the jobs themselves are cancelled rather than run.

If it is canceled, then new PR release won't be able to deploy. Reasonable to me.

.github/workflows/PR-build.yml
```diff
@@ -170,31 +200,35 @@ jobs:
           
   run-test-case:
```
anuraaga, line 201: Taking a step back, is it worth running these on every PR build? I think it's pretty conventional to have integration type of tests only running on main build since unit tests should have good coverage anyways. To reduce the overhead of adding if-statement complexity to so many steps, I think it's good to rethink this file from the ground up.

For PR builds, I think just the current `build` job is enough. And as long as go modules / tests are cached using actions cache there probably isn't a big reason to skip by path if it's the only job.

jefchien: We're only testing a small subset of the integration tests. Even if we're only doing the `build` job, I still think it's worth skipping it. There's no benefit to the build.

Issue #711, opened 2021-11-04
Add ec2 instance cleaner 


sethAmazon:
**Is your feature request related to a problem? Please describe.**
We do not have enough open subnets for our soaking tests because too many ec2 instances are open. 

**Describe the solution you'd like**
Terminate ec2 instances that are over 30 days old. 

**Describe alternatives you've considered**
Manually delete instances instead of wring a go method.

PR #712, opened 2021-11-04
Delete EC2 Instances That Are Running Over 30 Days For Name Integ-tes…

sethAmazon:
…t-Sample-App And Integ-test-aoc

**Description:** 
EC2 instances are staying alive too long. Thus reducing the amount of open subnets. This fails the soaking test

**Link to tracking Issue:** 
(#711 )

**Testing:** 
Only dry run since someone deleted the old instance manually and I did not want to delete any new instances.

### DIFF
--- a/.github/workflows/weekly-resources-clean.yml
+++ b/.github/workflows/weekly-resources-clean.yml
@@ -11,27 +11,27 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-name: Hourly Resources Cleaner
+name: Weekly Resources Cleaner
 
 on:
   schedule:
-    - cron: "0 */1 * * *"
+    - cron: "0 0 * * SAT" # Run every saturday at midnight
+  workflow_dispatch:
 
 jobs:
   clean-soaking-resources:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      
+
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-access-key-id: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
           aws-region: us-west-2
-          
-      - name: Set up terraform
-        uses: hashicorp/setup-terraform@v1
-      
-      - name: Clean resources which are created before yesterday
-        run: sh tools/workflow/clean-terraform-resources.sh
+
+      - name: Clean old ec2
+        run: go run tools/workflow/clean_ec2.go
+        env:
+          AWS_SDK_LOAD_CONFIG: true
--- a/tools/workflow/clean_ec2.go
+++ b/tools/workflow/clean_ec2.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"log"
+	"time"
+)
+
+func main() {
+	// set up aws go sdk ec2 client
+	testSession, err := session.NewSession()
+	if err != nil {
+		log.Fatalf("Error creating session %v", err)
+	}
+	ec2client := ec2.New(testSession)
+
+	// Get list of instance
+	//State filter
+	instanceStateFilter := ec2.Filter{Name: aws.String("instance-state-name"), Values: []*string{aws.String("running")}}
+
+	//Tag filter
+	instanceTagFilter := ec2.Filter{Name: aws.String("tag:Name"), Values: []*string{aws.String("Integ-test-Sample-App"), aws.String("Integ-test-aoc")}}
+
+	//get instances to delete
+	var deleteInstanceIds []*string
+	var nextToken *string
+	for {
+		describeInstancesInput := ec2.DescribeInstancesInput{Filters: []*ec2.Filter{&instanceStateFilter, &instanceTagFilter}, NextToken: nextToken}
+		describeInstancesOutput, err := ec2client.DescribeInstances(&describeInstancesInput)
+		if err != nil {
+			log.Fatalf("Failed to get instance for error %v", err)
+		}
+		for _, reservation := range describeInstancesOutput.Reservations {
+			for _, instance := range reservation.Instances {
+				//only delete instance if older than 30 days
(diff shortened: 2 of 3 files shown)


jefchien:
This shouldn't run hourly or even daily. Make a new workflow for a weekly clean.

sethAmazon:
@jefchien resolved

e2etest/clean/clean_ec2.go
```diff
@@ -0,0 +1,70 @@
...
+		nextToken = describeInstancesOutput.NextToken
+	}
+
+	//delete instances
+	var instanceIds []*string
+	for _, instance := range deleteInstances {
+		instanceIds = append(instanceIds, instance.InstanceId)
+		log.Printf("Delete instance %s tags %s launch-date %s", *instance.InstanceId, instance.Tags, instance.LaunchTime)
+	}
+	if len(instanceIds) < 1 {
```
SaxyPandaBear, line 61: Isn't this directly proportional to the length of `deleteInstances`? We can swap these around and put the if check before the for loop to short circuit earlier, though that's just a nitpick.

e2etest/clean/clean_ec2.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Print("No instances to delete")
+		return
+	}
+	terminateInstancesInput := ec2.TerminateInstancesInput{InstanceIds: instanceIds}
```
SaxyPandaBear, line 65: From the docs for [TerminateInstances](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_TerminateInstances.html), the list of instance IDs can hold up to 1000 instances in a batch. Should we be concerned about hitting that limit in this cleanup job? If we're only ever spinning up a handful of instances, I'm more inclined to leave this as is, but if this is bound to grow, then might as well implement the batching now so it doesn't bite us later.

sethAmazon: I don't think we will have over 1000 instances to delete since we will be cleaning every week and we only clean instances over 1 month old.

sethAmazon: We had only 350 open instances and this has not been cleaned for almost 6 months.

SaxyPandaBear (approved):
LGTM

.github/workflows/resources-clean.yml
```diff
@@ -36,7 +36,3 @@ jobs:
       - name: Clean resources which are created before yesterday
         run: sh tools/workflow/clean-terraform-resources.sh
 
-      - name: Clean distributor documents
```
sethAmazon, line 39: Added to the weekly.

e2etest/clean/ec2/clean_ec2.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(deleteInstances) < 1 {
+		log.Print("No instances to delete")
+		return
+	}
+
+	//delete instances
+	var instanceIds []*string
+	for _, instance := range deleteInstances {
+		instanceIds = append(instanceIds, instance.InstanceId)
+		log.Printf("Delete instance %s tags %s launch-date %s", *instance.InstanceId, instance.Tags, instance.LaunchTime)
```
jefchien, line 64: Is `instance.LaunchTime` a string? Should that use %v instead of %s? Same with `instance.Tags`

e2etest/clean/ec2/clean_ec2.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	if len(deleteInstances) < 1 {
+		log.Print("No instances to delete")
+		return
+	}
+
+	//delete instances
+	var instanceIds []*string
+	for _, instance := range deleteInstances {
```
jefchien, line 62: Couldn't you just append the instance IDs above in the first loop?

e2etest/clean/ec2/clean_ec2.go
```diff
@@ -0,0 +1,71 @@
...
+		instanceIds = append(instanceIds, instance.InstanceId)
+		log.Printf("Delete instance %s tags %s launch-date %s", *instance.InstanceId, instance.Tags, instance.LaunchTime)
+	}
+	terminateInstancesInput := ec2.TerminateInstancesInput{InstanceIds: instanceIds}
+	_, err = ec2client.TerminateInstances(&terminateInstancesInput)
+	if err != nil {
+		log.Fatalf("Failed to terminate instance %v", err)
```
jefchien, line 69: What would cause this to fail? Should this be fatal? Would it make sense to continue and try to terminate the other instances?

e2etest/clean/ec2/clean_ec2.go
```diff
@@ -0,0 +1,71 @@
+package main
```
jefchien, line 1: Consider putting this in tools/workflow with the other clean scripts.

e2etest/clean/ec2/clean_ec2.go
sethAmazon: It will convert to string but I change change to %v

e2etest/clean/ec2/clean_ec2.go
sethAmazon: See https://docs.aws.amazon.com/goto/WebAPI/ec2-2016-11-15/TerminateInstances
This can fail if an instance is protected. We should not protect these instances so this should be fine. Instances within the same az will not terminate but instances outside that az will. I think this is fine.

tools/workflow/clean_ec2.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	if len(deleteInstances) < 1 {
+		log.Print("No instances to delete")
+		return
+	}
+
+	//delete instances
+	for _, instance := range deleteInstances {
+		log.Printf("Try to delete instance %s tags %v launch-date %v", *instance.InstanceId, instance.Tags, instance.LaunchTime)
```
jefchien, line 58: Can't this just be printed when it's added to the slice in the first loop?

.github/workflows/weekly-resources-clean.yml
```diff
@@ -0,0 +1,37 @@
...
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+name: Weekly Resources Cleaner
+
+on:
+  schedule:
+    - cron: "0 0 * * 6" # Run every saturday at midnight
```
jefchien, line 18:
```suggestion
    - cron: "0 0 * * SAT" # Run every saturday at midnight
```

jefchien (approved):

Issue #716, opened 2021-11-05
Takes too long to build otel container in test cases


sethAmazon:
**Is your feature request related to a problem? Please describe.**
Takes too long to build otel container in test cases. This is because we do not cache the go mods when we build in the ec2 instance. 

**Describe the solution you'd like**
Build once and upload to ecr then take the docker image from there

**Describe alternatives you've considered**
Move go mod cache into the dockerfile
Move the building of otel outside of the docker file.

**Additional context**
https://github.com/aws-observability/aws-otel-collector/runs/4119657628?check_suite_focus=true

PR #719, opened 2021-11-08
Avoid building repetition docker image when doing PR Build Action

khanhntd:
To avoid repetition of building docker image amongst test case (avoid wasting resources and improve time), we need to ** add a step for PR Build Actions **: publishing docker image to ECR and let's the test framework pull the image. 
Ex. Fixing a bug - Describe the bug and how this fixes the issue.
Ex. Adding a feature - Explain what this achieves.>

https://github.com/aws-observability/aws-otel-collector/issues/716

**Testing:** 

### DIFF
--- a/.github/workflows/PR-build.yml
+++ b/.github/workflows/PR-build.yml
@@ -21,7 +21,14 @@ on:
 
 env:
   IMAGE_NAME: aws-otel-collector
-  PACKAGING_ROOT: build/packages 
+  PACKAGING_ROOT: build/packages
+  ECR_REPO: aws-observability/aws-otel-collector
+  IMAGE_LINK: "public.ecr.aws/aws-observability/aws-otel-collector"
+
+  TF_VAR_aws_access_key_id: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
+  TF_VAR_aws_secret_access_key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
+  TF_VAR_aoc_vpc_name: aoc-vpc-large
+  TF_VAR_aoc_vpc_security_group: aoc-vpc-security-group-large
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/PR-build.yml
```diff
@@ -134,9 +141,89 @@ jobs:
...
+    needs: build
+    steps:
+      - name: Cache Docker Image
+        id: cached_image
+        uses: actions/cache@v2
+        with:
+          key: "cached_image_${{ github.run_id }}"
+          path: "${{ env.PACKAGING_ROOT }}"
+
+      - name: Build Docker Image
```
anuraaga, line 156: Use the official action for building

https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md

It has docs on how to use cache natively too instead of a separate save step

anuraaga (commented):
Oh left a comment but then noticed it's PR build. Secrets aren't available to PR build so you can't login to AWS (or anything really) in them.

anuraaga:
I see a pattern of having many jobs which is probably causing problems. You should have as few jobs as possible so everything runs on the same runner. `packaging-msi` is of course special, but otherwise if it's all on one job you can just docker build once and reuse it.

khanhntd:
> I see a pattern of having many jobs which is probably causing problems. You should have as few jobs as possible so everything runs on the same runner. `packaging-msi` is of course special, but otherwise if it's all on one job you can just docker build (without push) once and reuse it.

Make sense to me when clearing jobs. Since I thought that a better explanation  and corresponding steps for each job  would be  good for our case.

anuraaga:
Yeah there could be some readability advantage of separate jobs but the technical limitations are too great, such as propagating artifacts between jobs. So I think in general, when there is no technical reason to do so, splitting is an antipattern for GitHub actions we need to get as much as we can out of a worker

khanhntd:
Reasonable to me. After reading what you said, seems the runner from my workflow  does not utilize each jobs. Also it's a pity that I couldn't use the native docker build cache inline. Another thing to ask is: I though the [secret key is set amongst the repository, env, organization based on its doc.](https://docs.github.com/en/actions/security-guides/encrypted-secrets). So don't know the reason why it is not usable for this workflow

anuraaga:
GitHub specifically does not mount secrets for `pull_request` workflows from forked repos - this is because it would otherwise be quite simple for some unmerged code in a PR to just steal the secrets. Most PR builds are from forked repos so one can't rely on secrets in any PR builds in practice.

https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

khanhntd:
Closing the issues since Seth works on another PQ and the result optimizes perfectly. https://github.com/aws-observability/aws-otel-collector/pull/721

PR #720, opened 2021-11-08
Separate ecs cloudwatch container insights into separate configs

genbit:
**Description:** v0.14.0 introduced two new configs for Amazon ECS that enables both application Otel metrics export to CloudWatch Metrics  and ECS container insights metrics to CloudWatch Container Insights metrics. This PR splits those configs into two parts, to allow customers enable both Container Insights at ECS cluster level and OpenTelemetry.


config/ecs/ecs-cloudwatch.yaml
```diff
@@ -19,47 +19,6 @@ processors:
       - system
       - ecs
       - ec2
-  filter:
```
hossain-rayhan, line 22: We need to remove line 11 as well. We are not using the `awsecscontainermetrics` anymore.

hossain-rayhan (approved):
One minor comment added. LGTM.

config/ecs/ecs-cloudwatch.yaml
genbit: addressed the change with force push.

alolita (approved):

Issue #677, opened 2021-10-21
Use multimod for versioning


anuraaga:
Upstream has migrated to using multimod for versioning, it would be good to do so here too.

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/4879

khanhntd:
Hey @anuraaga , could I tackle this problem?

PR #726, opened 2021-11-10
Go Multimod

khanhntd:
**Description:** 
Centralize all the module set into one files and use this version yaml to manage all the module set during the release.
**Note:** The scope of this tool is not for adopting different dependent sets with different versions.

**Link to tracking Issue:**
https://github.com/aws-observability/aws-otel-collector/issues/677

**Testing:** 
Using **Verify Module Versioning** from [OpenTelemetry-go-build-tools](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/multimod)
**Documentation:** 
**Know more about use cases and why should we use multimod:** https://aws.amazon.com/blogs/opensource/simplifying-opentelemetry-collector-and-go-library-releases-with-the-go-multimod-tool/
**Know more about using Go multimod:** https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/multimod


anuraaga (approved):

versions.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright The OpenTelemetry Authors
```
jefchien, line 1: This looks like the wrong license.

sethAmazon (approved):

versions.yaml
```diff
@@ -0,0 +1,26 @@
...
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+
+module-sets:
+  adot-base:
+    version: v0.14.0-RC1
```
Aneurysm9, line 17: Why `v0.14.0-RC1`?  `v0.14.0` has already shipped, right?

Aneurysm9 (approved):
This enables the multimod tool to be used, but doesn't seem to do anything with it.  I'd consider adding `multimod verify` to the PR/CI workflows [as seen in collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/build-and-test.yml#L85-L86) and maybe `Makefile` targets that can simplify syncing with an upstream repo and preparing releases, [as seen in go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/Makefile#L209-L225).  See [the `RELEASING` docs](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/RELEASING.md) for how those are used.

versions.yaml
khanhntd: Interesting. I thought I would need to setup in the separate PR. Also another typo for me, the v.0.14.0 was shipped to cx.

.github/workflows/CI.yml
```diff
@@ -125,6 +125,27 @@ jobs:
...
+      id: cached_tools
+      uses: actions/cache@v2
+      with:
+        key: v1-tools-${{ runner.os }}-${{ hashFiles('./tools/linters/go.sum') }}
+        path: ~/go/bin
+
+    #Run multimod - keep track of any modules are added to repo without added to multimod (centralized module set)
+    #Documents: https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/multimod/README.md
+    - name: Run multimod
+      if: steps.cached_tools.outputs.cache-hit != 'true'
```
Aneurysm9, line 138:
```suggestion
```

I don't think this condition is required or desired.

.github/workflows/CI.yml
```diff
@@ -125,6 +125,27 @@ jobs:
...
+      run: |
+        make multimod-verify
+
+
+    # Run linters - static analysis tools used to flag programming errors, bugs, stylistic errors and suspicious constructs
+    # Documents: https://en.wikipedia.org/wiki/Lint_(software)
+    - name: Run Linters
+      if: steps.cached_tools.outputs.cache-hit != 'true'
```
Aneurysm9, line 146:
```suggestion
```

Same here.

.github/workflows/PR-build.yml
```diff
@@ -73,6 +73,25 @@ jobs:
...
+      id: cached_tools
+      uses: actions/cache@v2
+      with:
+        key: v3-tools-${{ runner.os }}-${{ hashFiles('./tools/linters/go.sum') }}
+        path: ~/go/bin
+
+    #Run multimod - keep track of any modules are added to repo without added to multimod (centralized module set)
+    #Documents: https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/multimod/README.md
+    - name: Run multimod
+      if: steps.cached_tools.outputs.cache-hit != 'true'
```
Aneurysm9, line 86:
```suggestion
```

.github/workflows/PR-build.yml
```diff
@@ -73,6 +73,25 @@ jobs:
...
+      run: |
+        make multimod-verify
+
+    # Run linters - static analysis tools used to flag programming errors, bugs, stylistic errors and suspicious constructs
+    # Documents: https://en.wikipedia.org/wiki/Lint_(software)
+    - name: Run Linters
+      if: steps.cached_tools.outputs.cache-hit != 'true'
```
Aneurysm9, line 93:
```suggestion
```

Makefile
```diff
@@ -43,9 +43,18 @@ GOOS=$(shell go env GOOS)
...
 DOCKER_NAMESPACE=amazon
 COMPONENT=awscollector
+
 LINT=$(PWD)/bin/golangci-lint
 STATIC_CHECK=$(PWD)/bin/staticcheck
 
+TOOLS_MOD_DIR := ./tools/linters
+TOOLS_DIR := $(abspath ./.tools/linters)
+
+MULTIMOD = $(TOOLS_MOD_DIR)/multimod
```
Aneurysm9, line 53:
```suggestion
MULTIMOD = $(TOOLS_DIR)/multimod
```

This should reference the built binary, right?

Makefile
```diff
@@ -131,11 +140,36 @@ lint-static-check:
...
+multimod-tag: | $(MULTIMOD)
+    @[ "${MODSET}" ] || ( echo ">> env var MODSET is not set"; exit 1 )
+    $(MULTIMOD) verify && $(MULTIMOD) tag -m ${MODSET} -c ${COMMIT}
+
 .PHONY: install-tools
 install-tools:
-	cd tools/linters && GOBIN=$(PWD)/bin go install golang.org/x/tools/cmd/goimports
-	cd tools/linters && GOBIN=$(PWD)/bin go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.0
-	cd tools/linters && GOBIN=$(PWD)/bin go install honnef.co/go/tools/cmd/staticcheck@v0.2.0
+	cd $(TOOLS_MOD_DIR) && go install go.opentelemetry.io/build-tools/multimod
```
Aneurysm9, line 168:
```suggestion
	cd $(TOOLS_MOD_DIR) && GOBIN=$(PWD)/bin go install go.opentelemetry.io/build-tools/multimod
```

This should be installed in the same location as the other tools.  That said, I think if the other changes here are applied then this will be unused.

Makefile
```diff
@@ -131,11 +140,36 @@ lint-static-check:
...
+	cd $(TOOLS_MOD_DIR) && GOBIN=$(PWD)/bin go install golang.org/x/tools/cmd/goimports
+	cd $(TOOLS_MOD_DIR) && GOBIN=$(PWD)/bin go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.0
+	cd $(TOOLS_MOD_DIR) && GOBIN=$(PWD)/bin go install honnef.co/go/tools/cmd/staticcheck@v0.2.0
+	#cd $(TOOLS_MOD_DIR)/linters && GOBIN=$(PWD)/bin go install go.opentelemetry.io/build-tools/multimod
```
Aneurysm9, line 172:
```suggestion
```

Makefile
```diff
@@ -131,11 +140,36 @@ lint-static-check:
 lint: lint-static-check
 	$(LINT) run --timeout 5m
 
+.PHONY: multimod-verify
+multimod-verify: install-tools
+	@echo "Validating versions.yaml"
+	multimod verify
```
Aneurysm9, line 146:
```suggestion
multimod-verify: | $(MULTIMOD)
	@echo "Validating versions.yaml"
	$(MULTIMOD) verify
```

Consistency.

Makefile
```diff
@@ -131,11 +140,36 @@ lint-static-check:
...
+
+COREPATH ?= "../opentelemetry-go"
```
Aneurysm9, line 148:
```suggestion
COREPATH ?= "../opentelemetry-collector-contrib"
```

This is more likely to be the repository we care about syncing with here.

versions.yaml
```diff
@@ -0,0 +1,30 @@
...
+  #Goals: Centralize all the module set into one files and use this version yaml to manage all the module set during the release.
+  adot-base:
+    version: v0.14.0
+    modules:
+      - github.com/aws-observability/aws-otel-collector
+      - github.com/aws-observability/aws-otel-collector/pkg/lambdacomponents
+  adot-tools:
+    version: v0.14.0
+    modules:
+      - github.com/aws-observability/aws-otel-collector/tools/linters
```
Aneurysm9, line 26: Is this module expected to be versioned independently and to regularly change when the base modules do not?  I'm not sure I see a need to have a separate module set here.

Aneurysm9 (changes requested):

Makefile
khanhntd: Yeah you are right. I have deleted my changes because of the errors but forget to change some of the code back. Thanks for the notice.

versions.yaml
khanhntd: Same. At first I separate  the two since I intend to add another folder for the multi-mod but there is no reason now. Thanks for the notice.

Makefile
khanhntd: Yesterday, for some reasons, I ran this code locally but it turns out normal; however, when running in the github-env I was not able to run it because of not recognizing the multi-mod command. Therefore, I change this as a base to run in the github-env first.

Makefile
khanhntd: You are right. Another typo from me. Changed.

Makefile
khanhntd: It works now. As you suggested, should build the binaries first instead of install them asyn

Makefile
khanhntd: Yes. It should be reference to the built binary. Changed. Thanks for the call

Aneurysm9 (approved):

.github/workflows/CI.yml
```diff
@@ -116,7 +116,7 @@ jobs:
         path: |
           ~/go/pkg/mod
           ~/.cache/go-build
-        key: go-pkg-mod-${{ runner.os }}-${{ hashFiles('**/go.sum') }}
+        key: v1-go-pkg-mod-${{ runner.os }}-${{ hashFiles('**/go.sum') }}
```
jefchien, line 119: What was this change for?

Makefile
```diff
@@ -131,11 +148,35 @@ lint-static-check:
 lint: lint-static-check
 	$(LINT) run --timeout 5m
 
+.PHONY: multimod-verify
+multimod-verify: | $(MULTIMOD)
+	@echo "Validating versions.yaml"
+	$(MULTIMOD) verify
+
+COREPATH ?= "../opentelemetry-collector-contrib"
```
jefchien, line 156: I think we would need to have a sync core for both the collector core and contrib.

Makefile
```diff
@@ -131,11 +148,35 @@ lint-static-check:
...
+.PHONY: multimod-sync-core
+multimod-sync-core: | $(MULTIMOD)
+	@[ ! -d $COREPATH ] || ( echo ">> Path to core repository must be set in COREPATH and must exist"; exit 1 )
+	$(MULTIMOD) verify && $(MULTIMOD) sync -a -o ${COREPATH}
+
+
+.PHONY: multimod-prerelease
+multimod-prerelease: | $(MULTIMOD)
```
jefchien, line 164: We don't currently need this as we have no self dependencies. `aws-otel-collector` doesn't depend on `pkg/lambdacomponents` and vice versa.

Makefile
```diff
@@ -131,11 +148,35 @@ lint-static-check:
...
+.PHONY: multimod-verify
+multimod-verify: | $(MULTIMOD)
+	@echo "Validating versions.yaml"
+	$(MULTIMOD) verify
+
+COREPATH ?= "../opentelemetry-collector-contrib"
+.PHONY: multimod-sync-core
+multimod-sync-core: | $(MULTIMOD)
+	@[ ! -d $COREPATH ] || ( echo ">> Path to core repository must be set in COREPATH and must exist"; exit 1 )
+	$(MULTIMOD) verify && $(MULTIMOD) sync -a -o ${COREPATH}
```
jefchien, line 160: Do we need to call verify before each of these? If we do need to run it, could that be done with
```make
multimod-sync-core: multimod-verify
```

.github/workflows/CI.yml
khanhntd: Mostly naming convention. IMO it is easier for dev to test the CI without thinking another name.

Makefile
khanhntd: Yes. I agree with you at this point but mostly this one is a default value but prioritizing between collector core and contrib, I would say the contrib is more beneficial for the default value.

Makefile
khanhntd: Reasonable. Changed the pkg/lambdacomponents to exclude modules.

Makefile
khanhntd: Interesting thought. You are right. It would be more cleaner for me to do in this way.

Makefile
jefchien: Wait, no keep the `pkg/lambdacomponents` in the module set. We just don't need to do the prerelease which is for updating the current version dependencies of the modules in the module set to the proposed new version. Like if the `aws-otel-collector` depended on the `pkg/lambdacomponents`, it would update the version in the go.mod. But we don't have that kind of dependency. So we just don't need the prerelease step.

Makefile
jefchien: We can't do multiple?

.github/workflows/CI.yml
jefchien: I don't understand the convention. What is the v1 for?

Makefile
khanhntd: No we can't. Combine both is not an option since testing it out and return error. Creating an additional flag would not be an option too since it is wasteful for dev only used 1 flag. What I think was adding commenting in the Makefile and also additional notes in RELEASE doc if we used multimod in the future

.github/workflows/CI.yml
khanhntd: Its stand for version 1. One of the reason I used this is to easier delete or using another build cache, by only changing the v1 to v2 instead of thinking another name or delete some words.  Also [there is currently no delete cache ](https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions) .

Makefile
khanhntd: Instead of deleting the pre-release, would it be better to create another module set that includes ```pkg/lambdacomponents```? In this way, there will be no dependencies between the two and it will also preserve the ```prerelease``` functionality?

jefchien: From my understanding of `prerelease`, the only functionality we would get from it would be the creation of a branch named `prerelease-{module_set_name}-{new_version}` without any changes. I don't think that would be different if we had two module sets. Have you tried running it?

Makefile
jefchien: It looks like you can specify the path to the other repo with a flag `other-repo-root`.
https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/fdc91fd5380079ee89ea9315029a5b2304851b45/multimod/cmd/sync.go#L74
Couldn't you just have sync core run twice with both paths?

khanhntd: After rethinking it, make sense to me now. Add additional default core of ```opentelemetry-collector``` when ```sync-core```

Makefile
khanhntd: Tested it with changing the version to v0.15.0 and not only creating a new branch but also update all the ```go.mod``` in the module sets ```adot-base```. 
image

jefchien: What did it change in the go.mod?

jefchien: Right, so `prerelease` doesn't do anything for our use case.

khanhntd: Wait.I still not quite clear. Cause the **go.mod is up to date** so there are no changes. But if the ```go.mod``` for the base-dot is not up-to-date, then I meant pre-release is still useful right? Also ```pre-release``` is the only command that update all go.mod in the module-sets. So I don't see any reason why ```pre-release``` is not useful

Issue #729, opened 2021-11-10
Add docker image to public ecr on ci pass


sethAmazon:
**Is your feature request related to a problem? Please describe.**
A public container publish for all ci pass

**Describe the solution you'd like**
Add container to public ecr

**Describe alternatives you've considered**
None

alolita:
@sethAmazon is this done?

sethAmazon:
Yes closing now

PR #730, opened 2021-11-10
Publish CI Image To Public Repo (#729)

sethAmazon:
**Description:** 
Add public container publish to ecr

**Link to tracking Issue:** 
#729 

**Testing:** 
None, not sure how to test outside of a ci run


Aneurysm9 (approved):

.github/workflows/CI.yml
```diff
@@ -573,6 +574,46 @@ jobs:
...
+        if: steps.e2etest-public-ecr-release.outputs.cache-hit != 'true'
+        uses: actions/cache@v2
+        with:
+          path: build/packages
+          key: "cached_packages_${{ github.run_id }}"
+
+      - name: upload to public ECR
+        if: steps.e2etest-public-ecr-release.outputs.cache-hit != 'true'
+        run: |
+          aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-otel-test
```
anuraaga, line 610: Think it's better to use the official docker actions for login and build/push

https://github.com/docker/build-push-action

sethAmazon: aws-actions/amazon-ecr-login@v1 is for private ecr. 
docker/login-action@v1  is for dockerhub. 
We are not building the image just loading

sethAmazon: I added the github action for login but it seems there is no action to only push

.github/workflows/CI.yml
```diff
@@ -31,6 +31,7 @@ env:
   IMAGE_NAME: aws-otel-collector
   PACKAGING_ROOT: build/packages
   ECR_REPO: aws/aws-otel-collector
+  PUBLIC_ECR_REPO: aws-otel-test/dev
```
jefchien, line 34: Should this be a repo under https://gallery.ecr.aws/aws-observability/?

.github/workflows/CI.yml
```diff
@@ -573,6 +574,44 @@ jobs:
           #    docker tag $IMAGE_NAME $REPO_NAME:$TAG
           #    docker push $REPO_NAME:$TAG
 
+  e2etest-public-ecr-release:
```
jefchien, line 577: Why can't this just be done in the `upload to ECR` step above?

sethAmazon: Testing multiple docker logins within the same step may be a problem and I'm not sure how to tests this without affecting the current repo for prod. With them not in the same job now they can also run in parallel which will reduce ci time.

.github/workflows/CI.yml
sethAmazon: I think this is fine since it is a dev and not a prod release. People are only suppose to use this for testing.

jefchien:
It might be worth adding a section in the README with a link to the public ECR gallery with an explanation of what it's for.

.github/workflows/CI.yml
```diff
@@ -573,6 +574,44 @@ jobs:
...
+      - name: Cache if success
+        id:  e2etest-public-ecr-release
+        uses: actions/cache@v2
+        with:
+          path: |
+            VERSION
+          key: e2etest-public-ecr-release-${{ github.run_id }}
+
+      - name: Login to Public ECR
+        uses: docker/login-action@v1
```
khanhntd, line 593: IMO it should have **if the cache does not hit** in here also.

.github/workflows/CI.yml
```diff
@@ -573,6 +574,44 @@ jobs:
...
+          key: "cached_packages_${{ github.run_id }}"
+
+      - name: upload to public ECR
+        if: steps.e2etest-public-ecr-release.outputs.cache-hit != 'true'
+        run: |
+          docker load < build/packages/$IMAGE_NAME.tar
+          docker tag $IMAGE_NAME public.ecr.aws/$PUBLIC_ECR_REPO:latest
+          docker push public.ecr.aws/$PUBLIC_ECR_REPO:latest
+          docker tag $IMAGE_NAME public.ecr.aws/$PUBLIC_ECR_REPO:${{ needs.e2etest-preparation.outputs.version }}
+          docker push public.ecr.aws/$PUBLIC_ECR_REPO:${{ needs.e2etest-preparation.outputs.version }}
```
khanhntd, line 613: Would it possible for us to [multi-tag instead of separate the two?](https://stackoverflow.com/questions/25211198/docker-how-to-change-repository-name-or-rename-image/25214186#25214186)

anuraaga (approved):

.github/workflows/CI.yml
jefchien: Can push the two tags at once. https://docs.docker.com/engine/reference/commandline/push/#push-all-tags-of-an-image

.github/workflows/CI.yml
jefchien: Would rename the env variable to make it clear that this is for the dev repo
```suggestion
  DEV_ECR_REPO: aws-otel-test/dev
```

PR #732, opened 2021-11-10
Remove config and deployment-template from PR-build file filter.

jefchien:
**Description:** Since they don't affect either process, the PR-build shouldn't need to run.

### DIFF
--- a/.github/config/file-filters.yml
+++ b/.github/config/file-filters.yml
@@ -1,8 +1,6 @@
 changed:
   - '.github/workflows/PR-build.yml'
   - 'cmd/**'
-  - 'config/**'
-  - 'deployment-template/**'
   - 'e2etest/**'
   - 'pkg/**'
   - 'tools/**'


sethAmazon (approved):

.github/config/file-filters.yml
```diff
@@ -1,8 +1,6 @@
 changed:
   - '.github/workflows/PR-build.yml'
   - 'cmd/**'
-  - 'config/**'
```
khanhntd, line 4: IMO I should leave the config as the way it is, since someone [could change the config in the future develop](https://github.com/aws-observability/aws-otel-collector/blob/d2fcd75cc8654b08c04705de53955ee5be11f566/cmd/awscollector/Dockerfile#L69)

jefchien: It doesn't look like we build the image as part of the PR-build though.

jefchien: Just kidding. It builds the image as part of the test-cases.

PR #739, opened 2021-11-12
Enable the resource_to_telemetry_conversion to true in amp config.yaml files

vasireddy99:
This PR will add dimensions/tags and more detailed information to AMP metrics which helps to identify the metrics using cluster_name, task_id etc.,

Modified the following ecs-amp config files by enabling the resource_to_telemetry_conversion to true

Paths to config files 
config/ecs/ecs-amp.yaml
config/ecs/ecs-amp-xray.yaml
config/ecs/ecs-amp-xray-prometheus.yaml
config/ecs/ecs-amp-prometheus.yaml

### DIFF
--- a/config/ecs/ecs-amp-prometheus.yaml
+++ b/config/ecs/ecs-amp-prometheus.yaml
@@ -40,6 +40,8 @@ processors:
 exporters:
   awsprometheusremotewrite:
     endpoint: $AWS_PROMETHEUS_ENDPOINT
+    resource_to_telemetry_conversion:
+      enabled: true
 
 service:
   pipelines:
--- a/config/ecs/ecs-amp-xray-prometheus.yaml
+++ b/config/ecs/ecs-amp-xray-prometheus.yaml
@@ -50,6 +50,8 @@ exporters:
   awsxray:
   awsprometheusremotewrite:
     endpoint: $AWS_PROMETHEUS_ENDPOINT
+    resource_to_telemetry_conversion:
+      enabled: true
 
 service:
   pipelines:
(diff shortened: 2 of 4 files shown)


anuraaga (commented):
Oh it looks like this setting is available, cool.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusremotewriteexporter/config.go#L49

It's not documented at all 😱

anuraaga (commented):
Unfortunately, unlike the EMF exporter setting, this does not seem to support a list of labels

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/87ba2fa130a3b1353d717499e3b62c7bea4924a0/pkg/resourcetotelemetry/resource_to_telemetry.go#L31

We can't enable this because extremely high cardinality resource information will be sent and kill the prometheus

vasireddy99:
Please find the related PR and issues

Issue : https://github.com/open-telemetry/opentelemetry-collector/issues/2412 

PR : https://github.com/open-telemetry/opentelemetry-collector/pull/2899

alolita:
@anuraaga Agree with your concerns. Each service has to take a call on cardinality related performance concerns.

alolita (approved):
This is approved for AMP based on feedback from service team on no known limitations.

anuraaga:
I'm still worried about sending resource data without an allow list - assuming the cardinality can be handled without slowing down queries, it's still a lot of storage used (process.command_line can be on the kilobyte scale for some apps) and could cause cost balooning without value to customers.

alolita:
@anuraaga The service team has confirmed that they can handle high cardinality data.

alolita (approved):
lgtm based on service team feedback.

PR #750, opened 2021-11-17
Upgrade otel dependencies to v0.39.0.

jefchien:
**Description:** Main breaking change was that the `service/parserprovider` package was moved in upstream into `config/configmapprovider`.

**Testing:** Built collector and ran unit tests.

PR #751, opened 2021-11-17
Quick fix for multi-mod to sync-core optional path and core-path.

khanhntd:
**Description:** 
Fix changes to optional core-path and core path since finding bugs in syncing only 1 repo instead of 2

https://github.com/aws-observability/aws-otel-collector/issues/677


Makefile
```diff
@@ -154,11 +154,11 @@ multimod-verify: | $(MULTIMOD)
 	$(MULTIMOD) verify
 
 COREPATH ?= "../opentelemetry-collector-contrib"
-OPTIONAL_CORE_PATH ?="../opentelemetry-collector"
+SKIP_GO_MOD_TIDY ?= "true"
 .PHONY: multimod-sync-core
 multimod-sync-core: multimod-verify
-	@[ ! -d $COREPATH ] || ( echo ">> Path to core repository must be set in COREPATH and must exist"; exit 1 )
-	$(MULTIMOD) sync -a -o ${COREPATH} -o ${OPTIONAL_CORE_PATH}
```
khanhntd, line 161: Another thing why I delete the ```OPTIONAL_CORE_PATH``` is: after sync-core from the first repo, the dependencies from the upstream repo are updated. However, every time it runs a sync-core, it needs to be verified [if there are any changes before](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/fdc91fd5380079ee89ea9315029a5b2304851b45/multimod/internal/sync/sync.go#L46-L48). Therefore, between showing an error when running twice ```make multimod-sync-core``` and run sync-core once repo at a time, latter seems more feasible for me

go.mod
```diff
@@ -38,7 +38,7 @@ require (
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
-	go.opentelemetry.io/collector v0.38.0
+	go.opentelemetry.io/collector v0.39.0
```
jefchien, line 41: Don't bump up the version in this PR. It'll break things if it's only partially done. That's what https://github.com/aws-observability/aws-otel-collector/pull/750 is for.

Makefile
jefchien: If that's the case, then separate them out into separate targets: `multimod-sync-core` and `multimod-sync-contrib`.

Makefile
```diff
@@ -154,11 +154,11 @@ multimod-verify: | $(MULTIMOD)
...
+	@[ ! -d COREPATH ] || ( echo ">> Path to core repository must be set in COREPATH and must exist"; exit 1 )
+	$(MULTIMOD) sync -a -o ${COREPATH} --skip-go-mod-tidy ${SKIP_GO_MOD_TIDY}
```
jefchien, line 161: I don't think we should skip go mod tidy.

go.mod
khanhntd: Fixed it in the new commit but forgot to push it. Thanks for reminding me.

Makefile
khanhntd: Any reasons why we should separate the two? Since ```multimode-sync-core``` can be use as ```make sync-core COREPATH=../opentelemetry-collector```

Makefile
khanhntd: One of the reasons why I added the ```SKIP_GO_MOD_TIDY_PATH``` flag is: in case of the upstream repo changed the components, the ```go mod tidy``` does not allow us to update the dependencies. But mostly the ```sync-core``` main's function was to update dependencies, so that's why I want to keep that flow.
image

Makefile
khanhntd: Nice suggestion. Added and thanks for the call

Makefile
```diff
@@ -153,12 +153,19 @@ multimod-verify: | $(MULTIMOD)
...
+COREPATH ?= "../opentelemetry-collector"
+
 .PHONY: multimod-sync-core
 multimod-sync-core: multimod-verify
-	@[ ! -d $COREPATH ] || ( echo ">> Path to core repository must be set in COREPATH and must exist"; exit 1 )
-	$(MULTIMOD) sync -a -o ${COREPATH} -o ${OPTIONAL_CORE_PATH}
+	@[ ! -d COREPATH ] || ( echo ">> Path to core repository must be set in COREPATH and must exist"; exit 1 )
+	$(MULTIMOD) sync -a -o ${COREPATH}
+
+CORECONTRIBPATH ?= "../opentelemetry-collector-contrib"
```
jefchien, line 163: nitpick
```suggestion
CONTRIBPATH ?= "../opentelemetry-collector-contrib"
```

Another nitpick: Remove the space below between the target and the variable.

jefchien (approved):
Consider adding a `multimod-sync` that just calls them both.

Makefile
khanhntd: Fixed nitpick. For the purpose of verifying all the upstream-dependencies are up to date by using ```multimode-sync```? Interesting

Makefile
```diff
@@ -153,12 +153,19 @@ multimod-verify: | $(MULTIMOD)
...
-	@[ ! -d $COREPATH ] || ( echo ">> Path to core repository must be set in COREPATH and must exist"; exit 1 )
-	$(MULTIMOD) sync -a -o ${COREPATH} -o ${OPTIONAL_CORE_PATH}
+	@[ ! -d COREPATH ] || ( echo ">> Path to core repository must be set in COREPATH and must exist"; exit 1 )
+	$(MULTIMOD) sync -a -o ${COREPATH}
+
+CONTRIBPATH ?="../opentelemetry-collector-contrib"
+
+.PHONY: multimod-sync-contrib
+multimod-sync-contrib: multimod-verify
+	@[ ! -d CONTRIBPATH ] || ( echo ">> Path to core repository must be set in CONTRIBPATH and must exist"; exit 1 )
```
Aneurysm9, line 167:
```suggestion
	@[ ! -d CONTRIBPATH ] || ( echo ">> Path to contrib repository must be set in CONTRIBPATH and must exist"; exit 1 )
```

Aneurysm9 (approved):

PR #753, opened 2021-11-17
Added patch release workflow.

jefchien:
**Description:** Created a workflow to create a patch release. Added the dockerfile that will be used.


.github/workflows/patch-release.yml
```diff
@@ -0,0 +1,94 @@
...
+  IMAGE_NAMESPACE: amazon
+  ECR_REPO: aws-observability/aws-otel-collector
+
+jobs:
+  build-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build Image
+        run: docker build -t $IMAGE_NAME -f patch/${{ github.event.inputs.version }}/Dockerfile .
```
Aneurysm9, line 35: Does this already run from within `tools/release` or is this path wrong?

jefchien: It's the wrong path.

tools/release/patch/v0.14.1/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM amazon/aws-otel-collector:v0.14.0
+
+COPY config/ecs/ecs-*.yaml /etc/ecs/
```
sethAmazon, line 3: How to we plan on handling patches that are more than just a config change?

.github/workflows/patch-release.yml
```diff
@@ -0,0 +1,94 @@
...
+          aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
+          docker tag $IMAGE_NAME public.ecr.aws/$ECR_REPO:$TAG
+          docker tag $IMAGE_NAME public.ecr.aws/$ECR_REPO:latest
+          docker push public.ecr.aws/$ECR_REPO:$TAG
+          docker push public.ecr.aws/$ECR_REPO:latest
+
+      - name: Login Dockerhub
+        uses: docker/login-action@v1
+        with:
+          username: "mxiamxia"
```
sethAmazon, line 83: Can this be a secret?

jefchien: It could be. I don't think I'm able to add secrets. This is the way it is in our main branch's C/D workflow.

tools/release/patch/v0.14.1/Dockerfile
jefchien: We would likely need to fully rebuild the image and follow our normal release steps. This is a bit of a special case.

.github/workflows/patch-release.yml
Aneurysm9: I've added a `DOCKERHUB_RELEASE_USERNAME` secret with the relevant value.

Aneurysm9 (approved):

tools/release/patch/v0.14.1/Dockerfile
alolita: This is a special case. But we should have a patch building process which should include its own workflow.

alolita (approved):
Like the use of $TAG :-)

.github/workflows/patch-release.yml
```diff
@@ -0,0 +1,94 @@
...
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+
+name: Patch Release
```
anuraaga, line 15: I think it is a good idea to have a yaml file for this than doing steps manually so that's cool - but I would rename the file / naming of this to `release-v0.14.0.yml`, etc. to prevent confusion in the future. This isn't a general "patch release" workflow, as usually we would build a new binary. This is a special workflow only to update ECS configs for this specific release and we will want to delete it right away after - a normal patch release workflow is for building a binary with cherrypicks into a release branch, something like https://github.com/open-telemetry/opentelemetry-java/blob/main/.github/workflows/patch-release-build.yml

PR #755, opened 2021-11-18
Add prometheusremotewriteexporter in aws-otel-collector

Galactic21:
**Description:** 

Good morning, I'm creating this PR because I needed to adapt the code of aws-otel to support an exporter that exists in OpenTelemetry Collector. This exporter is prometheusremotewriteexporter. Although I have solved the problem using an own image, there are automatic processes that I do in which it is much easier for this to be in official aws-otel-collector. 

**Testing:** 

As I said earlier, I created an image of my own from this PR. This image has been tested to collect ECS metrics and send to a local Prometheus from this exporter. As you can see in the following image, there are ECS metrics that are being collected and displayed in my prometheus.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ This table represents the supported components of AWS OTel Collector in 2020. Th
 | zipkinreceiver                  | probabilisticsamplerprocessor | fileexporter                       |                        |
 | jaegerreceiver                  | spanprocessor                 | otlphttpexporter                   |                        |
 | `awscontainerinsightreceiver`   | filterprocessor               | prometheusexporter                 |                        |
+|                                 | filterprocessor               | prometheusremotewriteexporter      |                        |
 |                                 | metricstransformprocessor     | datadogexporter                    |                        |
 |                                 | resourcedetectionprocessor    | dynatraceexporter                  |                        |
 |                                 |                               | sapmexporter                       |                        |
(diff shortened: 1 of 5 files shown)


alolita:
Hi @Galactic21 thanks for the suggestion. The prometheusremotewriteexporter is indirectly included with the awsprometheusremotewriteexporter. Can you share more details about your use cases for this vanilla exporter? What is your monitoring service destination? Are you using both on-prem and on-cloud?

Galactic21:
Hi @alolita, I checked that prometheusremotewriteexporter is indirectly included in awsprometheusremotewriteexporter however as described in OpenTelemetry documentation, the AWS Prometheus Remote Write Exporter sends metrics in remote write format to Amazon Managed Service for Prometheus. My use case is to collect metrics from the task and containers where the aws-otel-collector container is and send it to a Prometheus that is in the cloud but is isolated, that is, it is hosted on its own server that exists on AWS. In other words, I will not be using Amazon Managed Service for Prometheus. In this test case, I'm using a local docker for this purpose of receiving the metrics. The entire process is done on-cloud.

Galactic21:
Okay, I assumed that awsprometheusremotewrite only works for Amazon Managed Service for Prometheus. I just tested it for my local Prometheus endpoint and it worked fine. Sorry, I really assumed that the exporter was just for one case. 

If you allow me other question, with awsecscontainermetricsreceiver I am able to collect the metrics of the task and containers where I implemented the aws-otel-collector container. But I would also like to know the metrics of the cluster where the task is located, how could I do that? Sorry again, I'm still new to this world of OpenTelemetry...

khanhntd:
Hi @Galactic21, thanks for asking the questions and it would be good for others to easy keep track of the issues instead of the PR. Therefore, I [create an issue for you](https://github.com/aws-observability/aws-otel-collector/issues/776) and our team is working on it to answer for you ASAP. For now, I will close this PR and we can continue from the issue if you have any questions.

PR #777, opened 2021-12-02
Change release branch naming expectations in workflows.

jefchien:
**Description:** Changed release branch naming to follow `release/v*`. Updated the concurrency group to allow the CI to run on both the main and release branches without interfering with each other.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -16,7 +16,7 @@ on:
   push:
     branches:
       - main
-      - release-branch-v*
+      - release/v*
       - dev
     paths-ignore:
       - '.github/**'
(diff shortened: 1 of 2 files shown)


aateeqi (approved):

.github/workflows/CI.yml
```diff
@@ -43,7 +43,7 @@ env:
   WIN_SIGNED_PKG_FOLDER: OTelCollectorAuthenticode/AuthenticodeSigner-SHA256-RSA
 
 concurrency:
-  group: ci
+  group: ci-${{ github.ref_name }}
```
khanhntd, line 46: Interesting with the use of ref_name

khanhntd (approved):

.github/workflows/CI.yml
jefchien: It's just so the different branches don't cancel each other when running the CI at the same time.

khanhntd: True. Also can be used with tags as well so interesting use cases.

jefchien: Yep, but as of now, we only run the CI on push to `main`, `dev`, or `release/v*`.

PR #783, opened 2021-12-04
Adding dispatch workflow for soaking test in version 14.

khanhntd:
Description:
Currently, the branch version 14.0 does not have any dispatch workflow. So, we cannot test the soaking test besides bumping version. To solve this issue, setting up a dispatch workflow in the soaking test.


jefchien (approved):
LGTM

Issue #793, opened 2021-12-07
Merging from PR is needed when testing with integration test


khanhntd:

sethAmazon:
Closing due to not needing this any more. Team decided running ci on dev/* was better.

PR #794, opened 2021-12-07
Add workflow dispatch for CI

khanhntd:
Currently to test the integration test, one needs to merge it from the PR. In order to easier testing with the CI, adding a workflow dispatch with the committed SHA would not only decrease un-necessary PR but it could also avoid impact to our customer who is using the main branch.

**Link to tracking Issue:** 
https://github.com/aws-observability/aws-otel-collector/issues/793

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -491,6 +491,10 @@ jobs:
           fi
           echo "::set-output name=commit_id::$commit_id"
 
+      - name: Checking branch for testing
+        run: |
+          echo ${{ steps.get_sha.outputs.commit_id }}
+
       - name: prepare CI&CD stack
         run: |
           cp .aoc-stack-test.yml build/packages/
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,7 +33,6 @@ on:
         required: true
 
 env:
-  CI_DEFAULT_BRANCH: 'main'
   IMAGE_NAME: aws-otel-collector
   PACKAGING_ROOT: build/packages
   ECR_REPO: aws/aws-otel-collector
@@ -483,7 +482,7 @@ jobs:
         id: get_sha
         run: |
           # Checking whether it is for integration testing or for the merge pull request from CI
-          commit_id="$CI_DEFAULT_BRANCH"
+          commit_id="$GITHUB_REF_NAME"
           if [ "${{ github.event.inputs.sha }}" != "" ]
           then
             commit_id = "${{ github.event.inputs.sha }}"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -565,11 +565,11 @@ jobs:
       ec2-matrix-3: ${{ steps.set-matrix.outputs.ec2-matrix-3 }}
     steps:
       - name: Checkout main
-        if: github.event.inputs.sha == ''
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -27,6 +27,12 @@ on:
   repository_dispatch:
     types: [dependency-build, workflow-run]
 
+  workflow_dispatch:
```
sethAmazon, line 30: I don't think this will work since you are sending in the sha but not using it

khanhntd: You are right. Adding the checkout condition to checkout the sha if it has the sha; otherwise, it is default.

.github/workflows/CI.yml
```diff
@@ -557,8 +564,15 @@ jobs:
       ec2-matrix-2: ${{ steps.set-matrix.outputs.ec2-matrix-2 }}
       ec2-matrix-3: ${{ steps.set-matrix.outputs.ec2-matrix-3 }}
     steps:
-      - name: Checkout
+      - name: Checkout main
+        if: ${{ github.event.inputs.sha == null }}
+        uses: actions/checkout@v2
+
+      - name: Checkout sha
+        if: ${{ github.event.inputs.sha != null }}
```
sethAmazon, line 572: Do you need to do this in every checkout step?

sethAmazon: Ex here https://github.com/aws-observability/aws-otel-collector/pull/794/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8L599

khanhntd: At first I thought the following EC2 test are not needed to checkout for the following ref. But after considering, I would use the set-output https://github.com/aws-observability/aws-otel-collector/pull/794/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R482-R492 to avoid repeating these unnecessary steps.

sethAmazon (approved):

.github/workflows/CI.yml
```diff
@@ -473,6 +479,17 @@ jobs:
...
+      - name: Setting branch for testing
+        id: get_sha
+        run: |
+          # Checking whether it is for integration testing or for the merge pull request from CI
+          commit_id="$CI_DEFAULT_BRANCH"
+          if [ "${{ github.event.inputs.sha }}" != "" ]
+          then
+            commit_id = "${{ github.event.inputs.sha }}"
+          fi
+          echo "::set-output name=commit_id::$commit_id"
```
sethAmazon, line 491: Will this work on release branches?

khanhntd: Yes. Changed the default commit id to use **GITHUB_REF_NAME** would work not only on release branch, but also the main and the dev branch as well.

alolita:
Why is this PR not merged yet @sethAmazon ?

sethAmazon (approved):

khanhntd:
Will close this since using branches with `dev/*` is more feasible without adding any changes to the code and destroy branches would be easier when dealing with scaling branches problem. However, we got one problem when creating [subbranch from main branch ](https://stackoverflow.com/questions/22630404/git-push-refs-heads-my-subbranch-exists-cannot-create/22630664#22630664). Therefore, we would need to create another PR to deal with this issue

Issue #695, opened 2021-10-28
Numerous integration tests are timing out


jefchien:
The failing tests aren't consistent and the tests will pass on a re-run.

Request timeout
[e2etest-ec2-1 (signalfx_exporter_metric_mock, debian9) ](https://github.com/aws-observability/aws-otel-collector/runs/4015433209?check_suite_focus=true#step:9:1671)
```
null_resource.setup_sample_app_and_mock_server[0] (remote-exec): ERROR: for tmp_mocked_server_1  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)

null_resource.setup_sample_app_and_mock_server[0] (remote-exec): ERROR: for tmp_sample_app_1  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)

null_resource.setup_sample_app_and_mock_server[0] (remote-exec): ERROR: for mocked_server  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)

null_resource.setup_sample_app_and_mock_server[0] (remote-exec): ERROR: for sample_app  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)
null_resource.setup_sample_app_and_mock_server[0] (remote-exec): ERROR: An HTTP request took too long to complete. Retry with --verbose to obtain debug information.
null_resource.setup_sample_app_and_mock_server[0] (remote-exec): If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 60).
╷
│ Error: remote-exec provisioner error
│ 
│   with null_resource.setup_sample_app_and_mock_server[0],
│   on main.tf line 364, in resource "null_resource" "setup_sample_app_and_mock_server":
│  364:   provisioner "remote-exec" {
│ 
│ error executing "/tmp/terraform_729692501.sh": Process exited with status 1
╵
```

Job timeout
[e2etest-ec2-3 (sapm_exporter_trace_mock, arm_redhat8)](https://github.com/aws-observability/aws-otel-collector/runs/4000963227?check_suite_focus=true)
> The job running on runner GitHub Actions 66 has exceeded the maximum execution time of 360 minutes.
```
null_resource.start_collector[0]: Still creating... [5h57m2s elapsed]
null_resource.start_collector[0]: Still creating... [5h57m12s elapsed]
[...]

PR #800, opened 2021-12-10
Clean ELB resource over 5 days that are not destroyed during the terraform process 

khanhntd:
Currently, the quota for elb is 50 quota/region and the integration test has thrown some errors relating to load balancer since the terraform destroy does not work properly sometimes. Therefore, creating a elb cleaning script would help us with this issue. 

https://github.com/aws-observability/aws-otel-collector/issues/695

Using cd tools/workflow && go run clean_aws_rsc.go


tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+	destroyLoadBalancerError := destroyLoadBalancerResource()
+
+	if (destroyLoadBalancerError != nil){
+		log.Fatalf("Exit destroy Load Balancer Resources")
+	}
+
+	log.Printf("Finish destroy AWS resources")
+}
+
+func terminateEc2Instances()(error){
```
sethAmazon, line 50: Too bad it did not treat this file as a rename. So that it would keep me as the git blame. You will now be supporting this code @khanhntd :)

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+
+
+	describeLoadBalancerOutputs, describeLoadBalancerErr := svc.DescribeLoadBalancers(describeLoadBalancerInputs)
+
+	if describeLoadBalancerErr != nil {
+		log.Printf("Failed to get metadata for load balancer because of %v", describeLoadBalancerErr)
+		return describeLoadBalancerErr
+	}
+
+	for _, lb := range describeLoadBalancerOutputs.LoadBalancers {
```
sethAmazon, line 132: Will we need to take into account pagination? Probs not, since this will run on a timer and eventually we will end up with less than the number give for describe.

.github/workflows/weekly-resources-clean.yml
```diff
@@ -33,5 +33,10 @@ jobs:
 
       - name: Clean old ec2
         run: go run tools/workflow/clean_ec2.go
```
sethAmazon, line 35: Does this line need to be deleted?

khanhntd: Forgot to push it haha. Thanks for reminding me.

tools/workflow/clean_aws_rsc.go
khanhntd: I will use the coauthor one to share the ownership @sethAmazon :).

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+		log.Fatalf("Exit destroy Load Balancer Resources")
+	}
+
+	log.Printf("Finish destroy AWS resources")
+}
+
+func terminateEc2Instances()(error){
+	// set up aws go sdk ec2 client
+	testSession := session.Must(
+		session.NewSessionWithOptions(
```
jefchien, line 53: Is all of this necessary? Is the environment variable setup insufficient?
> Creating a Session without additional options will load credentials region, and profile loaded from the environment and shared config automatically.

https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+
+		//Skipping lb that does not contain aoc-lb string (relating to aws-otel-test-framework)
+		if filterLbNameResult := strings.Contains(*lb.LoadBalancerName, containLbString); !filterLbNameResult {
+			log.Printf("Skipping lb %s because of not containing %s", *lb.LoadBalancerName, containLbString)
+			continue
+		}
+
+		//Skipping lb that does not older than 5 days
+		if !time.Now().UTC().Add(-1 * time.Hour * 24 * pastDayDelete).After(*lb.CreatedTime) {
+			log.Printf("Skipping lb %s with launch-date %v", *lb.LoadBalancerName, lb.CreatedTime)
```
jefchien, line 142: This and the other skipped logging seem like they'll overwhelm the logging. I don't think we need to log every single load balancer we have. Just the ones we delete.

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+
+	//Tag filter
+	instanceTagFilter := ec2.Filter{Name: aws.String("tag:Name"), Values: []*string{aws.String("Integ-test-Sample-App"), aws.String("Integ-test-aoc")}}
+
+	//get instances to delete
+	var deleteInstanceIds []*string
+	var nextToken *string
+	for {
+		describeInstancesInput := ec2.DescribeInstancesInput{Filters: []*ec2.Filter{&instanceStateFilter, &instanceTagFilter}, NextToken: nextToken}
+		describeInstancesOutput, describeInstanceErr := ec2client.DescribeInstances(&describeInstancesInput)
```
jefchien, line 72: nitpick: Using and reassigning the same `err` variable is fine. Didn't need to create different variables for each.

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+		if describeInstanceErr != nil {
+			log.Printf("Failed to get instance for error %v", describeInstanceErr)
+			return describeInstanceErr
+		}
+		for _, reservation := range describeInstancesOutput.Reservations {
+			for _, instance := range reservation.Instances {
+				//only delete instance if older than 5 days
+				if time.Now().UTC().Add(-1 * time.Hour * 24 * pastDayDelete).After(*instance.LaunchTime) {
```
jefchien, line 80: Can change the const to be the entire calculation instead of just the day or can have two consts: one with the day and another that performs this calculation. It's used here and in the load balancer deletion, so it might as well be reused.

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+			return deleteLoadBalancerError
+		}
+
+		log.Printf("Delete lb %s successfully", *lb.LoadBalancerName)
+
+
+	}
+
+	return nil
+
```
jefchien, line 166: nitpick: Remove the extra spacing everywhere and run the formatter. It's fine to have some, but I try to avoid leading and trailing spacing.

khanhntd: Reasonable. Deleting some extra space.

tools/workflow/clean_aws_rsc.go
khanhntd: Its weird for me that went I ran with the same ```err``` variable, it shows me the error of not able to assigning multiple values to err var. That's why I changed to this to be able to run locally

jefchien: You should be able to assign (=) multiple times. You just won't be able to declare (:=) more than once unless another variable is being declared at the same time.

```go
a, err := someFunc()
_, err = otherFunc() // := would be invalid
b, err := anotherFunc() // b is being declared
```

tools/workflow/clean_aws_rsc.go
khanhntd: Yes. You are right. The elbv2 only supports application and network lb, excludes the classic one. That's why it's hard for me to imagine these elbv2 can exceed 50 which is the default for paginate. However, adding next Marker to prevent that in the long run.

tools/workflow/clean_aws_rsc.go
khanhntd: Reasonable. Updating with the latter solution (using two const).

tools/workflow/clean_aws_rsc.go
khanhntd: Make sense. Another thing to learn in go. Fixed it

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+	if (destroyLoadBalancerError != nil){
+		log.Fatalf("Exit destroy Load Balancer Resources")
+	}
+
+	log.Printf("Finish destroy AWS resources")
+}
+
+func terminateEc2Instances()(error){
+	// set up aws go sdk ec2 client
+	testSession := session.Must(
```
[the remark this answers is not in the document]
khanhntd: Hmm what do you mean? I don't understand the question much.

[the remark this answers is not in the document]
jefchien: Did you run `go fmt` on the file? Did the formatter make it multi-lined?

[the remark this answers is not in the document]
khanhntd: I haven't run it until now. I thought multiline would be easier to read so changed it manually.

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func terminateEc2Instances() error {
+	// set up aws go sdk ec2 client
+	testSession, err := session.NewSession(&aws.Config{
+		Region: aws.String(rsc_region)},
+	)
+
+	if err != nil {
+		log.Fatalf("Error creating session %v", err)
```
jefchien, line 60: I think you should pick one: either return the errors or exit immediately with log.Fatalf. Don't mix the two within the function.

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,176 @@
...
+		log.Fatalf("Exit destroy Load Balancer Resources")
+	}
+
+	log.Printf("Finish destroy AWS resources")
+}
+
+func terminateEc2Instances() error {
+	// set up aws go sdk ec2 client
+	testSession, err := session.NewSession(&aws.Config{
+		Region: aws.String(rsc_region)},
```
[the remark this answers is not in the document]
khanhntd: One of the reason is that I was not able to run the script without setting the region to us-west-2 even thought I have set it up with ```aws configure```. So I think this must be needed for anyone to run this script locally.
image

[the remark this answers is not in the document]
jefchien: That's interesting. It should be able to pick up the region being used. @sethAmazon's original ec2 clean script didn't need it.

tools/workflow/clean_aws_rsc.go
khanhntd: True. I fixed it in the latest commit but waiting for the PR build to success to use the cache.

tools/workflow/clean_aws_rsc.go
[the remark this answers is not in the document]
khanhntd: True. It seems the go sdk does not recognize the ```AWS_DEFAULT_REGION``` variable (https://stackoverflow.com/a/64802402) . So would need to use ```export AWS_REGION="us-west-2"``` to setup it.

.github/workflows/weekly-resources-clean.yml
```diff
@@ -31,7 +31,8 @@ jobs:
           aws-secret-access-key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
           aws-region: us-west-2
 
-      - name: Clean old ec2
-        run: go run tools/workflow/clean_ec2.go
+      - name: Clean old aws resources
+        run: go run tools/workflow/clean_aws_rsc.go
```
sethAmazon, line 35: Leave it as it you don't agree but I would prefer not to use rsc and just spell it out

jefchien: I agree, but I'm nitpicky. `rsc` looks like an acronym.

tools/workflow/clean_aws_rsc.go
```diff
@@ -0,0 +1,167 @@
...
+	containLbString = "aoc-lb"
+	pastDayDelete = 5
+)
+
+func main() {
+
+	log.Printf("Beging terminating EC2 Instances")
+	terminateEc2InstancesError := terminateEc2Instances()
+	if (terminateEc2InstancesError != nil){
+		log.Fatalf("Exit terminating EC2 Instances")
```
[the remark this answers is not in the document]
sethAmazon: This code was written before this pr. If a change is need, it should be in another pr not this one.

tools/workflow/clean_aws_rsc.go
sethAmazon: No need :) I don't want to be paged :)

sethAmazon (approved):

.github/workflows/weekly-resources-clean.yml
khanhntd: Agreed. Not using acronym anymore lol.

tools/workflow/clean_aws_rsc.go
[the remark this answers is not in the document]
jefchien: The point of returning an error instead of just panicking in the function is to be able to give the user the option to handle the error themselves. If you're going to log the error and then return it, why not just use `log.Fatalf` right away?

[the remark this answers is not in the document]
khanhntd: In terms of user's perspective, reasonable for me. Fixed it in the last commit @sethAmazon  @jefchien

tools/workflow/clean_aws_resources.go
```diff
@@ -0,0 +1,162 @@
...
+		return
+	}
+
+	terminateInstancesInput := ec2.TerminateInstancesInput{InstanceIds: deleteInstanceIds}
+	_, err = ec2client.TerminateInstances(&terminateInstancesInput)
+	if err != nil {
+		log.Fatalf("Failed to terminate instances %v because of %v", deleteInstanceIds, err)
+	}
+
+	return
```
jefchien, line 99: Don't need this return statement here.

tools/workflow/clean_aws_resources.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+			log.Printf("Delete lb %s successfully", *lb.LoadBalancerName)
+		}
+
+		if describeLoadBalancerOutputs.NextMarker == nil {
+			break
+		}
+		nextMarker = describeLoadBalancerOutputs.NextMarker
+	}
+	return
```
jefchien, line 161: Don't need this return statement.

tools/workflow/clean_aws_resources.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+const (
+	containLbName            = "aoc-lb"
+	pastDayDelete            = 5
+	pastDayDeleteCalculation = -1 * time.Hour * 24 * pastDayDelete //Currently, deleting resources over 5 days
+)
+
+func main() {
+
```
jefchien, line 35: Remove empty leading line.

jefchien (approved):

PR #810, opened 2021-12-15
Fix CD for 0.15.1

aateeqi:
Adding input to update docker latest image

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -13,7 +13,7 @@
 
 name: C/D
 
-on:        
+on:
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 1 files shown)


khanhntd (approved):
LGTM since the adot collector is not involving with the github workflow and to avoid having error https://github.com/aws-observability/aws-otel-collector/runs/4537449051?check_suite_focus=true

.github/workflows/CD.yml
```diff
@@ -13,7 +13,7 @@
 
 name: C/D
 
-on:
```
sethAmazon, line 16: Anyway you can leave out the formatting changes?

.github/workflows/CD.yml
```diff
@@ -323,38 +331,42 @@ jobs:
...
+          fi
+          if [ ${{ github.event.inputs.mirror-operator }} == 'true' ]; then
+            cd adot/tools/release/adot-operator-images-mirror && go run ./
+          fi
 
       - name: Login Dockerhub
         uses: docker/login-action@v1
         with:
-          username: "mxiamxia"
+          username: "${{ secrets.DOCKERHUB_RELEASE_USERNAME }}"
```
sethAmazon, line 369: nice catch

sethAmazon (approved):

.github/workflows/CD.yml
aateeqi: so i've copied this from khanhs release and i think all its doing is removing white spaces which shouldnt be an issue

PR #813, opened 2021-12-16
Using main branch for tools/workflow in CD

khanhntd:

Instead of using the committed cha branch, it would be better for us to use the tools/workflow from the main branch since it is up to date with the github workflow.


.github/workflows/CD.yml
```diff
@@ -74,7 +76,7 @@ jobs:
       - name: Cache packages
         uses: actions/cache@v2
         with:
-          key: "${{ env.PACKAGE_CACHE_KEY }}"
+          key: "${{ env.TESTING_FRAMEWORK_REPO }}"
```
jefchien, line 79: Why was this changed? The cached build is meant to be the release candidate.

.github/workflows/CD.yml
```diff
@@ -315,14 +317,24 @@ jobs:
...
           path: adot
+
+      #Since the tools from workflow are always uptodate and to keep it up with the github workflow, we need to check out the main branch
+      #And use the tools from there instead of the commited sha branch to avoid commit additional code for the commited sha's branch
+      #Documentation on side by side checkout: https://github.com/actions/checkout#checkout-multiple-repos-side-by-side
+
+      - name: Checkout main branch to get workflow tools
+        uses: actions/checkout@v2
+        with:
+          ref:  "${{ env.MAIN_BRANCH }}"
```
jefchien, line 336: I don't think you need to make this an environment variable.

.github/workflows/CD.yml
```diff
@@ -315,14 +317,24 @@ jobs:
...
+          path: main
```
jefchien, line 337: This also doesn't need to be in a separate path, since the input SHA checkout is already under the `adot` path, but it's fine.

.github/workflows/CD.yml
```diff
@@ -340,7 +352,7 @@ jobs:
         id: version
         run: |
           TAG=`cat build/packages/VERSION`
-          TARGET_VERSION=$TAG bash adot/tools/workflow/docker-version-compare.sh
+          TARGET_VERSION=$TAG bash main/tools/workflow/docker-version-compare.sh
```
jefchien, line 355: If you're going to change this to use the main branch instead of the input SHA, you should also change the path for the other tools being used.

.github/workflows/CD.yml
khanhntd: True. Intend to do the nested checkout but it would clear this way.

.github/workflows/CD.yml
khanhntd: Reasonable. Fixed it.

.github/workflows/CD.yml
khanhntd: The only used tool with the path ```adot``` is the releasing of dot operator. Other steps are using cache with build/packages. So I think it should be fine.

.github/workflows/CD.yml
```diff
@@ -124,7 +125,7 @@ jobs:
       - name: Restore cached packages
         uses: actions/cache@v2
         with:
-          key: "${{ env.PACKAGE_CACHE_KEY }}"
+          key: "${{ env.TESTING_FRAMEWORK_REPO }}"
```
jefchien, line 128: Again. Why is this the testing framework repo? Why are you changing the cache keys?

.github/workflows/CD.yml
jefchien: That is not true. On line 370, the image mirror is using the `adot` path.

.github/workflows/CD.yml
```diff
@@ -315,14 +316,24 @@ jobs:
...
```
jefchien, line 319: This one as well.

.github/workflows/CD.yml
khanhntd: Another fault from me. Changed it back to the original. Should check the repo commit before going somewhere else.

.github/workflows/CD.yml
khanhntd: This one is my fault as well. Changed it back to original.

.github/workflows/CD.yml
khanhntd: Yeah you are right. I misunderstood the use of the image mirror. Fixed it.

khanhntd: Also if that's the case, then we don't need to checkout for the input SHA anymore. Nice catch.

.github/workflows/CD.yml
```diff
@@ -33,6 +33,7 @@ env:
   TF_VAR_aws_secret_access_key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
   TF_VAR_aoc_vpc_name: aoc-vpc-large
   TF_VAR_aoc_vpc_security_group: aoc-vpc-security-group-large
+  TESTING_FRAMEWORK_REPO: 'aws-observability/aws-otel-collector-test-framework'
```
sethAmazon, line 36: Good clean up

sethAmazon (approved):

.github/workflows/CD.yml
```diff
@@ -318,11 +319,12 @@ jobs:
...
 
-      - name: Checkout the sha
+      #Since the tools from workflow are always up-to-date and to keep it up with the github workflow, we need to check out the main branch
+      #And use the tools from there instead of the commited sha branch to avoid commit additional code for the commited sha's branch
+      - name: Checkout main branch to get workflow tools
         uses: actions/checkout@v2
         with:
-          ref: ${{ github.event.inputs.sha }}
-          path: adot
+          ref: "main"
```
jefchien, line 327: This is not required. If we trigger the workflow dispatch, it'll use whatever branch is selected, which should be `main`, but can be another branch.

khanhntd: Reasonable to me after thinking. Fixed it.

jefchien (approved):
LGTM

PR #815, opened 2021-12-16
Support roll back for CD when validation test are failing

khanhntd:

Currently, the CD has the fast-forward behavior, meaning it would keep the image on the public repo **aws-observability/aws-otel-collector** even thought the CD validation test fails. If the test fails, we will retry until it passed and while we keep retrying, customers will be able to retrieve the not-stable version image. Therefore, it would be best for us to delete the version-image if it fails and when it success, we will keep it there and mark it as latest also.

Testing each command and script separately and also locally. To be more specific:
- **Delete binaries uploading to S3:** Create a public s3 and test it with the command: cd tools/release && bash delete-s3-release.sh (using my s3)
- **Delete ECR Image Tag:** Create a public repo and test it with the command: ``` aws ecr-public batch-delete-image --repository-name adot-collector-test --image-ids imageTag=latest --region us-east-1``` (using my public repo)
- **Delete Docker Image Tag:** Create a private repo in docker registry hub and test it with the command cd tools/release && bash delete-docker-image-tag.sh (using my s3)

- **Delete Object from S3:** https://docs.aws.amazon.com/cli/latest/reference/s3/rm.html
- **Delete ECR Image Tag:** https://docs.aws.amazon.com/pt_br/cli/latest/reference/ecr-public/batch-delete-image.html
- **Delete Docker Image Tag:** https://stackoverflow.com/a/59334315


tools/release/image-binary-release/delete-docker-image-tag.sh
```diff
@@ -0,0 +1,45 @@
+#! /bin/bash
```
sethAmazon, line 1: Can the sh files be written in go?

sethAmazon (approved):

khanhntd:
Closed PR due to some commits conflict with the main branch.

Issue #811, opened 2021-12-16
Create github page for benchmarking


jefchien:
Add a gh-pages branch and move the performance report there.

jefchien:
https://aws-observability.github.io/aws-otel-collector/

PR #818, opened 2021-12-20
Trigger performance test after successful CI run.

jefchien:
**Description:** Removes the scheduled performance tests. Also, auto-commits results to gh-pages and updates benchmarking.

**Link to tracking Issue:** https://github.com/aws-observability/aws-otel-collector/issues/811

**Testing:** Tested it with my fork.

**Documentation:**


.github/workflows/perf.yml
```diff
@@ -147,28 +148,26 @@ jobs:
...
-          branch: "perf-test-${{ github.run_id }}"
-          file_pattern: docs/performance_model.md
-      
-      - name: pull-request
-        uses: repo-sync/pull-request@v2
+      - name: Produce performance report
+        run: python e2etest/get-performance-model-table.py ${{ needs.get-testing-version.outputs.testing_version }}
+
+      - name: Store benchmark result
+        uses: jefchien/github-action-benchmark@v1.11.12-alpha2
```
khanhntd, line 155: Anyway for us to know the parameters that are required in your private GitHub action thought?

.github/workflows/perf.yml
```diff
@@ -147,28 +148,26 @@ jobs:
...
-          pr_title: "Update Performance Model"
-          pr_body: "Generated by performance test workflow [#${{ github.run_number }}](https://github.com/aws-observability/aws-otel-collector/actions/runs/${{ github.run_id }}) using https://github.com/aws-observability/aws-otel-collector/commit/${{ needs.get-testing-version.outputs.commit_id }}."
+          tool: "customSmallerIsBetter"
+          output-file-path: performance-data.json
+          benchmark-data-dir-path: benchmark/trend
+          max-items-in-chart: ${{ env.MAX_BENCHMARKS_TO_KEEP }}
+          gh-pages-branch: ${{ env.GH_PAGES_BRANCH }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          commit-sha: ${{ github.event.inputs.sha }}
+          auto-push: false
```
khanhntd, line 164: Can I ask what is this flag for?

e2etest/get-performance-model-table.py
```diff
@@ -58,9 +58,38 @@ def flatten_performance_models(models):
         x["data_mode"], x["data_rate"]))
     return models_list
 
+def get_benchmark_data():
+    """
+    Splits models by test
+    """
+    benchmark_data = []
```
khanhntd, line 65: IMO, instead of using only [], using list(dict) would have more clarification in defining what is added into the list.

.github/workflows/perf.yml
```diff
@@ -147,28 +148,26 @@ jobs:
...
-        uses: stefanzweifel/git-auto-commit-action@v4
-        with:
-          commit_message: "Update benchmarking"
-          branch: "perf-test-${{ github.run_id }}"
-          file_pattern: docs/performance_model.md
-      
-      - name: pull-request
-        uses: repo-sync/pull-request@v2
+      - name: Produce performance report
+        run: python e2etest/get-performance-model-table.py ${{ needs.get-testing-version.outputs.testing_version }}
```
khanhntd, line 152: IMO, we should use flags instead of passing environment variables since we could have more flags for future use and it would make more clarification when catching these flags in the python code.

.github/workflows/perf.yml
jefchien: It's a fork of the existing https://github.com/benchmark-action/github-action-benchmark. The change was needed to pass the workflow dispatch input SHA into the action so the benchmarks are tagged with the correct commit. I have a PR https://github.com/benchmark-action/github-action-benchmark/pull/93 open for the change, but until that gets merged or another method of inserting the commit is found, we'll need to use the fork.

.github/workflows/perf.yml
jefchien: The `auto-push` flag, when set to true, will push the changes straight to the remote branch (https://github.com/benchmark-action/github-action-benchmark#auto-push-optional). Since we want to bundle the benchmark update with other changes, we need to disable the auto-push. It's false by default, so maybe it doesn't need to be explicitly set.

.github/workflows/perf.yml
khanhntd: Got cha. Make more sense to me now. Currently reviewing yours other PR as well.

.github/workflows/perf.yml
khanhntd: Reasonable. I am fine with explicitly set this variable as false since I would know more about this command and its has more clarification with the flags though.

.github/workflows/perf.yml
```diff
@@ -13,12 +13,9 @@
 
 name: 'Performance test'
 on:
-  schedule:
```
sethAmazon, line 16: Do we no longer want to run a on a schedule?

.github/workflows/perf.yml
sethAmazon: Please add a comment

.github/workflows/perf.yml
jefchien: No. If it runs after each CI run as intended, it should be up-to-date and a scheduled run should not be needed.

.github/workflows/perf.yml
jefchien: Added a -v/--version flag.

Aneurysm9 (approved):

anuraaga (approved):

Issue #816, opened 2021-12-20
[Security] Workflow CI.yml is using vulnerable action codecov/codecov-action


qazaqstan2025:
The workflow CI.yml is referencing action codecov/codecov-action using references v1.0.12. However this reference is missing the commit [c6a36d2dbbf6a734e6ef75af344acdf122fcd9f9](https://github.com/codecov/codecov-action/commits/c6a36d2dbbf6a734e6ef75af344acdf122fcd9f9) which may contain fix to the some vulnerability. 
The vulnerability fix that is missing by actions version could be related to: 
(1) CVE fix 
(2) upgrade of vulnerable dependency 
(3) fix to secret leak and others. 
Please consider to update the reference to the action.

khanhntd:
Hi @qazaqstan2025 , thanks for reporting this issue and this issue should be fixed since another reason is that the[ current version is deprecated](https://github.com/codecov/codecov-action#%EF%B8%8F--deprecration-of-v1) because of the underlying bash uploader. We will solve this issue in the upcoming days.

PR #820, opened 2021-12-21
Update Version for some GitHub Action

khanhntd:

Currently, there are some GitHub Actions that aren't up-to-date such as Codecov and it would not only cause vulnerability but the older version would be stopped supporting during Feb .Therefore, a fixed is need.

https://github.com/aws-observability/aws-otel-collector/issues/816

Using the PR build to test it

Deprecation of codecov v1: https://github.com/codecov/codecov-action#%EF%B8%8F--deprecration-of-v1

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -167,8 +167,9 @@ jobs:
 
       - name: Set up JDK 1.11
         if: steps.s3-release-validation-1.outputs.cache-hit != 'true'
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v2
         with:
+          distribution: 'zulu'
           java-version: 1.11
 
       - name: Set up terraform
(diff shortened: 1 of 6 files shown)


anuraaga (approved):

.github/workflows/CD.yml
```diff
@@ -165,11 +165,12 @@ jobs:
...
           aws-region: us-west-2
 
-      - name: Set up JDK 1.11
+      - name: Set up JDK 11
         if: steps.s3-release-validation-1.outputs.cache-hit != 'true'
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v2
         with:
-          java-version: 1.11
+          distribution: 'zulu'
```
sethAmazon, line 172: Do we need to specify distribution

sethAmazon (approved):

.github/workflows/CD.yml
khanhntd: Yes. Since the setup-java@v1 has a default options of ```Zulu distribution```. However, based on [setup-java@v2 documents](https://github.com/aws-observability/aws-otel-collector/pull/820), we need to specify the distribution.

PR #823, opened 2021-12-21
 Support roll back for CD when validation test are failing

khanhntd:

Currently, the CD has the fast-forward behavior, meaning it would keep the image on the public repo **aws-observability/aws-otel-collector** even thought the CD validation test fails. If the test fails, we will retry until it passed and while we keep retrying, customers will be able to retrieve the not-stable version image. Therefore, it would be best for us to delete the version-image if it fails and when it success, we will keep it there and mark it as latest also.

Testing each command and script separately and also locally. To be more specific:
- **Delete binaries uploading to S3:** Create a public s3 and test it with the command: cd tools/release && bash delete-s3-release.sh (using my s3)
- **Delete ECR Image Tag:** Create a public repo and test it with the command: ``` aws ecr-public batch-delete-image --repository-name adot-collector-test --image-ids imageTag=latest --region us-east-1``` (using my public repo)
- **Delete Docker Image Tag:** Create a private repo in docker registry hub and test it with the command cd tools/release && bash delete-docker-image-tag.sh (using my s3)

- **Delete Object from S3:** https://docs.aws.amazon.com/cli/latest/reference/s3/rm.html
- **Delete ECR Image Tag:** https://docs.aws.amazon.com/pt_br/cli/latest/reference/ecr-public/batch-delete-image.html
- **Delete Docker Image Tag:** https://stackoverflow.com/a/59334315


tools/release/image-binary-release/delete-docker-image-tag.sh
```diff
@@ -0,0 +1,44 @@
+#! /bin/bash
```
khanhntd, line 1: Yes @sethAmazon  this can be written in go but it would be more easier this way and also its hard for me to find any reason to use the package main go to write a similar script.

sethAmazon (approved):

tools/release/image-binary-release/delete-docker-image-tag.sh
```diff
@@ -0,0 +1,44 @@
...
+
+#Get access token from docker hub to authorize when deleting
+dockerhub_token=$(curl -s -X "POST" \
+  -H "Content-Type: application/json" \
+  -d "{\"username\": \"${dockerhub_username}\", \"password\": \"${dockerhub_password}\"}" \
+  "${dockerhub_url}/v2/users/login/" | jq -r .token)
+
+#Checking if the input username or input password is invalid (faster failing)
+if [ ${dockerhub_token} == "null" ];  then
+  echo "Username or password is not valid."
```
aateeqi, line 36: So here we are saying that the username or password is incorrect but would it be possible that the POST fails for a different reason?

aateeqi, line 36: So here we are saying that the username or password is incorrect but would it be possible that the POST fails for a different reason?

khanhntd: Mostly there are two possible errors I can think about: connection timeout and invalid credentials. However, the former error is less likely to happen and also there is another [docker push command before hand](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L371-L377). Therefore, the chance is slim. However, I guess you are right about it is best to log another errors thought.

aateeqi (approved):

PR #827, opened 2021-12-21
Add Retry To EC2 CI Tests

sethAmazon:
**Description:** 
Add retries to ci

**Testing:** 
Tested to see fi retry would trigger on fork

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -645,13 +645,15 @@ jobs:
 
       - name: Run testing suite on ec2
         if: steps.e2etest-ec2-1.outputs.cache-hit != 'true'
-        run: |
-          if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars" ; else opts="" ; fi
-          cd testing-framework/terraform/ec2 && terraform init && terraform apply -auto-approve -lock=false $opts -var="testing_ami=${{ matrix.testing_ami }}" -var="aoc_version=${{ needs.e2etest-preparation.outputs.version }}" -var="testcase=../testcases/${{ matrix.testcase }}"
-      - name: Destroy resources
-        if: ${{ always() && steps.e2etest-ec2-1.outputs.cache-hit != 'true' }}
-        run: |
-          cd testing-framework/terraform/ec2 && terraform destroy -auto-approve
+        uses: nick-invision/retry@v2
+        with:
+          max_attempts: 3
+          timeout_minutes: 120
+          command: |
+            if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars" ; else opts="" ; fi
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -707,13 +709,15 @@ jobs:
...
-          if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars" ; else opts="" ; fi
-          cd testing-framework/terraform/ec2 && terraform init && terraform apply -auto-approve -lock=false $opts -var="testing_ami=${{ matrix.testing_ami }}" -var="aoc_version=${{ needs.e2etest-preparation.outputs.version }}" -var="testcase=../testcases/${{ matrix.testcase }}"
-      - name: Destroy resources
-        if: ${{ always() && steps.e2etest-ec2-2.outputs.cache-hit != 'true'}}
-        run: |
-          cd testing-framework/terraform/ec2 && terraform destroy -auto-approve
+        uses: nick-invision/retry@v2
+        with:
+          max_attempts: 3
+          timeout_minutes: 120
```
aateeqi, line 715: How did we come up with 120 to be an appropriate timeout?

sethAmazon: Just a shot in the dark. This can be changed.

aateeqi (approved):

khanhntd (approved):
LGTM.

.github/workflows/CI.yml
```diff
@@ -645,13 +645,15 @@ jobs:
...
+          command: |
+            if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars" ; else opts="" ; fi
+            cd testing-framework/terraform/ec2
+            terraform init
+            if terraform apply -auto-approve -lock=false $opts -var="testing_ami=${{ matrix.testing_ami }}" -var="aoc_version=${{ needs.e2etest-preparation.outputs.version }}" -var="testcase=../testcases/${{ matrix.testcase }}" ; then terraform destroy -auto-approve ; else terraform destroy -auto-approve && exit 1 ; fi
```
jefchien, line 656: Does this work? Can you attach some screenshots with the steps of a successful retry? From my understanding, a failing terraform apply emits an exit code of 1 on failure and usually an if-then statement will not catch that.

sethAmazon:
Screen Shot 2021-12-22 at 11 33 38 AM
Screen Shot 2021-12-22 at 11 34 35 AM
Since I removed the packing step as I do not have signing creds in my fork ec2 tests fail. Thus we then destroy, exit with 1 then retry. https://unix.stackexchange.com/questions/22726/how-to-conditionally-do-something-if-a-command-succeeded-or-failed why command successful returns 0 but is true in bash I'm not sure. This link shows they get the same result.

sethAmazon:
https://stackoverflow.com/questions/2933843/why-0-is-true-but-false-is-1-in-the-shell#:~:text=By%20convention%2C%20all%20system%20calls,succeeded%2C%20non%2Dzero%20otherwise. so it seems 0 means true in bash

jefchien (approved):

PR #829, opened 2021-12-22
Always Destroy In Case Of Workflow Cancel

sethAmazon:
**Description:** 
In case of cancel destroy does not run

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -655,6 +655,12 @@ jobs:
             terraform init
             if terraform apply -auto-approve -lock=false $opts -var="testing_ami=${{ matrix.testing_ami }}" -var="aoc_version=${{ needs.e2etest-preparation.outputs.version }}" -var="testcase=../testcases/${{ matrix.testcase }}" ; then terraform destroy -auto-approve ; else terraform destroy -auto-approve && exit 1 ; fi
 
+      #This is here just in case workflow cancel
+      - name: Destroy resources
+        if: ${{ cancelled() && steps.e2etest-ec2-1.outputs.cache-hit != 'true' }}
+        run: |
+          cd testing-framework/terraform/ec2 && terraform destroy -auto-approve
+
   e2etest-ec2-2:
     runs-on: ubuntu-latest
     needs: [get-testing-suites, e2etest-release, e2etest-preparation]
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -655,6 +655,12 @@ jobs:
             terraform init
             if terraform apply -auto-approve -lock=false $opts -var="testing_ami=${{ matrix.testing_ami }}" -var="aoc_version=${{ needs.e2etest-preparation.outputs.version }}" -var="testcase=../testcases/${{ matrix.testcase }}" ; then terraform destroy -auto-approve ; else terraform destroy -auto-approve && exit 1 ; fi
```
khanhntd, line 656: If we are using always, then is there any reason for us to having the ```terraform destroy``` in this code?

sethAmazon: I guess not for the pass but we do for the fail since we restart this step with the retry action

khanhntd: Would it be better to use the retry with [this line](https://github.com/aws-observability/aws-otel-collector/pull/829/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R659-R660)? We can also retry if ```terraform destroy`` failed as well when we cancel the workflow as well?

sethAmazon: I don't think retry will trigger on cancel.

aateeqi (approved):

.github/workflows/CI.yml
```diff
@@ -655,6 +655,12 @@ jobs:
...
 
+      #This is here just in case workflow cancel
+      - name: Destroy resources
+        if: ${{ always() && steps.e2etest-ec2-1.outputs.cache-hit != 'true' }}
```
Aneurysm9, line 660:
```suggestion
        if: ${{ steps.e2etest-ec2-1.outputs.cache-hit != 'true' }}
```

if `always()` does what it says on the tin and always returns true, is it needed here?

sethAmazon: Yes because it forces run on cancel according to the docs https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions

.github/workflows/CI.yml
sethAmazon: This will now only trigger on cancel

Aneurysm9 (approved):

PR #832, opened 2021-12-23
Remove soaking test workflow.

jefchien:
**Description:** Add checks for commit benchmarking.

**Link to tracking Issue:** https://github.com/aws-observability/aws-otel-collector/issues/833

**Testing:** <Describe what testing was performed and which tests were added.> Has been tested on my fork.


anuraaga (approved):

tools/workflow/is-commit-benchmarked.sh
```diff
@@ -0,0 +1,34 @@
+#! /bin/bash
```
sethAmazon, line 1: Should we write this in go?

sethAmazon (approved):

tools/workflow/is-commit-benchmarked.sh
jefchien: I don't think it's necessary for all of our tooling to be in go. These tools are simple enough where it should be fine.

PR #838, opened 2021-12-28
change max parralel for eks from 10 to 5

khanhntd:

Currently, one test case in EKS consumes 2 classic load balancer. However, we are running [10 eks test case per CI ](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CI.yml#L867-L873) so up to 20 per CI in maximum. So the test case would be crash if up to 2 Ci are running simultaneously because of [reaching Load balancer quota](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html) (in this case, I am assuming other test cases are using LB as well such as ECS and EC2). 

Seeing 102 load balancers created on Friday, December 24th.
image

**Note**:
CD has already used [5 max-parallel with eks test case](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L423-L429)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -869,7 +869,7 @@ jobs:
     needs: [get-testing-suites, e2etest-release, e2etest-preparation]
     strategy:
       fail-fast: false
-      max-parallel: 10
+      max-parallel: 5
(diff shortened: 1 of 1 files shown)


sethAmazon (approved):

sethAmazon:
lgtm

PR #840, opened 2021-12-31
Convert private ecr to public ecr

khanhntd:

Currently, the CI publishes the image to private repo and we need to pull it from there for integration test. However, as we discussed with the adot team, converting from private to public ECR would give us two advantages:
* More visibility for the customer and dev to pull down the image without having the credentials.
* Able to use multiple regions later on.


jefchien:
What's the difference between https://github.com/aws-observability/aws-otel-collector/pull/730 and this one?

sethAmazon (approved):

.github/workflows/CI.yml
```diff
@@ -529,8 +529,8 @@ jobs:
         if: steps.e2etest-release.outputs.cache-hit != 'true'
         run: |
           docker load < build/packages/$IMAGE_NAME.tar
```
anuraaga, line 531: BTW, I'd like to remove docker load from the workflow - we should just push to the ECR repo instead of docker saving. Then we could use the official docker build github action, using buildx, and making arm64 build trivial

anuraaga (approved):

.github/workflows/CI.yml
alolita: @khanhntd imo the repo should be named adot-collector-tests ; we could potentially have other test images there too wdty?

alolita (commented):
let's fix the ECR repo name and in the readme

.github/workflows/CI.yml
khanhntd: True thought. Mostly I thought the docker saving there is for separating the e2e preparations part with the build parts. Other than that, I don't see any reasons to use ```docker save``` so I am down with using docker buildx for arm64.

khanhntd:
> let's fix the ECR repo name and in the readme

I'm down with that too. Having a hard time to decide the naming convention though.

.github/workflows/CI.yml
khanhntd: @alolita the only two tests that we have are: integration test from the CI and validation test from the CD. But the CD release is using ```public.ecr.aws/aws-observability/aws-otel-collector```; so the only one we are using for this repo is the CI. So IMO, if we are planning to add additional testing such as unit testing or config testing such as FB in the future, then we should go with yours; if we don't, then specific enough such as ```adot-collector-integration-test``` is fine since its only for the CI.

khanhntd: However, such tests are not feasible since we are a downstream repo so I cannot imagine we have addition tests beside integration test such as unit test or config test.

.aoc-stack-test.yml
```diff
@@ -4,4 +4,4 @@ s3ReleaseCandidateBucketName: "aws-otel-collector-release-candidate"
 s3BucketName: "aws-otel-collector-test"
 sshKeyS3BucketName: "aoc-ssh-key"
 traceDataS3BucketName: "trace-expected-data"  
-testingImageRepoName: "611364707713.dkr.ecr.us-west-2.amazonaws.com/aws/aws-otel-collector"
+testingImageRepoName: "public.ecr.aws/aws-otel-test/adot-collector-integration-test"
```
jefchien, line 7: What is this used for?

.github/workflows/CI.yml
```diff
@@ -520,6 +488,10 @@ jobs:
         if: steps.e2etest-release.outputs.cache-hit != 'true'
         run: s3_bucket_name=aws-otel-collector-test upload_to_latest=0 bash tools/release/image-binary-release/s3-release.sh
 
+      - name: Build Image
```
jefchien, line 491: Think we can just use https://github.com/marketplace/actions/build-and-push-docker-images

README.md
```diff
@@ -84,14 +84,19 @@ Use the following instructions to build your own AWS OTel Collector artifacts:
...
 ### Benchmark
 
 The latest performance model result is [here](https://github.com/aws-observability/aws-otel-collector/blob/main/docs/performance_model.md). 
 The performance test can be conducted by following the [instructions](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/docs/get-performance-model.md) here.
 
+### Support
+For each successful pull request, the corresponding test image would be pushed to 
+[public.ecr.aws/aws-otel-test/adot-collector-integration-test](https://gallery.ecr.aws/aws-otel-test/adot-collector-integration-test) 
+and pulled to validate with the integration test. Therefore, you can pull the testing image from there; however, we will not 
+support any issues and pull requests for these testing images.
```
jefchien, line 98:
```suggestion
For each merged pull request, a corresponding image is pushed to [public.ecr.aws/aws-otel-test/adot-collector-integration-test](https://gallery.ecr.aws/aws-otel-test/adot-collector-integration-test). This image is used for the integration tests. You can pull any of the images from there, however, we will not support any issues and pull requests for these test images.
```

anuraaga (commented):
Sorry didn't mean to do the build / push refactoring in this PR but thanks for helping with it :)

.github/workflows/CI.yml
khanhntd: Agreed. Seems the [load tar involves in the CD](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L336-L337) so I should do all of this in one PR. Will post another PR for this step soon.

khanhntd:
> Sorry didn't mean to do the build / push refactoring in this PR but thanks for helping with it :)
No worries. Thought the build step only involve [in the CI but it also involves the CD](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L336-L337) so need to move back to another PR :(

.aoc-stack-test.yml
khanhntd: I assume its purpose is for notifying what [resources are used with the release and testing](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CI.yml#L476-L479). Currently ask Ying to confirm about it.

sethAmazon (approved):

.aoc-stack-test.yml
khanhntd: Got a confirmation from @wyTrivail. So it's purpose is notify what resources are being used for the validation test and release validation. An example would be https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L187. Somehow we decide to define it explicitly instead of using a yaml file anymore. Seems weird for me to delete it since most of the release validation or integration test, its used the same bucket so either define a yaml or define an environment variable is better for me instead of defining it explicitly though.

PR #844, opened 2022-01-03
Support multiple regions

khanhntd:

Currently, some of the tests are bounded to one region. However, we can separate these tests such as the integration  test in two regions: us-west-2 and us-east-2. This will help us in increasing the capacities/ quota for each resource. My intention was to get all the ec2-test in region us-east-2 and other platform such as eks, ecs on region us-west-2.

**Link to tracking Issue:**
https://github.com/aws-observability/aws-otel-collector/issues/845

Testing locally with [each platform](https://github.com/aws-observability/aws-otel-test-framework) and adding variables ```-var="region={{ target region}}"```


sethAmazon:
I would split the ec2 tests in different regions but I think the ecs and eks can both be in us-west-2 since they have required clusters in the env.

.github/workflows/canary.yml
```diff
@@ -24,6 +24,7 @@ env:
   TF_VAR_aws_access_key_id: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
   TF_VAR_aws_secret_access_key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
   TF_VAR_patch: 'true'
+  CANARY_INTEGRATION_TEST_REGION: us-east-2
```
aateeqi, line 27: so we're using us-east-2 here and we were using us-west-2 before, is there a reason for that?

.github/workflows/CI.yml
```diff
@@ -42,6 +42,11 @@ env:
   WIN_SIGNED_PKG_BUCKET: aoc-aws-signer-signed-artifact-dest
   WIN_UNSIGNED_PKG_FOLDER: OTelCollectorAuthenticode/AuthenticodeSigner-SHA256-RSA
   WIN_SIGNED_PKG_FOLDER: OTelCollectorAuthenticode/AuthenticodeSigner-SHA256-RSA
+  INTEGRATION_TEST_FIRST_REGION: us-east-2
+  INTEGRATION_TEST_AMP_FIRST_REGION: "https://aps-workspaces.us-east-2.amazonaws.com/workspaces/ws-1de68e95-0680-42bb-8e55-67e7fd5d0861"
+  INTEGRATION_TEST_SECOND_REGION: us-west-2
+  INTEGRATION_TEST_AMP_SECOND_REGION: "https://aps-workspaces.us-west-2.amazonaws.com/workspaces/ws-cd278045-1d6e-4550-a207-c5046a1b40b8"
```
aateeqi, line 48: I feel like the naming of these variables can be a little bit confusing, could we rename them as `INTEGRATION_TEST_REGION_A` and B or something that makes it more explicit that it's part of a group?

.github/workflows/canary.yml
khanhntd: Not much. Mostly, I am still finding the right balancer when splitting these test into two different regions, maybe three at most though.

tools/workflow/clean_aws_resources.go
```diff
@@ -33,18 +33,25 @@ const (
 )
 
 func main() {
-	log.Printf("Beging terminating EC2 Instances")
-	terminateEc2Instances()
+	awsRegions := []string{"us-west-2","us-east-2"}
```
aateeqi, line 36: should we define these regions outside of the main func with the other const?

.github/workflows/CD.yml
```diff
@@ -407,8 +423,12 @@ jobs:
...
         run: |
-          if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars" ; else opts="" ; fi
-          cd testing-framework/terraform/ecs && terraform init && terraform apply -auto-approve -lock=false $opts -var="ecs_launch_type=${{ matrix.launch_type }}" -var="aoc_version=${{ github.event.inputs.version }}" -var="aoc_image_repo=$IMAGE_LINK" -var="testcase=../testcases/${{ matrix.testcase }}"
+          opts=""
+          if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then
+            opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars"
+          fi
+          cd testing-framework/terraform/ecs && terraform init \
+            && terraform apply -auto-approve -lock=false $opts -var="region=$INTEGRATION_TEST_SECOND_REGION" -var="ecs_launch_type=${{ matrix.launch_type }}" -var="aoc_version=${{ github.event.inputs.version }}" -var="aoc_image_repo=$IMAGE_LINK" -var="testcase=../testcases/${{ matrix.testcase }}" -var="cortex_instance_endpoint=$INTEGRATION_TEST_AMP_SECOND_REGION"
```
sethAmazon, line 431: You could probably remove the \
            &&
since it is using run: |

.github/workflows/CI.yml
sethAmazon: Maybe INTEGRATION_TEST_REGION_US_WEST_2 and INTEGRATION_TEST_REGION_US_EAST_2

khanhntd: ```INTEGRATION_TEST_REGION_A``` sounds similar to sub-region to me and mostly I don't specify the region since if we decided to change the region, we don't need to change the naming convention.

tools/workflow/clean_aws_resources.go
khanhntd: True though. I intend to passed these regions as environment variables but nice idea if I did not manage to do that.

.github/workflows/CD.yml
khanhntd: Reasonable. Fixed it in the newest commit though.

jefchien: This is done everywhere.

.github/workflows/CI.yml
aateeqi: @khanhntd I agree that it would sound like a sub-region. Could we instead do something like `INTEGRATION_TEST_REGION_ONE` and `INTEGRATION_TEST_REGION_TWO` because it would be easier to tell that these two variables are together.

jefchien:
```suggestion
  INTEG_TEST_1_REGION: us-east-2
  INTEG_TEST_1_AMP: "https://aps-workspaces.us-east-2.amazonaws.com/workspaces/ws-1de68e95-0680-42bb-8e55-67e7fd5d0861"
  INTEG_TEST_2_REGION: us-west-2
  INTEG_TEST_2_AMP: "https://aps-workspaces.us-west-2.amazonaws.com/workspaces/ws-cd278045-1d6e-4550-a207-c5046a1b40b8"
```

.github/workflows/weekly-resources-clean.yml
```diff
@@ -15,7 +15,7 @@ name: Weekly Resources Cleaner
 
 on:
   schedule:
-    - cron: "0 0 * * SAT" # Run every saturday at midnight
+    - cron: "0 0 */5 * *" # Run every 5 days
```
jefchien, line 18: What's this change for?

tools/workflow/clean_aws_resources.go
```diff
@@ -33,18 +33,25 @@ const (
...
 	log.Printf("Finish destroy AWS resources")
 }
 
-func terminateEc2Instances() {
+func terminateEc2Instances(awsRegion string) {
 	// set up aws go sdk ec2 client
-	testSession, err := session.NewSession()
+	testSession, err := session.NewSession(&aws.Config{
+		Region: aws.String(awsRegion),
+	})
```
jefchien, line 54:
```suggestion
	testSession, err := session.NewSession(aws.NewConfig().WithRegion(awsRegion))
```

tools/workflow/clean_aws_resources.go
```diff
@@ -33,18 +33,25 @@ const (
...
 
-	log.Printf("Begin destroy Load Balancer resources")
-	destroyLoadBalancerResource()
+	for _, region:= range awsRegions {
+		log.Printf("Beging terminating EC2 Instances %s", region)
+		terminateEc2Instances(region)
+
+		log.Printf("Begin destroy Load Balancer resources %s", region)
+		destroyLoadBalancerResource(region)
+
```
jefchien, line 44: nitpick: extra newline

.github/workflows/CI.yml
```diff
@@ -1090,4 +1146,9 @@ jobs:
       - name: clean up SSM test package
         run: |
           ssm_pkg_version=`cat build/packages/VERSION`
-          aws ssm delete-document --name "testAWSDistroOTel-Collector" --version-name ${ssm_pkg_version}
+          for region in $INTEGRATION_TEST_FIRST_REGION $INTEGRATION_TEST_SECOND_REGION
+          do
+              aws ssm describe-document --name ${SSM_PACKAGE_NAME} --version-name ${ssm_pkg_version} --region ${region} >/dev/null 2>&1 && \
```
jefchien, line 1151: Why are you describing the document before deleting it? It'll always exist.

khanhntd: It depends though. In some way if we had an error before the SSM was created, then we cannot delete what's not there. Got an error when testing with one of my PR.

.github/workflows/weekly-resources-clean.yml
khanhntd: Since we destroy every resources that are created by the integration test framework for 5 days; therefore, it would be better for us to set the schedule in every 5 days thought.

tools/workflow/clean_aws_resources.go
khanhntd: Delete extra newline. Fixed it with the newest commit.

.github/workflows/CI.yml
khanhntd: Reasonable. Fixed it to ```INTEGRATION_TEST_REGION_1``` and similar variables. Nice opinions thought.

.github/workflows/CD.yml
```diff
@@ -36,6 +36,14 @@ env:
   TF_VAR_aoc_vpc_name: aoc-vpc-large
   TF_VAR_aoc_vpc_security_group: aoc-vpc-security-group-large
   TESTING_FRAMEWORK_REPO: 'aws-observability/aws-otel-collector-test-framework'
+  VALIDATION_TEST_REGION_1: us-east-2
+  VALIDATION_TEST_AMP_1: "https://aps-workspaces.us-east-2.amazonaws.com/workspaces/ws-1de68e95-0680-42bb-8e55-67e7fd5d0861"
+  VALIDATION_TEST_REGION_2: us-west-2
```
aateeqi, line 41: Nice 👍

.github/workflows/CI.yml
```diff
@@ -1075,5 +1131,8 @@ jobs:
         run: |
           ssm_pkg_version=`cat build/packages/VERSION`
           aws ssm describe-document --name ${SSM_PACKAGE_NAME} --version-name ${ssm_pkg_version} >/dev/null 2>&1 && \
-            aws ssm delete-document --name ${SSM_PACKAGE_NAME} --version-name ${ssm_pkg_version}
+            for region in $INTEGRATION_TEST_REGION_1 $INTEGRATION_TEST_REGION_2
```
aateeqi, line 1134: are we planning on adding more test regions? because if its only two it may be simpler to just write the 2 commands

khanhntd: Yes. I assume we would add more ask our test growing. For now, I'm fine with YAGNI.

jefchien: Might be better to use a matrix strategy.
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

.github/workflows/CD.yml
```diff
@@ -122,7 +130,7 @@ jobs:
           echo ${{ steps.set-matrix.outputs.ec2-matrix-2 }}
           echo ${{ steps.set-matrix.outputs.ec2-matrix-3 }}
           
-  release-to-s3:
+  release-version-preparation:
```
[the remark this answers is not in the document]
khanhntd: At first I thought it would be better if we support pull image arm, amd64 on docker for ec2 test, the same as the CI workflow. Therefore, I combined both of these to in case of that since they are already supported on there thought. However, it would be best to do it in another PR or when ever we needs it though (YAGNI).

.github/workflows/CI.yml
khanhntd: Reasonable. Fixed it for the newest commit. For ```ec2-preparation``` job, I thought of using it too but it ways too overkill for this PR. Need to separate the release S3 bucket - SSM package name and  build-image. Added matrix in the CD too and for resources weekly cleaning.

alolita:
@jefchien @aateeqi @sethAmazon @anuraaga can you please re-review?

.github/workflows/weekly-resources-clean.yml
alolita: This should be weekly - on a weekday e.g. mondays

khanhntd: I agree that we could delete the resources weekly thought; however, IMO it should be better to keep it up with past day delete (e.g if we decide to delete EC2 instances that have existed over the past 5 days, then we could delete every instances in every 5 days though.

aateeqi (approved):

tools/workflow/clean_aws_resources.go
```diff
@@ -66,10 +74,14 @@ func terminateEc2Instances() {
 		describeInstancesInput := ec2.DescribeInstancesInput{Filters: []*ec2.Filter{&instanceStateFilter, &instanceTagFilter}, NextToken: nextToken}
 		describeInstancesOutput, err := ec2client.DescribeInstances(&describeInstancesInput)
 
+
+
 		if err != nil {
 			log.Fatalf("Failed to get instance for error %v", err)
 		}
 
+		fmt.Print(describeInstancesOutput)
```
jefchien, line 83: Use log.

khanhntd: Forgot to delete when testing yesterday. Deleted it in the newest commit.

.github/workflows/CD.yml
```diff
@@ -121,10 +129,23 @@ jobs:
           echo ${{ steps.set-matrix.outputs.ec2-matrix-1 }}
           echo ${{ steps.set-matrix.outputs.ec2-matrix-2 }}
           echo ${{ steps.set-matrix.outputs.ec2-matrix-3 }}
-          
-  release-to-s3:
+
+  get-release-region:
```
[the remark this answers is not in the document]
khanhntd: Yes. Because matrix does not support environment variables currently (tested it out with my branch) and only support hard code one, therefore, I use similar idea with get-test-suit it the CI to do the dynamic variable.

tools/workflow/clean_aws_resources.go
```diff
@@ -16,35 +16,43 @@
...
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/elbv2"
+	"fmt"
+	"log"
+	"os"
+	"strings"
+	"time"
```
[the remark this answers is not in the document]
khanhntd: Even thought I used the go fmt. its gave me the same result. However, changed back to the original import if that's good.

.github/workflows/CI.yml
```diff
@@ -487,15 +493,20 @@ jobs:
         with:
           aws-access-key-id: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
-          aws-region: us-west-2
+          aws-region: ${{ env.INTEGRATION_TEST_REGION_2 }}
 
       - name: Create SSM package
```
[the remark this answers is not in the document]
khanhntd: I thought it would be too overkill for separate releases ssm and release image in this PR and do it in another PR. However, its fine to do this in this PR.

tools/workflow/clean_aws_resources.go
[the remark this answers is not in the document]
jefchien: That's bizarre. `goimports` should run. I tried copying your changes and running the formatter and it fixed it.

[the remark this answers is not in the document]
khanhntd: Forgot the ```goimports```. It works now with ```make fmt```. Nice catch.

.github/workflows/weekly-resources-clean.yml
```diff
@@ -18,9 +18,25 @@ on:
...
+      - name: Get Region to clean SSM package
+        id: set-matrix
+        run: echo "::set-output name=region-matrix::{\"include\":[{\"region\":\"${{ env.REGION_1 }}\"},{\"region\":\"${{ env.REGION_2 }}\"}]}"
+
   clean-soaking-resources:
+    needs: [ get-resources-cleaning-region ]
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix: ${{fromJson(needs.get-resources-cleaning-region.outputs.region-matrix)}}
```
jefchien, line 39: The environment variables aren't being used anywhere else. Keep it simple.
```suggestion
      matrix:
        region: [us-east-2, us-west-2]
```

tools/workflow/clean_aws_resources.go
```diff
@@ -97,11 +105,11 @@ func terminateEc2Instances() {
 	}
 }
 
-func destroyLoadBalancerResource() {
+func destroyLoadBalancerResource(awsRegion string) {
 	// Set up aws go sdk session
 	// Only using default environment variables instead of loading other metadata from session.NewSessionWithOptions
 	//Documents: https://docs.aws.amazon.com/ja_jp/sdk-for-go/v1/developer-guide/configuring-sdk.html
-	testSession, err := session.NewSession()
+	testSession, err := session.NewSession(aws.NewConfig().WithRegion(awsRegion))
```
jefchien, line 112: Can we reuse the session and just pass it into these functions instead of the region string?

.github/workflows/weekly-resources-clean.yml
khanhntd: Thought it the same as first but yeah, let's just keep it simple for now.

.github/workflows/CD.yml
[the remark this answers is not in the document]
jefchien: Didn't know that. It's so strange that it doesn't work.

tools/workflow/clean_aws_resources.go
khanhntd: Yes. We can delete the getting the environment variables and use the configure credentials with Session. Fixed it with newest commit

.github/workflows/CD.yml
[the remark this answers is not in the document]
khanhntd: Same.I got the startup failure many times so.

.github/workflows/weekly-resources-clean.yml
```diff
@@ -29,7 +33,7 @@ jobs:
         with:
           aws-access-key-id: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
-          aws-region: us-west-2
+          aws-region: ${{ matrix.region }}
 
       - name: Clean old aws resources
         run: go run tools/workflow/clean_aws_resources.go
```
jefchien, line 39: I think you still need the `AWS_REGION` environment variable for it to use it.

jefchien: https://docs.aws.amazon.com/sdk-for-go/api/aws/session/
Doesn't seem to have anchors, so I can't link directly to it, but check out the "Environment Variables" section.

khanhntd: From what I am understanding thought, ```AWS_SDK_LOAD_CONFIG``` is turn on; therefore, the session will use the the shared config (~/.aws/config) and shared credentials (~/.aws/credentials) files. Moreover, matrix is in separate instances. So I don't see why we should use the ```AWS_REGION``` though

jefchien (approved):

PR #849, opened 2022-01-04
Fix the lumberjack logger to ensure that all context fields are recorded

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

**Description:** The auto-rotating file writer added by the `logger` package is integrated with the collector's `zap` logger as a hook, which does not give it access to additional context fields on the logger.  This changes the integration to use a core wrapper that tees the log events to the collector's logger and to the `lumberjack` file writer.

### DIFF
--- a/pkg/logger/logger_test.go
+++ b/pkg/logger/logger_test.go
@@ -23,24 +23,12 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-	"go.uber.org/zap/zapcore"
 	"gopkg.in/natefinch/lumberjack.v2"
 )
 
 func setupLogEnv() {
 	logfile = getLogFilePath()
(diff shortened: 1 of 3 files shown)


pkg/logger/logger_test.go
```diff
@@ -33,16 +31,6 @@ func setupLogEnv() {
 	lumberjackLogger = tryNewLumberJackLogger()
 }
 
-func TestGetLumberHook(t *testing.T) {
```
jefchien, line 36: Please replace this unit test with one for the `WrapCoreOpt`.

Aneurysm9: This test had zero value and a test for `WrapCoreOpt` would likely also have zero value.

jefchien: That's fair. I agree that the original test didn't really do anything and I suppose a new one would just be validating the config.

jefchien (approved):

PR #858, opened 2022-01-06
Use Docker Login For Public ECR Instead Of AWS ECR Login

sethAmazon:
**Description:** 
Can't log in to public ecr

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -522,7 +522,11 @@ jobs:
       - name: Login ECR
         if: steps.e2etest-release.outputs.cache-hit != 'true'
         id: login-ecr
-        uses: aws-actions/amazon-ecr-login@v1
+        uses: docker/login-action@v1
+        with:
+          registry: public.ecr.aws
+          username: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
+          password: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
 
       - name: upload to ECR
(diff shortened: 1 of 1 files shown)


aateeqi (approved):

.github/workflows/CI.yml
```diff
@@ -523,6 +523,10 @@ jobs:
         if: steps.e2etest-release.outputs.cache-hit != 'true'
         id: login-ecr
         uses: aws-actions/amazon-ecr-login@v1
+        with:
```
jefchien, line 526: I don't think this action takes the credentials. https://github.com/aws-actions/amazon-ecr-login#credentials-and-region
That's done by `aws-actions/configure-aws-credentials@v1`, which happens in an earlier step.

jefchien:
Perhaps we can use https://github.com/docker/login-action#aws-public-elastic-container-registry-ecr
```
uses: docker/login-action@v1
with:
  registry: public.ecr.aws
  username: ${{ secrets.AWS_ACCESS_KEY_ID }}
  password: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
env:
  AWS_REGION: <region>
```

.github/workflows/CI.yml
khanhntd: True. I think you are using this instead of aws-actions/amazon-ecr-login@1 aws-actions/amazon-ecr-login@v1

khanhntd (approved):
LGTM.

jefchien (approved):

Issue #846, opened 2022-01-04
Remove usages of docker load from workflows


anuraaga:
We have an ECR repository for storing docker images, there shouldn't be a reason to be using docker load within workflows as it would be equivalent to just docker pull. Not using docker load will allow using official docker github actions / buildx and make it trivial to build multiarch docker images.

_Originally posted by @anuraaga in https://github.com/aws-observability/aws-otel-collector/pull/840#discussion_r777770149_

PR #758, opened 2021-11-19
Change CI to always run create SSM package if it doesn't exist.

jefchien:
**Description:** The SSM document was being removed at the end of each run and not being recreated on a re-run of the CI. This was causing SSM tests that weren't cached to fail on these re-run attempts. Changed the CI to always check if the version exists before skipping the step.

**Link to tracking Issue:** https://github.com/aws-observability/aws-otel-collector/issues/695#issuecomment-964316357

PR #860, opened 2022-01-06
add buildtool kit for adot collector image

khanhntd:

Eliminate docker load and docker save, instead using [GitHub Action Build Push Action](https://github.com/docker/build-push-action) to support building adot collector image with build.

https://github.com/aws-observability/aws-otel-collector/issues/846

Finding a way to test. Currently seeing https://github.com/nektos/act but another potential ways is to use dev branch.

Build Push Action: https://github.com/docker/build-push-action


.github/workflows/CI.yml
```diff
@@ -350,22 +350,67 @@ jobs:
...
           path: build
 
-      - name: Build Image
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
```
jefchien, line 366: Why are we logging into the dockerhub? We aren't pushing the CI images there. Why not just login to the public ECR and build/push in one step?

.github/workflows/CI.yml
```diff
@@ -350,22 +350,67 @@ jobs:
...
+          password: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
+        env:
+          AWS_REGION: us-east-1
+
+      - name: upload to ECR
         if: steps.cached_image.outputs.cache-hit != 'true'
         run: |
-          mkdir -p $PACKAGING_ROOT
-          docker save --output $PACKAGING_ROOT/$IMAGE_NAME.tar $IMAGE_NAME
+          docker tag $IMAGE_NAME:latest public.ecr.aws/$ECR_REPO:${{ needs.e2etest-preparation.outputs.version }}
```
jefchien, line 405: Can just use the action to do the tagging/pushing.

.github/workflows/CI.yml
khanhntd: Yes we can do that. I am testing in different branch so this is not the final built yet. Still trying to figure it out how to pass data between jobs and steps at the same time so we do not need to use docker load or docker save.

.github/workflows/CI.yml
khanhntd: True. I was testing different but was not able to do it with the sub_branch. So I changed my latest commit in another branch and test it there.

.github/workflows/CI.yml
```diff
@@ -343,29 +344,76 @@ jobs:
...
+          registry: public.ecr.aws
+          username: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
+          password: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
+        env:
+          AWS_REGION: us-east-1
+
+      - name: Versioning for adot collector image
+        id: versioning
+        run: |
+          Version="`cat VERSION`-$GITHUB_RUN_ID"
```
anuraaga, line 374: If it's same behavior as before can be in another PR but shouldn't this use commit ID instead of run ID?

.github/workflows/CI.yml
```diff
@@ -343,29 +344,76 @@ jobs:
...
+          file: cmd/awscollector/Dockerfile
+          context: .
+          push: true
+          tags: public.ecr.aws/${{ env.ECR_REPO }}:${{ steps.versioning.outputs.version }}
+
+      #Export docker artifact since push and export cannot be in the same steps
+      #There were many different ways such as pulling the image and save as a tar file but
+      #we would like to get rid of load and also it would be better for us to build a second time
+      #since the image layers have been built before.
+      - name: Export ADOT Collector Image's artifact
```
anuraaga, line 395: Why do we still need to export into an artifact? We should be able to pull directly anywhere that is currently fetching the artifact

.github/workflows/CI.yml
anuraaga: Also good if we can use the metadata action

https://github.com/docker/metadata-action

.github/workflows/CI.yml
khanhntd: True thought. At first I thought we would need to export the artifcact in order to load the image for the CD. However, we can [pull directly from the public ECR and retag it. Fixed this.](https://github.com/aws-observability/aws-otel-collector/pull/860/files#diff-a64cbb6e953e697cb4302ab31af9ebbd95a8de6216f874bb0ffdefce8e3c1cd9R340-R343)

.github/workflows/CI.yml
khanhntd: True thought. If it is the same behavior, using Commit ID would be better than Run ID.

.github/workflows/CD.yml
```diff
@@ -320,27 +321,32 @@ jobs:
...
         run: |
           curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
           unzip awscliv2.zip
           sudo ./aws/install --update
 
-      - name: Load Image
-        run: docker load < build/packages/$IMAGE_NAME.tar
-        
-      - name: upload to ECR
+      - name: Pull image from integration test ECR
```
jefchien, line 340: Can use https://github.com/akhilerm/tag-push-action to pull/retag/push. I think you should be able to login to both of the public ECRs before using the action.

.github/workflows/CI.yml
```diff
@@ -75,82 +76,82 @@ jobs:
...
-      if: steps.cached_binaries.outputs.cache-hit != 'true'
-      run: make build
-
-    # upload the binaries to artifact as well because cache@v2 hasn't support windows
-    - name: Upload
-      uses: actions/upload-artifact@v2
-      with:
-        name: binary_artifacts
-        path: build
+      # Set up building environment, patch the dev repo code on dispatch events.
```
jefchien, line 79: Are these all just formatting changes?

.github/workflows/CI.yml
jefchien: Where is `$GITHUB_RUN_ID` coming from? Shouldn't it be `${{ github.run_id }}`. I agree that a short commit SHA would be more useful since no one knows the run IDs, but that change should be in another PR.

.github/workflows/CD.yml
```diff
@@ -552,11 +555,18 @@ jobs:
...
+          REPO_NAME="$IMAGE_NAMESPACE/$IMAGE_NAME"
+          docker tag public.ecr.aws/$ECR_REPO_INTEGRATION_TEST:$TESTING_TAG $REPO_NAME:$TAG
+          docker push $REPO_NAME:$TAG
+
+      - name: Tag latest and upload to release dockerhub
         if: steps.version.outputs.any-update == 'true'
         run: |
           REPO_NAME="$IMAGE_NAMESPACE/$IMAGE_NAME"
-          docker tag $IMAGE_NAME $REPO_NAME:latest
+          docker tag public.ecr.aws/$ECR_REPO_INTEGRATION_TEST:$TESTING_TAG $REPO_NAME:latest
```
jefchien, line 569: Where are you setting this `$TESTING_TAG`?

.github/workflows/CD.yml
```diff
@@ -320,27 +321,32 @@ jobs:
...
-          aws-secret-access-key: ${{ secrets.RELEASE_SECRET }}
-          aws-region: us-west-2    
-          
+          registry: public.ecr.aws
+          username: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
+          password: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
+        env:
+          AWS_REGION: us-east-1
+
       - name: Install AWS Cli 2.0
```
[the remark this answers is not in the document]
khanhntd: Reasonable. Fixed it.

.github/workflows/CD.yml
```diff
@@ -349,14 +355,6 @@ jobs:
           username: "${{ secrets.DOCKERHUB_RELEASE_USERNAME }}"
           password: "${{ secrets.DOCKERHUB_RELEASE_TOKEN }}"
 
-      - name: Push Image to dockerhub
```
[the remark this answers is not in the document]
khanhntd: We only need to push image to DockerHub when validation test success. In this way, we don't need to support the roll back for deleting image on DockerHub. Moreover, the validation test is only using the images from our public ECR so I don't see any reasons to push image before the validation test begins. One thing to note is that I move the [pushing to DockerHub step after the validation test](support_buildx_docker), so we still push the image to DockerHub

.github/workflows/CD.yml
```diff
@@ -320,27 +321,32 @@ jobs:
...
+          docker pull public.ecr.aws/$ECR_REPO_INTEGRATION_TEST:$TESTING_TAG
+
+
+      - name: Upload to Public Release ECR
         run: |
           TAG=`cat build/packages/VERSION`
-          aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
-          docker tag $IMAGE_NAME public.ecr.aws/$ECR_REPO:$TAG
+          docker tag public.ecr.aws/$ECR_REPO_INTEGRATION_TEST:$TESTING_TAG public.ecr.aws/$ECR_REPO:$TAG
           docker push public.ecr.aws/$ECR_REPO:$TAG
```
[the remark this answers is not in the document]
khanhntd: Yes. There is already [a log in step](https://github.com/aws-observability/aws-otel-collector/pull/860/files#diff-a64cbb6e953e697cb4302ab31af9ebbd95a8de6216f874bb0ffdefce8e3c1cd9R324-R332) to push it to the release repo.

.github/workflows/CD.yml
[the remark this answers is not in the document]
jefchien: So, what's the DockerHub login for?

.github/workflows/CD.yml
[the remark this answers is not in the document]
jefchien: That login step uses the integ test account credentials instead of the release account one.

.github/workflows/CD.yml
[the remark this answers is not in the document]
khanhntd: Deleted it in the newest commit. Nice catch.

.github/workflows/CD.yml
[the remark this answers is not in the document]
khanhntd: Reasonable. Fixed it in the newest commit.

.github/workflows/CD.yml
khanhntd: Fixed in it the newest commit. Set these [in the versioning step and release-version-image jobs ](https://github.com/aws-observability/aws-otel-collector/pull/860/files#diff-a64cbb6e953e697cb4302ab31af9ebbd95a8de6216f874bb0ffdefce8e3c1cd9R334-R340)

.github/workflows/CI.yml
khanhntd: Yes. Formatting changes. I tested the workflow in a different branch so when copy paste, the format changes too.

.github/workflows/CD.yml
khanhntd: Fixed it. Currently using akhilerm/tag-push-action. An example would be [this](https://github.com/aws-observability/aws-otel-collector/pull/860/files#diff-a64cbb6e953e697cb4302ab31af9ebbd95a8de6216f874bb0ffdefce8e3c1cd9R525-R539)

.github/workflows/CI.yml
khanhntd: Its the [same so](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) should be fine. Agree to be in another PR.

sethAmazon (approved):

.github/workflows/CI.yml
```diff
@@ -504,39 +530,31 @@ jobs:
...
-        uses: docker/login-action@v1
-        with:
-          registry: public.ecr.aws
-          username: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
-          password: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
-        env:
-          AWS_REGION: us-east-1
-
-      - name: upload to ECR
+      - name: Find if adot collector image exist in public ecr
```
jefchien, line 549: Would it make sense to remove the `packaging-image` job and just have it done here instead? `e2etest-preparation` doesn't actually have a dependency on that job and `packaging-image` is a misleading job anyway.

.github/workflows/CD.yml
```diff
@@ -523,41 +499,44 @@ jobs:
...
+          username: ${{ secrets.RELEASE_KEY_ID }}
+          password: ${{ secrets.RELEASE_SECRET }}
+        env:
+          AWS_REGION: us-east-1
 
       - name: Compare version with Dockerhub latest
         id: version
         run: |
-          TAG=`cat build/packages/VERSION`
+          TAG="${{ needs.release-version-image.outputs.tag }}"
```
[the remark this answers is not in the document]
khanhntd: Weird. Set it [here](https://github.com/aws-observability/aws-otel-collector/pull/860/files#diff-a64cbb6e953e697cb4302ab31af9ebbd95a8de6216f874bb0ffdefce8e3c1cd9R339-R340). Maybe I'm setting it wrong or somehow?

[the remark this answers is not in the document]
jefchien: That sets the output for the step. It doesn't set it for the job.

.github/workflows/CD.yml
```diff
@@ -523,41 +499,44 @@ jobs:
...
-          docker tag $IMAGE_NAME public.ecr.aws/$ECR_REPO:latest
-          docker push public.ecr.aws/$ECR_REPO:latest
-
-      - name: Login Dockerhub
-        uses: docker/login-action@v1
+        uses: akhilerm/tag-push-action@v2.0.0
         with:
-          username: "${{ secrets.DOCKERHUB_RELEASE_USERNAME }}"
-          password: "${{ secrets.DOCKERHUB_RELEASE_TOKEN }}"
+          src: public.ecr.aws/$ECR_REPO_INTEGRATION_TEST:${{ needs.release-version-image.outputs.testing_tag }}
```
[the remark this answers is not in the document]
khanhntd: Since the image is based on condition ```if: steps.version.outputs.any-update == 'true'```. Therefore, if the condition is not true, we cannot find the re-tagged image though. Unless, if we want to rotate the steps ( push versioning image first then mark latest later) then we could do it though.

.github/workflows/CD.yml
```diff
@@ -320,42 +321,30 @@ jobs:
...
-      - name: upload to ECR
+      - name: Get version tag and testing version tag
+        id: versioning
         run: |
-          TAG=`cat build/packages/VERSION`
-          aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
-          docker tag $IMAGE_NAME public.ecr.aws/$ECR_REPO:$TAG
-          docker push public.ecr.aws/$ECR_REPO:$TAG
+          Tag=`cat build/packages/VERSION`
+          Testing_tag=`cat $PACKAGING_ROOT/TESTING_VERSION`
```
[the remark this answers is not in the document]
khanhntd: Reasonable. Fixed it in the [newest commit](https://github.com/aws-observability/aws-otel-collector/pull/860/files#diff-a64cbb6e953e697cb4302ab31af9ebbd95a8de6216f874bb0ffdefce8e3c1cd9R29)

.github/workflows/CD.yml
```diff
@@ -320,42 +321,30 @@ jobs:
...
-      - name: Push Image to dockerhub
-        id: push-to-dockerhub
-        run: |
-          TAG=`cat build/packages/VERSION`
-          REPO_NAME="$IMAGE_NAMESPACE/$IMAGE_NAME"
-          docker tag $IMAGE_NAME $REPO_NAME:$TAG
-          docker push $REPO_NAME:$TAG
+          src: public.ecr.aws/$ECR_REPO_INTEGRATION_TEST:${{ steps.versioning.outputs.testing_tag }}
+          dst: |
+            public.ecr.aws/$ECR_REPO:${{ steps.versioning.outputs.tag }}
```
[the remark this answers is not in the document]
khanhntd: Reasonable. Fixed it.

.github/workflows/CD.yml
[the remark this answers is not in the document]
khanhntd: Reasonable. New things to learn lol. Thanks for the catch.

.github/workflows/CI.yml
khanhntd: Make sense. Mostly I thought that ```packaging-image``` would be sufficient for building the image and push it to the publish ECR. Then for the ```e2etest-preparation```, we only need to check whether the image is exist or now.

.github/workflows/CD.yml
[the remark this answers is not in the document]
khanhntd: Reasonable. Fixed it in the newest commit

.github/workflows/CD.yml
```diff
@@ -523,41 +499,44 @@ jobs:
           key: "${{ env.PACKAGE_CACHE_KEY }}"
           path: build/packages
 
-      - name: Load Image
-        run: docker load < build/packages/$IMAGE_NAME.tar
+      - name: Login Dockerhub
+        uses: docker/login-action@v1
```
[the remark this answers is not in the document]
khanhntd: Reasonable. Fixed it in the newest commit with the cache

.github/workflows/CD.yml
```diff
@@ -523,41 +499,44 @@ jobs:
...
       - name: Compare version with Dockerhub latest
         id: version
         run: |
-          TAG=`cat build/packages/VERSION`
+          TAG="${{ needs.release-version-image.outputs.tag }}"
           TARGET_VERSION=$TAG bash tools/workflow/docker-version-compare.sh
 
-      - name: Tag latest and upload to ecr
+      - name: Pull image from integration test ECR, tag as latest and push to public release ECR and DockerHub
         if: steps.version.outputs.any-update == 'true'
```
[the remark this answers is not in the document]
khanhntd: Reasonable. Fixed it in the newest commit with the cache

.github/workflows/CD.yml
```diff
@@ -517,50 +497,64 @@ jobs:
...
-          aws-region: us-west-2
+          key: "${{ env.PACKAGE_CACHE_KEY }}"
+          path: $PACKAGING_ROOT
 
       - name: Compare version with Dockerhub latest
         id: version
+        if: steps.release-latest-image.outputs.cache-hit != 'true'
         run: |
-          TAG=`cat build/packages/VERSION`
+          TAG="${{ needs.release-version-image.outputs.tag }}"
```
jefchien, line 519: This won't work because this job doesn't currently have `release-version-image` in its needs field.

.github/workflows/CD.yml
```diff
@@ -517,50 +497,64 @@ jobs:
...
 
-      - name: Tag latest and upload to docker
-        if: steps.version.outputs.any-update == 'true'
-        run: |
-          REPO_NAME="$IMAGE_NAMESPACE/$IMAGE_NAME"
-          docker tag $IMAGE_NAME $REPO_NAME:latest
-          docker push $REPO_NAME:latest
+      - name: Login to Public Integration Test ECR
+        id: login-ecr
+        if: ${{ steps.version.outputs.any-update == 'true' && steps.release-latest-image.outputs.cache-hit != 'true' }}
```
jefchien, line 531: Check for the cache miss first. If there's a cache hit, the `version` step will be skipped.

.github/workflows/CD.yml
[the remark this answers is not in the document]
jefchien: Maybe I'm missing something, but isn't the image pushed to the release ECR as part of `release-version-image`?

.github/workflows/CI.yml
```diff
@@ -493,50 +451,74 @@ jobs:
...
+          path: $PACKAGING_ROOT
+
+      - name: Restore cached binaries
+        if: steps.e2etest-release.outputs.cache-hit != 'true'
+        uses: actions/cache@v2
+        with:
+          key: "cached_binaries_${{ github.run_id }}"
+          path: build
+
+      - name: Set up QEMU
```
sethAmazon, line 467: why does QEMU need to be setup

.github/workflows/CD.yml
```diff
@@ -318,44 +323,31 @@ jobs:
...
-        
-      - name: upload to ECR
+      - name: Get version tag and testing version tag
+        id: versioning
         run: |
-          TAG=`cat build/packages/VERSION`
-          aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
-          docker tag $IMAGE_NAME public.ecr.aws/$ECR_REPO:$TAG
-          docker push public.ecr.aws/$ECR_REPO:$TAG
+          Tag=`cat $PACKAGING_ROOT/VERSION`
```
jefchien, line 341: Instead of getting the version here, I would suggest outputting the version as part of https://github.com/aws-observability/aws-otel-collector/blob/85c027bf2347be126a77a48e308f9cfc64765b83/.github/workflows/CD.yml#L57-L68 and then adding the `release-checking` job to the `needs` of each job that needs the version. Would still name them `version` and `testing-version` instead of `tag` and `testing_tag` too in case they could be used elsewhere.

.github/workflows/CI.yml
```diff
@@ -493,50 +451,74 @@ jobs:
...
         if: steps.e2etest-release.outputs.cache-hit != 'true'
         run: s3_bucket_name=aws-otel-collector-test upload_to_latest=0 bash tools/release/image-binary-release/s3-release.sh
 
-      - name: Login to Public ECR
+      #Build the adot collector image for two primary reasons:
+      #-Using the adot collector image to do the integration test
+      #-Export it for delivery version image in CD
+      #Documentation: https://github.com/docker/build-push-action
+      - name: Build adot collector image
+        id: docker_build
```
jefchien, line 512: Don't need an `id` if it isn't going to be used.

.github/workflows/CI.yml
```diff
@@ -493,50 +451,74 @@ jobs:
...
-        run: |
-          docker load < build/packages/$IMAGE_NAME.tar
-          docker tag $IMAGE_NAME public.ecr.aws/$ECR_REPO:${{ needs.e2etest-preparation.outputs.version }}
-          docker push public.ecr.aws/$ECR_REPO:${{ needs.e2etest-preparation.outputs.version }}
+          file: cmd/awscollector/Dockerfile
+          context: .
+          push: true
+          tags: public.ecr.aws/${{ env.ECR_REPO }}:${{ needs.e2etest-preparation.outputs.version }}
+          cache-from: type=registry
+          cache-to: type=inline
```
jefchien, line 521: What are these cache fields used for? Aren't we only running this on cache miss?

.github/workflows/CD.yml
khanhntd: One more thing to learn since I thought the previous validation tests inherit the ```release-version-image``` so this step would also inherits it too. Fixed it

.github/workflows/CI.yml
khanhntd: Reasonable. Fixed it with newest commit.

.github/workflows/CD.yml
khanhntd: Reasonable. Fixed it with newest commit though.

.github/workflows/CD.yml
khanhntd: Reasonable. Thought everything would make sense to keep it nearest to the building version. However, after considering the visibility and short-condition, make sense to me now.

.github/workflows/CI.yml
khanhntd: It is supposed to support multiple OS, mostly the same as buildx but currently we don't need it so we can remote it.

.github/workflows/CI.yml
khanhntd: These cache fields are used for building an image not from scratch [but building from a cache](https://docs.docker.com/registry/recipes/mirror/). So this is not relating to the cache miss.

khanhntd: Tested it with my CI and the [result was from 1m52 ](https://github.com/khanhntd/aws-otel-collector/runs/4735221080?check_suite_focus=true) to [roughly 50s](https://github.com/khanhntd/aws-otel-collector/runs/4736381891?check_suite_focus=true)

.github/workflows/CD.yml
```diff
@@ -495,72 +486,73 @@ jobs:
...
-
-      - name: Load Image
-        run: docker load < build/packages/$IMAGE_NAME.tar
+          path: |
+            VERSION
+          key: release-latest-image-${{ github.run_id }}
 
-      - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v1
+      - name: Restore cached packages
```
jefchien, line 506: What do you need the cached packages for? This job is just for retagging images isn't it?

.github/workflows/CD.yml
```diff
@@ -495,72 +486,73 @@ jobs:
...
+          username: ${{ secrets.RELEASE_KEY_ID }}
+          password: ${{ secrets.RELEASE_SECRET }}
+        env:
+          AWS_REGION: us-east-1
+
+      - name: Pull image from integration test ECR, tag as latest and push to public release ECR and DockerHub
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.any-update == 'true' }}
+        uses: akhilerm/tag-push-action@v2.0.0
+        with:
+          src: public.ecr.aws/$ECR_REPO_INTEGRATION_TEST:${{ needs.release-checking.outputs.testing_version }}
```
jefchien, line 542:
```suggestion
          src: public.ecr.aws/$ECR_REPO:${{ needs.release-checking.outputs.version }}
```
Still think this should be the one already pushed into the release ECR. They're technically the same image, so it doesn't really matter.

.github/workflows/CD.yml
```diff
@@ -495,72 +486,73 @@ jobs:
...
           username: "${{ secrets.DOCKERHUB_RELEASE_USERNAME }}"
           password: "${{ secrets.DOCKERHUB_RELEASE_TOKEN }}"
 
-      - name: Tag latest and upload to docker
-        if: steps.version.outputs.any-update == 'true'
-        run: |
-          REPO_NAME="$IMAGE_NAMESPACE/$IMAGE_NAME"
-          docker tag $IMAGE_NAME $REPO_NAME:latest
-          docker push $REPO_NAME:latest
+      - name: Login to Public Integration Test ECR
```
jefchien, line 527:
```suggestion
      - name: Login to Public Release ECR
```

.github/workflows/CI.yml
```diff
@@ -493,50 +451,68 @@ jobs:
...
-      - name: upload to ECR
-        if: steps.e2etest-release.outputs.cache-hit != 'true'
-        run: |
-          docker load < build/packages/$IMAGE_NAME.tar
-          docker tag $IMAGE_NAME public.ecr.aws/$ECR_REPO:${{ needs.e2etest-preparation.outputs.version }}
-          docker push public.ecr.aws/$ECR_REPO:${{ needs.e2etest-preparation.outputs.version }}
+          file: cmd/awscollector/Dockerfile
+          context: .
+          push: true
+          tags: public.ecr.aws/${{ env.ECR_REPO }}:${{ needs.e2etest-preparation.outputs.version }}
```
jefchien, line 513:
```suggestion
          tags: public.ecr.aws/$ECR_REPO:${{ needs.e2etest-preparation.outputs.version }}
```
nitpick: For consistency.

.github/workflows/CD.yml
khanhntd: You are right thought. The last commit takes the VERSION from build/packages/VERSION so that's why it needs a ```restore cached packages``` step. However, it is not needed here anymore so forget to delete it.

.github/workflows/CI.yml
khanhntd: Cannot used this since I tested it once and [get an error with it though](https://github.com/khanhntd/aws-otel-collector/runs/4734505542?check_suite_focus=true). (an example is a similar-one but yeah its works the same)

.github/workflows/CD.yml
khanhntd: True thought. Fixed it in the newest commit though.

.github/workflows/CD.yml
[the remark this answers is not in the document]
khanhntd: Yeah. I thought that you mentioned [about the image ](https://github.com/aws-observability/aws-otel-collector/pull/860/files#diff-a64cbb6e953e697cb4302ab31af9ebbd95a8de6216f874bb0ffdefce8e3c1cd9R531-R538) that is tagged as latest so answer the wrong one.

.github/workflows/CI.yml
jefchien: That's interesting. How does it work in the CD then?

khanhntd: [Not worked and having a similar error as expected though](https://github.com/khanhntd/aws-otel-collector/runs/4767860865?check_suite_focus=true). Fixed it with the newest commit

.github/workflows/CD.yml
```diff
@@ -517,50 +495,57 @@ jobs:
...
+          path: |
+            VERSION
+          key: release-latest-image-${{ github.run_id }}
 
       - name: Compare version with Dockerhub latest
         id: version
+        if: steps.release-latest-image.outputs.cache-hit != 'true'
         run: |
-          TAG=`cat build/packages/VERSION`
+          TAG="${{ github.event.inputs.version }}"
```
jefchien, line 510:
```suggestion
          TAG="${{ needs.release-checking.outputs.version }}"
```
Don't use the input version even though it should be the same. That goes for all of the other ones as well. It's better to use variables that we've defined.

.github/workflows/CD.yml
```diff
@@ -517,50 +495,57 @@ jobs:
...
+          dst: |
+            public.ecr.aws/${{ env.ECR_REPO }}:latest
+            $IMAGE_NAMESPACE/${{ env.IMAGE_NAME }}:latest
+
+      - name: Pull image from integration test ECR, tag with input version and push to DockerHub
+        uses: akhilerm/tag-push-action@v2.0.0
+        if: steps.release-latest-image.outputs.cache-hit != 'true'
+        with:
+          src: public.ecr.aws/${{ env.ECR_REPO }}:${{ github.event.inputs.version }}
+          dst: $IMAGE_NAMESPACE/${{ env.IMAGE_NAME }}:${{ github.event.inputs.version }}
```
jefchien, line 545: So `$IMAGE_NAMESPACE` is fine, but `$IMAGE_NAME` is not? That's weird.

.github/workflows/CI.yml
```diff
@@ -219,7 +220,7 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: msi_artifacts
-          path: build/packages
+          path: $PACKAGING_ROOT
```
jefchien, line 223: Does this work? 
>To use the value of an environment variable in a workflow file, you should use the env context. If you want to use the value of an environment variable inside a runner, you can use the runner operating system's normal method for reading environment variables.
>
>If you use the workflow file's run key to read environment variables from within the runner operating system (as shown in the example above), the variable is substituted in the runner operating system after the job is sent to the runner. For other parts of a workflow file, you must use the env context to read environment variables; this is because workflow keys (such as if) require the variable to be substituted during workflow processing before it is sent to the runner.

https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables

.github/workflows/CI.yml
```diff
@@ -493,50 +451,68 @@ jobs:
             VERSION
           key: e2etest-release-${{ github.run_id }}
 
+      - name: restore cached rpms
+        uses: actions/cache@v2
```
jefchien, line 455: Should there be a cache miss check here?

.github/workflows/CI.yml
```diff
@@ -493,50 +451,68 @@ jobs:
...
-      - name: restore cached rpms
-        uses: actions/cache@v2
-        with:
-          key: "cached_packages_${{ github.run_id }}"
-          path: build/packages
-
       - name: Create SSM package
         run: |
-          ssm_pkg_version=`cat build/packages/VERSION`
+          ssm_pkg_version=`cat $PACKAGING_ROOT/VERSION`
```
jefchien, line 490: Should there also be a cache miss check here?

khanhntd: IMO its not a good idea since we want to always create a SSM for later destroy and [you have filed a PR for this before ](https://github.com/aws-observability/aws-otel-collector/pull/758/files).

.github/workflows/CI.yml
khanhntd: IMO its not a good idea with the same reasons as cache for SSM package.

.github/workflows/CD.yml
khanhntd: Reasonable. Fixed it with the newest commit. Thought we can reduce the needs step and inputs step if we used the ```${{github.event.inputs.version}}```

.github/workflows/CD.yml
khanhntd: It's not. My mistake for not checking carefully. Fixed it with the newest commit.

.github/workflows/CI.yml
khanhntd: Reasonable to me. That's why some of the actions are not able to use the $VARIABLE convention though. Fixed it with the newest commit.

.github/workflows/CI.yml
jefchien: Haha you're correct. Completely forgot about that.

jefchien (approved):
LGTM

PR #861, opened 2022-01-06
add for all dev branch

khanhntd:

For all of the branch that has the prefix "dev", it would trigger the CI. In this way, we can test the CI.

An alternative solution would be using act https://github.com/nektos/act

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ on:
     branches:
       - main
       - release/v*
-      - dev
+      - dev/*
     paths-ignore:
       - '.github/**'
       - '!.github/workflows/CI.yml'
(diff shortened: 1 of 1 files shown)


sethAmazon (approved):

jefchien (approved):

.github/workflows/CI.yml
```diff
@@ -17,7 +17,7 @@ on:
     branches:
       - main
       - release/v*
-      - dev
+      - dev/*
```
jefchien, line 20: We still have a `dev` branch even if it isn't really in use. Should we still have the CI run on it? This change will remove that.

khanhntd: There are two ways for us to change it: one is to rename the branch as dev/ and second is that change the prefix pattern to dev*.

jefchien: The other way is to just have two entries
```
- dev
- dev/*
```

khanhntd: Reasonable. Fixed it.

jefchien (approved):

Issue #766, opened 2021-11-23
ARM build and dependencies


nrooney:
A team I've been working with has reported that the ARM build is broken, as the dependancies were not building. They rewrote the docker file to get around the issue. Apologies for lack of insight here as I haven't used the tool myself; I'll get some more info tomorrow and report back in this issue.

rakyll:
Related to https://github.com/aws-observability/aws-otel-collector/issues/536.

sethAmazon:
I do not think we support arm64 docker images. @alolita

paya-cz:
There is no arm64 support, at least yet.

Compare:
https://hub.docker.com/r/amazon/aws-otel-collector/tags
With:
https://hub.docker.com/r/amazon/aws-xray-daemon/tags

Disappointing.

alolita:
Hi @paya-cz @nrooney we'll take a look at adding support for arm64 docker images for ADOT. Can you share what your pipeline is - this may be useful for our testing. Thx.

paya-cz:
@alolita Thank you, being able to use ADOT with Graviton2 would certainly be nice.

Please elaborate what you mean by "pipeline".

alolita:
Hi @paya-cz could you specify the pipeline you plan to use these arm64 images for? 

i.e. specify your setup - data source(s) --> receiver (otlp or prometheus) --> exporter (otlp or custom) --> monitoring backend

paya-cz:
@alolita I fear my pipeline is rather uninspiring.

I just follow this, to get traces from nodejs ECS container to show up in AWS XRay:
https://aws-otel.github.io/docs/getting-started/js-sdk/trace-manual-instr

alolita:
Thanks @paya-cz we'll take a look at adding this support.

sethAmazon:
Write up arm64 docker images

Goal

* Be able to build both adm64 and arm64 docker images

How

* Use buildx ability to build multi-platform docker images 
* Split out into 3 buildx commands
    * Why? 
        * You can only load 1 platform image outside of a registry
    * 1 amd64 in make file
    * 1 arm64 in make file
    * 1 build and push both amd64 and arm64 to ecr in ci
        * This command will be run in the ci and require a load
* If statement in dockerfile to decide to build amd64 image or arm64 image based on platform input 
* Make sure we are using a buildx by default supported version of docker
    * Some earlier versions require buildx to be create and not available by default

Aneurysm9:
[...]

PR #870, opened 2022-01-10
Support buildx for amd64 and arm64 builds (#766)

sethAmazon:
**Description:** 
Support Buildx for amd64 and arm64 builds 

**Link to tracking Issue:** 
#766

**Testing:**
(Make creates binaries) 
make docker-build
make docker-build-arm

(With binary not created -takes a very long time since it requires go mod download-)
docker buildx build --platform linux/amd64 -t amazon/awscollector:v0.15.0 -f ./cmd/awscollector/Dockerfile .
docker buildx build --platform linux/arm64 -t amazon/awscollector:v0.15.0 -f ./cmd/awscollector/Dockerfile .

(defaults to amd64)
docker build -t amazon/awscollector:latest -f ./cmd/awscollector/Dockerfile .

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -464,10 +464,6 @@ jobs:
           key: "cached_binaries_${{ github.run_id }}"
           path: build
 
-      - name: Set up Docker Buildx
-        if: steps.e2etest-release.outputs.cache-hit != 'true'
-        uses: docker/setup-buildx-action@v1
-
       - name: Login to Public Integration Test ECR
         if: steps.e2etest-release.outputs.cache-hit != 'true'
         uses: docker/login-action@v1
@@ -499,6 +495,20 @@ jobs:
         if: steps.e2etest-release.outputs.cache-hit != 'true'
         run: s3_bucket_name=aws-otel-collector-test upload_to_latest=0 bash tools/release/image-binary-release/s3-release.sh
 
+      - name: Set up Docker Buildx
+        if: steps.e2etest-release.outputs.cache-hit != 'true'
+        uses: docker/setup-buildx-action@v1
+
+      - name: Set up QEMU
+        if: steps.e2etest-release.outputs.cache-hit != 'true'
+        uses: docker/setup-qemu-action@v1
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: public.ecr.aws/${{ env.ECR_REPO }}
+
       #Build the adot collector image for two primary reasons:
       #-Using the adot collector image to do the integration test
       #-Export it for delivery version image in CD
@@ -515,8 +525,8 @@ jobs:
             public.ecr.aws/${{ env.ECR_REPO }}:latest
           cache-from: type=registry
           cache-to: type=inline
-
-
+          platforms : linux/amd64, linux/arm64
+          labels: ${{ steps.meta.outputs.labels }}
 
   get-testing-suites:
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,10 @@ build: install-tools lint multimod-verify
 amd64-build: install-tools lint multimod-verify
 	GOOS=linux GOARCH=amd64 $(GOBUILD) $(LDFLAGS) -o ./build/linux/aoc_linux_x86_64 ./cmd/awscollector
 
+.PHONY: arm64-build
+arm64-build: install-tools lint multimod-verify
+	GOOS=linux GOARCH=arm64 $(GOBUILD) $(LDFLAGS) -o ./build/linux/aoc_linux_aarch64 ./cmd/awscollector
+
 # For building container image during development, no lint nor other platforms
 .PHONY: amd64-build-only
 amd64-build-only:
@@ -107,7 +111,11 @@ package-deb: build
 
 .PHONY: docker-build
 docker-build: amd64-build
-	docker build -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
+	docker buildx build --platform linux/amd64 -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
+
+.PHONY: docker-build-arm
+docker-build-arm: arm64-build
+	docker buildx build --platform linux/arm64 -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
 
 .PHONY: docker-push
 docker-push:
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -4,42 +4,19 @@
 ################################
 FROM alpine:latest as build 
 
-ENV GOPROXY=direct
-
 # update cert
 RUN apk --update add ca-certificates
 
-# install libs
-RUN apk add --update bash git make build-base
-
-# config go
-COPY --from=golang:1.17-alpine /usr/local/go/ /usr/local/go/
-ENV PATH="/usr/local/go/bin:${PATH}"
-
 WORKDIR /workspace
 
-# copy artifacts
-COPY cmd  /workspace/cmd
-COPY config.yaml /workspace/config.yaml
-COPY pkg /workspace/pkg
-COPY tools /workspace/tools
-COPY VERSION /workspace/VERSION
-COPY .git /workspace/.git
-COPY Makefile /workspace/Makefile
-COPY config /workspace/config
+ARG TARGETPLATFORM
+RUN echo "Targate platform is $TARGETPLATFORM"
 
-#Copy binary if exist else build binary
-COPY go.* build/linux/aoc_linux_x86_6* /workspace/
-RUN if [[ -f /workspace/aoc_linux_x86_64 ]] ; \
-    then \
-        echo Copying binary \
-        && mkdir -p /workspace/build/linux \
-        && cp /workspace/aoc_linux_x86_64 /workspace/build/linux/aoc_linux_x86_64 ; \
-    else \
-        echo Building binary \
-        && go mod download  \
-        && make amd64-build ; \
-    fi
+# copy artifacts
+# always assume binary is created
+COPY build/linux/* /workspace/
+COPY cmd/awscollector/copy_docker_binary.sh  /workspace/cmd/awscollector/copy_docker_binary.sh
+RUN cmd/awscollector/copy_docker_binary.sh
 
 ################################
 #	Final Stage            #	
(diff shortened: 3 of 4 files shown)


jefchien:
Wouldn't it be simpler to add arm64 support after https://github.com/aws-observability/aws-otel-collector/pull/860 gets merged in? Could just add it to the platforms https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md

sethAmazon:
Merge https://github.com/aws-observability/aws-otel-test-framework/pull/480 first

.github/workflows/CI.yml
```diff
@@ -360,7 +360,7 @@ jobs:
 
       - name: Build Image
         if: steps.cached_image.outputs.cache-hit != 'true'
-        run: docker build -t $IMAGE_NAME -f cmd/awscollector/Dockerfile .
+        run: docker buildx build --platform linux/amd64 -t $IMAGE_NAME -f cmd/awscollector/Dockerfile .
```
khanhntd, line 363: Same idea as Jeffrey. Would it be good to pass in after using build-and-push action PR?

cmd/awscollector/Dockerfile
```diff
@@ -28,17 +28,36 @@ COPY .git /workspace/.git
...
 #Copy binary if exist else build binary
-COPY go.* build/linux/aoc_linux_x86_6* /workspace/
-RUN if [[ -f /workspace/aoc_linux_x86_64 ]] ; \
+COPY go.* build/linux/* /workspace/
+RUN if [[ "$TARGETPLATFORM" == "linux/amd64" ]] ; \
     then \
-        echo Copying binary \
-        && mkdir -p /workspace/build/linux \
-        && cp /workspace/aoc_linux_x86_64 /workspace/build/linux/aoc_linux_x86_64 ; \
+        if [[ -f /workspace/aoc_linux_x86_64 ]] ; \
```
khanhntd, line 38: Should be cleaner when using variables instead of hard code? Such as aoc_build_platform = aoc_linux_x86_64 if the target platform is linux=am64.

.github/workflows/CI.yml
sethAmazon: This line does not need to change I can revert

cmd/awscollector/build_docker_binary.sh
```diff
@@ -0,0 +1,17 @@
...
+fi
+echo binary name is ${BINARY}
+if [[ -f /workspace/${BINARY} ]]; then
+  echo Copying binary &&
+    mkdir -p /workspace/build/linux &&
+    cp /workspace/${BINARY} /workspace/build/linux/aoc_linux
+else
+  echo Building binary &&
+    go mod download &&
+    make amd64-build &&
```
khanhntd, line 14: So the assumption we make here is if one of the binaries do not exist, we will build all the binaries right?

sethAmazon: Yes

cmd/awscollector/Dockerfile
```diff
@@ -28,18 +28,12 @@ COPY .git /workspace/.git
...
-        echo Copying binary \
-        && mkdir -p /workspace/build/linux \
-        && cp /workspace/aoc_linux_x86_64 /workspace/build/linux/aoc_linux_x86_64 ; \
-    else \
-        echo Building binary \
-        && go mod download  \
-        && make amd64-build ; \
-    fi
+COPY go.* build/linux/* /workspace/
+RUN cmd/awscollector/build_docker_binary.sh
```
Aneurysm9, line 36: It isn't necessary to build the executable  again, is it?  The `docker-build` and `docker-build-arm` make targets already require the `amd64-build` and `arm64-build` targets, respectively.  We should simply copy the already-built binary into the final stage image.

This should also ensure that any packages we build from the binaries built from `a(md|rm)64-build`, such as RPM or DEP packages, are using the same binary as the container image.

Makefile
```diff
@@ -107,7 +111,13 @@ package-deb: build
 
 .PHONY: docker-build
 docker-build: amd64-build
-	docker build -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
+	chmod +x cmd/awscollector/build_docker_binary.sh
+	docker buildx build --platform linux/amd64 -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
+
+.PHONY: docker-build-arm
+docker-build-arm: arm64-build
+	chmod +x cmd/awscollector/build_docker_binary.sh
```
jefchien, line 119: Why not just include the chmod step inside the Dockerfile if it's required for each run?

cmd/awscollector/Dockerfile
sethAmazon: The idea was to keep backward compatibility where if we do not have the binary we build it. People can use docker commands not through the makefile if they want. Why are they doing this, not sure.

sethAmazon: What would be your recommendation? Just remove the ability to build the binary inside of the dockerfile if no binary is found?

Makefile
sethAmazon: done

cmd/awscollector/Dockerfile
```diff
@@ -4,42 +4,24 @@
 ################################
 FROM alpine:latest as build 
 
-ENV GOPROXY=direct
```
sethAmazon, line 7: After speaking with @Aneurysm9 we should not build the binary inside the docker container since make will build the binary and in ci we cache the binary.

cmd/awscollector/build_docker_binary.sh
sethAmazon: After speaking with @Aneurysm9 we should not build the binary inside the docker container since make will build the binary and in ci we cache the binary.

cmd/awscollector/Dockerfile
sethAmazon: After speaking with @Aneurysm9 we should not build the binary inside the docker container since make will build the binary and in ci we cache the binary.

cmd/awscollector/Dockerfile
```diff
@@ -4,42 +4,22 @@
...
-        echo Copying binary \
-        && mkdir -p /workspace/build/linux \
-        && cp /workspace/aoc_linux_x86_64 /workspace/build/linux/aoc_linux_x86_64 ; \
-    else \
-        echo Building binary \
-        && go mod download  \
-        && make amd64-build ; \
-    fi
+#Always assume binary is created
+COPY go.* build/linux/* /workspace/
```
jefchien, line 20: What are the `go.*` files used for in this case?

cmd/awscollector/Dockerfile
```diff
@@ -4,42 +4,22 @@
...
-RUN apk add --update bash git make build-base
-
-# config go
-COPY --from=golang:1.17-alpine /usr/local/go/ /usr/local/go/
-ENV PATH="/usr/local/go/bin:${PATH}"
-
 WORKDIR /workspace
 
 # copy artifacts
 COPY cmd  /workspace/cmd
```
jefchien, line 13: Is this just to copy the script? If so, make it more explicit.
```suggestion
COPY cmd/awscollector/copy_docker_binary.sh  /workspace/
```

cmd/awscollector/Dockerfile
anuraaga: We were previously only building in the container when there was no prebuilt binary available, which isn't the case in the CI scripts. I think having a Dockerfile that can build without any external tooling / commands is quite useful, for example when developing I can just put a repo URL into a docker-compose file and things just work. A separate Dockerfile could work for this, but it seems like we can seamlessly support both cases with the previous approach.

khanhntd: I would prefer @anuraaga 's approach since it would increase independence between files. However, if we planned used this Dockerfile only for the CI and make, then writing a notification with our customer about not supporting these would be a recommended way though.

sethAmazon: I'm good either way but I think I'll go with a separate dockerfile that builds the binary since I like how lean our current docker file is.

sethAmazon: Resolved

cmd/awscollector/binaryBuild.Dockerfile
```diff
@@ -0,0 +1,73 @@
+################################
+#	Building Stage         #
+#			       #
+################################
+FROM alpine:latest as build
+
+ENV GOPROXY=direct
```
sethAmazon, line 7: This will take over 1hr to run since no go proxy is used.

.github/workflows/CI.yml
```diff
@@ -485,6 +485,9 @@ jobs:
           aws-secret-access-key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
           aws-region: us-west-2
 
+      - name: Set up QEMU
```
jefchien, line 488: Should probably only run on cache miss. Also, nitpick, but maybe keep the buildx and qemu set up steps together. Maybe move them both to right before the build-push action.

sethAmazon, line 488: required for multi arc according to docs

.github/workflows/CI.yml
```diff
@@ -515,6 +518,8 @@ jobs:
             public.ecr.aws/${{ env.ECR_REPO }}:latest
           cache-from: type=registry
           cache-to: type=inline
+          platforms : linux/amd64, linux/arm64
+          build-args: ${{ needs.e2etest-preparation.outputs.version }}
```
jefchien, line 522:
```suggestion
          build-args: VERSION=${{ needs.e2etest-preparation.outputs.version }}
```
Doesn't this need to know what argument to assign itself to?

sethAmazon:
Screen Shot 2022-01-11 at 1 37 48 PM

cmd/awscollector/Dockerfile
```diff
@@ -4,63 +4,45 @@
 ################################
```
khanhntd, line 4: Any reasons to separate the Dockerfile and binaryBuild.Dockerfile thought? Also would be better to [update the doc ](https://github.com/aws-observability/aws-otel-collector/blob/main/docs/developers/build-docker.md)and notify we have supported bold amd64 and arm64

.github/workflows/CI.yml
khanhntd, line 522: Mind if I ask what we need this flag thought?

cmd/awscollector/Dockerfile
```diff
@@ -4,63 +4,45 @@
...
+RUN cmd/awscollector/copy_docker_binary.sh
 
 ################################
 #	Final Stage            #	
 #			       #	
 ################################
 FROM scratch
 
+ARG VERSION=unknown
+LABEL VERSION=$VERSION
```
jefchien, line 31: It looks like labels are typically lowercase.
>Label keys should begin and end with a lower-case letter and should only contain lower-case alphanumeric characters, the period character (.), and the hyphen character (-). Consecutive periods or hyphens are not allowed.

https://docs.docker.com/config/labels-custom-metadata/#key-format-recommendations

khanhntd: Nice catch.

cmd/awscollector/Dockerfile
sethAmazon: The reason being that we have a docker file when we build the binary (we do not advise to use since it take a long time to build). The other docker file will copy the binary (should be used by the makefile and the ci).

.github/workflows/CI.yml
sethAmazon: From the slack chat earlier. It is a way adding a label called version with the version of the collector.

sethAmazon: docker inspect {image}
under labels it will give the adot version. This is good for people that use latest tag instead of a version for their tag.

khanhntd: Reasonable.

cmd/awscollector/Dockerfile
```diff
@@ -4,63 +4,45 @@
...
+RUN cmd/awscollector/copy_docker_binary.sh
 
 ################################
 #	Final Stage            #	
 #			       #	
 ################################
 FROM scratch
 
+ARG VERSION=unknown
+LABEL version=$VERSION
```
Aneurysm9, line 31: This should use the [OCI standard labels](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys).  While we're at it we should see if there are any others that can easily be populated.

anuraaga: We can use docker-metadata action to populate some standard labels automatically

https://github.com/docker/metadata-action

sethAmazon: From our chat with @anuraaga I will use docker metadata

cmd/awscollector/copy_docker_binary.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+BINARY=aoc_linux_x86_64
```
anuraaga, line 2: Small nit I would probably output the same `aoc_linux` binary into a folder with the platform, just to reduce moving parts (renaming files).

Even better could be just `awscollector` perhaps since that's what we end up renaming to anyways. Results in less names to deal with.

anuraaga (approved):

Aneurysm9 (approved):

cmd/awscollector/Dockerfile
```diff
@@ -4,42 +4,20 @@
...
-    else \
-        echo Building binary \
-        && go mod download  \
-        && make amd64-build ; \
-    fi
+# copy artifacts
+# always assume binary is created
+COPY build/linux/* /workspace/
+COPY cmd/awscollector/copy_docker_binary.sh  /workspace/cmd/awscollector/copy_docker_binary.sh
+RUN chmod +x cmd/awscollector/copy_docker_binary.sh
```
anuraaga, line 19: You should be able to check in the script with permissions set

cmd/awscollector/Dockerfile
```diff
@@ -48,19 +26,19 @@ RUN if [[ -f /workspace/aoc_linux_x86_64 ]] ; \
...
-COPY --from=build /workspace/config/eks/prometheus/config-memcached.yaml /etc/eks/prometheus/config-memcached.yaml
-COPY --from=build /workspace/config/eks/prometheus/config-nginx.yaml /etc/eks/prometheus/config-nginx.yaml
-COPY --from=build /workspace/config/eks/prometheus/config-redis.yaml /etc/eks/prometheus/config-redis.yaml
-COPY --from=build /workspace/config/eks/prometheus/config-all.yaml /etc/eks/prometheus/config-all.yaml
-COPY --from=build /workspace/config/ecs/container-insights/otel-task-metrics-config.yaml /etc/ecs/container-insights/otel-task-metrics-config.yaml
-COPY --from=build /workspace/config/ecs/ecs-*.yaml /etc/ecs/
-COPY --from=build /workspace/config/ecs/otel-instance-metrics-config.yaml /etc/ecs/otel-instance-metrics-config.yaml
-COPY --from=build /workspace/config/apprunner/apprunner-default-config.yaml /etc/apprunner/apprunner-default-config.yaml
+COPY --from=build /workspace/build/linux/aoc_linux /awscollector
+COPY config.yaml /etc/otel-config.yaml
```
anuraaga, line 30: In a future PR it would be good to redo the folder structure in this repo to allow just one glob COPY command to copy in all the configs to reduce this maintenance

cmd/awscollector/copy_docker_binary.sh
sethAmazon: Yeah that was for backward compatibility in the make file. Any sort of backwards not compatible change we should have a meeting just to discuss that.  Or I can just make more make commands.

cmd/awscollector/Dockerfile
sethAmazon: It seems I can add execute permission in git. I'm going to look into if this line is still needed.

cmd/awscollector/Dockerfile
sethAmazon: Will create an issue for this

cmd/awscollector/copy_docker_binary.sh
sethAmazon: https://github.com/aws-observability/aws-otel-collector/issues/880

cmd/awscollector/Dockerfile
sethAmazon: https://github.com/aws-observability/aws-otel-collector/issues/879

jefchien (approved):
Do you need to change the CD as well or will retagging pull all platforms?

khanhntd:
> Do you need to change the CD as well or will retagging pull all platforms?

Will need to copy multiple platforms since the local image cache can [only store a single architecture per image.](https://github.com/docker/for-mac/issues/5625#issuecomment-971716379)

sethAmazon:
@khanhntd it seems you use https://github.com/akhilerm/tag-push-action this should support multi platform out of box according to their docs

PR #874, opened 2022-01-11
Fix condition for ssm

khanhntd:

The condition for deleting SSM would be if we able to find the SSM packages or if the cache packages exist. However, such condition won't be  met if we cancel the workflow earlier than the ```e2e-preparation``` (we won't have the cache packages if this job and previous needs jobs do not succeed) or if we got some errors during creating SSM package (we won't have SSM in our aws resources). Therefore, such conditions are need to avoid have errors at this always ```clean-ssm-package``` job.


sethAmazon (approved):

.github/workflows/CI.yml
```diff
@@ -1061,12 +1060,16 @@ jobs:
...
           key: "cached_packages_${{ github.run_id }}"
 
       - name: clean up SSM test package
+        if: steps.cache_packages.outputs.cache-hit == 'true'
         run: |
-          ssm_pkg_version=`cat $PACKAGING_ROOT/VERSION`
-          aws ssm delete-document --name "testAWSDistroOTel-Collector" --version-name ${ssm_pkg_version}
+          ssm_pkg_version=`cat build/packages/VERSION`
+          aws ssm describe-document --name ${SSM_PACKAGE_NAME} --version-name ${ssm_pkg_version} >/dev/null 2>&1 && \
+            aws ssm delete-document --name ${SSM_PACKAGE_NAME} --version-name ${ssm_pkg_version}
```
aateeqi, line 1074: so this is probably my lack of yml knowledge but why do we have to write the command twice?

khanhntd: We only need to delete (second command) if the ```ssm package name``` exist (first command).

aateeqi: oh right, sorry i misread them both as describe

aateeqi (approved):

PR #875, opened 2022-01-11
Change input sha for testing version

khanhntd:

When we open a PR, the similar behavior is found across the PR. Therefore, we don't need to use the RUN_ID to tag the image but instead, using the COMMIT_ID (or COMMIT SHA). If we did it this way, it will give us two advantages:
* Increase more visibility to the customer and also for the dev  since we publish CI image public.
* Avoid having similar layers image and also wasting name. In this way, it would show us only the final success image.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -379,8 +379,9 @@ jobs:
         if: steps.cached_ssm.outputs.cache-hit != 'true'
         run: |
           # use a version with github run id as same as e2etest-preparation
+          ssm_pkg_version="`cat VERSION`-$GITHUB_SHA"
           rm -rf $PACKAGING_ROOT/ssm
-          python3 tools/ssm/ssm_manifest.py "`cat VERSION`-${{ github.run_id }}"
+          python3 tools/ssm/ssm_manifest.py ${ssm_pkg_version}
 
   e2etest-preparation:
     runs-on: ubuntu-latest
@@ -432,7 +433,7 @@ jobs:
         id: versioning
         run: |
           # build a version with github run id so that we can distingush each build for integ-test
-          Version="`cat VERSION`-$GITHUB_RUN_ID"
+          Version="`cat VERSION`-$GITHUB_SHA"
           echo $Version > $PACKAGING_ROOT/VERSION
           cat $PACKAGING_ROOT/VERSION
           echo "::set-output name=version::$Version"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -375,10 +375,7 @@ jobs:
 
       - run: ls -R
 
-        # Build the ssm package and manifest by using a version with github short sha as same as e2etest-preparationfor the following reasons:
-        # - Distingush each build for Integration test
-        # - Increase more visibility to the customer and also for the dev since we publish the image for integration test.
-        # - Avoid having similar layers image and only have a final result image.
+        # Build the ssm package and manifest by using a version with github short sha as same as e2etest-preparation.
       - name: Create zip file and manifest for SSM package
         if: steps.cached_ssm.outputs.cache-hit != 'true'
         run: |
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -439,10 +439,10 @@ jobs:
       - name: Versioning for testing
         id: versioning
         run: |
-          Version="$(cat VERSION)-$(git rev-parse --short HEAD)"
-          echo $Version > $PACKAGING_ROOT/VERSION
+          version="$(cat VERSION)-$(git rev-parse --short HEAD)"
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -432,7 +433,7 @@ jobs:
         id: versioning
         run: |
           # build a version with github run id so that we can distingush each build for integ-test
-          Version="`cat VERSION`-$GITHUB_RUN_ID"
+          Version="`cat VERSION`-$GITHUB_SHA"
```
jefchien, line 436: Would rather not have the whole SHA. They can get rather long.
```suggestion
          Version="$(cat VERSION)-$(git rev-parse --short HEAD)"
```

README.md
```diff
@@ -95,7 +95,8 @@ The performance test can be conducted by following the [instructions](https://gi
 
 ### Support
 
-For each merged pull request, a corresponding image is pushed to [public.ecr.aws/aws-otel-test/adot-collector-integration-test](https://gallery.ecr.aws/aws-otel-test/adot-collector-integration-test). 
+For each merged pull request, a corresponding image in the naming convention: ```ADOT_COLLECTOR_VERSION - GITHUB_SHA```
```
jefchien, line 98:
```suggestion
For each merged pull request, a corresponding image with the naming convention of `[ADOT_COLLECTOR_VERSION]-[GITHUB_SHA]```
```

jefchien:
Are we going to remove the existing images that don't follow this format?

.github/workflows/CI.yml
khanhntd: Interesting. Nice catch for getting the short SHA though.

README.md
khanhntd: Fixed it with the newest commit.

khanhntd:
> Are we going to remove the existing images that don't follow this format?

IMO, we should since it would avoid confusion for the previous format though and in assumption, no customers have been used the previous image before except the nightly build team.

.github/workflows/CI.yml
```diff
@@ -432,7 +433,7 @@ jobs:
         id: versioning
         run: |
           # build a version with github run id so that we can distingush each build for integ-test
-          Version="`cat VERSION`-$GITHUB_RUN_ID"
+          Version="$(cat VERSION)-$(git rev-parse --short HEAD)"
```
jefchien, line 442:
```suggestion
          version="$(cat VERSION)-$(git rev-parse --short HEAD)"
```
nitpick: Local variable names should be lower snake case. https://google.github.io/styleguide/shellguide.html#naming-conventions

README.md
```diff
@@ -95,7 +95,8 @@ The performance test can be conducted by following the [instructions](https://gi
 
 ### Support
 
-For each merged pull request, a corresponding image is pushed to [public.ecr.aws/aws-otel-test/adot-collector-integration-test](https://gallery.ecr.aws/aws-otel-test/adot-collector-integration-test). 
+For each merged pull request, a corresponding image with the naming convention of:
```
jefchien, line 98: nitpick: Don't need the colon.

khanhntd: Fixed it in the newest commit.

.github/workflows/CI.yml
khanhntd: The version output is using version but either way, I'm good with the nitpick lol

aateeqi (approved):

.github/workflows/CI.yml
```diff
@@ -375,12 +375,16 @@ jobs:
 
       - run: ls -R
 
+        # Build the ssm package and manifest by using a version with github short sha as same as e2etest-preparationfor the following reasons:
+        # - Distingush each build for Integration test
+        # - Increase more visibility to the customer and also for the dev since we publish the image for integration test.
+        # - Avoid having similar layers image and only have a final result image.
```
jefchien, line 381: These reasons don't really apply for the SSM package. I don't think we give customers access and it isn't an image.

.github/workflows/CI.yml
```diff
@@ -375,12 +375,16 @@ jobs:
 
       - run: ls -R
 
+        # Build the ssm package and manifest by using a version with github short sha as same as e2etest-preparationfor the following reasons:
```
jefchien, line 378:
```suggestion
        # Build the ssm package and manifest by using a version with github short sha same as e2etest-preparation for the following reasons:
```

.github/workflows/CI.yml
khanhntd: True thought. Reverted to previous comment but add some extra things.

.github/workflows/CI.yml
khanhntd: Reasonable. Fixed it in the newest commit.

jefchien (approved):

PR #881, opened 2022-01-12
Fix typos

kzys:
- Amazon Linux 2 is not "AWS Linux"
- We shouldn't put AWS before other brands such as Debian.

Signed-off-by: Kazuyoshi Kato <katokazu@amazon.com>

Fixed some typos around brand names.

NA

NA

NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ AWS OTel Collector supports all AWS computing platforms and Docker/Kubernetes. H
 * [Run it with Docker](docs/developers/docker-demo.md)
 * [Run it with ECS](docs/developers/ecs-demo.md)
 * [Run it with EKS](docs/developers/eks-demo.md)
-* [Run it on AWS Linux EC2](docs/developers/linux-rpm-demo.md)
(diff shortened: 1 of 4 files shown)


khanhntd (approved):
Nice catch on the typo. IMO, we should specify what Cloud Platform are being used for Amazon Linux since Amazon Linux is only supported on AWS. For GCP, I heard GCP Migration has the [support to migrate from Amazon Linux ](https://cloud.google.com/migrate/compute-engine/docs/4.8/reference/supported-os-versions) but that is not what our customer's intention. For Azure it does not support the OS.

kzys:
How about using "AWS EC2 (OS NAME)"? Also can we remove Beta from the titles?

khanhntd:
> How about using "AWS EC2 (OS NAME)"? Also how about changing "AWSOTelCollector Beta" to "AWSOTelCollector"?

Totally agree with both of these changes. For the latter changes, would even more agree since the s3 binaries/ artifact links in these OS Doc are release links.

Issue #879, opened 2022-01-12
Dockerfile mass copy config


sethAmazon:
**Is your feature request related to a problem? 
Dockerfile mass copy config. Currently we copy each config file separately in our docker file

**Describe the solution you'd like**
Use something like COPY config/* config/*

**Describe alternatives you've considered**
Keep docker file copy as is.

PR #885, opened 2022-01-13
Simplify Dockerfile copy.

jefchien:
**Description:** The file structure of the config directory already matches what's in the image.

**Link to tracking Issue:** https://github.com/aws-observability/aws-otel-collector/issues/879

**Testing:** Built the original/new image and used [dive](https://github.com/wagoodman/dive) to compare the files.

Existing image: public.ecr.aws/aws-observability/aws-otel-collector:latest
image
Newly built:
image

### DIFF
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -10,13 +10,15 @@ RUN apk --update add ca-certificates
 WORKDIR /workspace
 
 ARG TARGETPLATFORM
-RUN echo "Targate platform is $TARGETPLATFORM"
+RUN echo "Target platform is $TARGETPLATFORM"
 
 # copy artifacts
 # always assume binary is created
 COPY build/linux/* /workspace/
-COPY cmd/awscollector/copy_docker_binary.sh  /workspace/cmd/awscollector/copy_docker_binary.sh
-RUN cmd/awscollector/copy_docker_binary.sh
+COPY config.yaml /workspace/config/otel-config.yaml
+COPY config/ /workspace/config/
+COPY cmd/awscollector/copy_docker_binary.sh /workspace/copy_binary.sh
+RUN ./copy_binary.sh
 
 ################################
 #	Final Stage            #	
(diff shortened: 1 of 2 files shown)


cmd/awscollector/Dockerfile
```diff
@@ -10,12 +10,14 @@ RUN apk --update add ca-certificates
...
 ARG TARGETPLATFORM
-RUN echo "Targate platform is $TARGETPLATFORM"
+RUN echo "Target platform is $TARGETPLATFORM"
 
 # copy artifacts
 # always assume binary is created
 COPY build/linux/* /workspace/
-COPY cmd/awscollector/copy_docker_binary.sh  /workspace/cmd/awscollector/copy_docker_binary.sh
+COPY config.yaml /workspace/config/otel-config.yaml
+COPY config/ /workspace/config/
```
khanhntd, line 19: Any reason why we should not copy directly but instead copy twice?

jefchien: It's mainly to reduce the number of COPY layers in the actual image to just one.

khanhntd:
Nitpick: change to exist  with [this line](https://github.com/aws-observability/aws-otel-collector/blob/main/cmd/awscollector/copy_docker_binary.sh#L9)

khanhntd:
Another nitpick: change to finished copying binary with [this line ](https://github.com/aws-observability/aws-otel-collector/blob/main/cmd/awscollector/copy_docker_binary.sh#L12) when you are at it.

cmd/awscollector/Dockerfile
```diff
@@ -10,12 +10,14 @@ RUN apk --update add ca-certificates
...
+COPY cmd/awscollector/copy_docker_binary.sh /workspace/cmd/awscollector/copy_docker_binary.sh
```
khanhntd, line 20: IMO we don't need this line since we don't plan to exec from inside containers anymore.

jefchien: We need this to copy the correct binary into the image.

cmd/awscollector/Dockerfile
khanhntd: Reasonable.

cmd/awscollector/Dockerfile
khanhntd: True but we execute the script not inside the container so I don't see any reasons to copy the script to the containers though.

sethAmazon: The container will not include the script. Only the build container. Only after the FROM scratch will include inside the container

cmd/awscollector/Dockerfile
```diff
@@ -26,18 +28,7 @@ FROM scratch
 
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=build /workspace/build/linux/aoc_linux /awscollector
-COPY config.yaml /etc/otel-config.yaml
```
sethAmazon, line 29: Does this map 1 to 1? Let us hope we don't break anything.

cmd/awscollector/copy_docker_binary.sh
```diff
@@ -1,13 +1,26 @@
 #!/bin/sh
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sethAmazon, line 3: Good catch

cmd/awscollector/Dockerfile
```diff
@@ -10,12 +10,14 @@ RUN apk --update add ca-certificates
 WORKDIR /workspace
 
 ARG TARGETPLATFORM
-RUN echo "Targate platform is $TARGETPLATFORM"
+RUN echo "Target platform is $TARGETPLATFORM"
```
sethAmazon, line 13: Good catch

sethAmazon (approved):

cmd/awscollector/Dockerfile
jefchien: It does. I did a comparison with dive.
image

cmd/awscollector/Dockerfile
khanhntd: True though. But I am still not seeing any reasons why we should copy this file to the ```build container``` if we do not plan to exec inside the ```build container```

Aneurysm9: it is run inside the build container on the next line.

jefchien: Changed the path it copies the script to to `copy_binary.sh` for clarity.

khanhntd: Though we could execute script outside the container. However, it does not. Thanks guys for clarifying though.

anuraaga (approved):

Issue #880, opened 2022-01-12
Naming convention for binary


sethAmazon:
**Is your feature request related to a problem? 
We have to hardcode binary paths in dockerfile

**Describe the solution you'd like**
Change the names of binary paths in make create commands. The linux binary will chosen by path not name. Ex linux/amd64/aoc instead of aoc_linux_x86_64 

**Describe alternatives you've considered**
Create more make commands so as to not change anything.

anuraaga:
We definitely don't want to change the output file within a published archive, e.g. a docker image. But just changing names of intermediate artifacts from Make commands is acceptable IMO but let's confirm together.

PR #890, opened 2022-01-14
Change build path for linux binaries to match GOARCH.

jefchien:
**Description:** Simplifies the docker build process by changing the linux binary build paths to match GOARCH/TARGETARCH.

**Link to tracking Issue:** https://github.com/aws-observability/aws-otel-collector/issues/880

**Testing:** Ran `create_rpm.sh` and `create_deb.sh` (partially). Ran `make docker-build` and `make docker-build-arm`.


Makefile
```diff
@@ -78,22 +78,22 @@ all-srcs:
 .PHONY: build
 build: install-tools lint multimod-verify
 	GOOS=darwin GOARCH=amd64 $(GOBUILD) $(LDFLAGS) -o ./build/darwin/aoc_darwin_amd64 ./cmd/awscollector
-	GOOS=linux GOARCH=amd64 $(GOBUILD) $(LDFLAGS) -o ./build/linux/aoc_linux_x86_64 ./cmd/awscollector
```
sethAmazon, line 81: Have we talked with @alolita  @anuraaga @Aneurysm9 that being backwards not compatible is okay here

cmd/awscollector/Dockerfile
```diff
@@ -9,16 +9,13 @@ RUN apk --update add ca-certificates
 
 WORKDIR /workspace
 
-ARG TARGETPLATFORM
-RUN echo "Target platform is $TARGETPLATFORM"
```
sethAmazon, line 13: Think we should echo the target arch we are running for

Makefile
Aneurysm9: Is this just an intermediate name used during the build process?  Does the name of the executable included in the `deb`, `rpm`, or container image change?

Aneurysm9 (commented):
Is it possible to separate out the lint/formatting changes?  They make it hard to have confidence that the functional changes have all been identified.

jefchien:
> Is it possible to separate out the lint/formatting changes? They make it hard to have confidence that the functional changes have all been identified.

@Aneurysm9  I've separated it out into two commits. The first one should be the actual changes and the second one will be the linting changes.

Makefile
jefchien: Yes. This is just an intermediate name. `create_rpm.sh` and `create_deb.sh` both copy the binary in and rename it to `aws-otel-collector`. The Dockerfile also renames it, but to `awscollector`.

cmd/awscollector/Dockerfile
jefchien: Wouldn't the TARGETARCH just show up in the COPY instruction when it tries to find the file?

Aneurysm9 (approved):

Makefile
anuraaga: Yeah intermediate names should not need to be preserved.

While we're changing these, can we dedupe, and maybe simplify, by moving into the folder

`./build/linux-amd64/aoc`?

sethAmazon (approved):

PR #891, opened 2022-01-15
Always upload terraform state to know what resources have been created and destroy a second time (avoid failed destroy and have resources exist)

khanhntd:

Sometimes, the ```terraform destroy``` does not always destroy resources; therefore, we can upload the ```terraform state``` to s3 bucket to record it and destroy it later (same as destroy second times).
 
N/A

Testing it with my branch on the CI workflow: https://github.com/aws-observability/aws-otel-collector/actions/runs/1729411415

N/A


.github/workflows/CI.yml
```diff
@@ -77,82 +77,82 @@ jobs:
...
-      if: steps.cached_binaries.outputs.cache-hit != 'true'
-      run: make build
-
-    # upload the binaries to artifact as well because cache@v2 hasn't support windows
-    - name: Upload
-      uses: actions/upload-artifact@v2
-      with:
-        name: binary_artifacts
-        path: build
+      # Set up building environment, patch the dev repo code on dispatch events.
```
sethAmazon, line 80: Will the change indents still be valid yml?

.github/workflows/CD.yml
```diff
@@ -196,12 +196,21 @@ jobs:
...
           if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars" ; else opts="" ; fi
           cd testing-framework/terraform/ec2 && terraform init && terraform apply -auto-approve -lock=false $opts -var="testing_ami=${{ matrix.testing_ami }}" -var="aoc_version=${{ needs.release-checking.outputs.version }}" -var="package_s3_bucket=aws-otel-collector" -var="testcase=../testcases/${{ matrix.testcase }}" -var="ssm_package_name=AWSDistroOTel-Collector"
-          
+
+      - name: Upload test case's terraform state
+        if: ${{ always() && steps.s3-release-validation-1.outputs.cache-hit != 'true' }}
+        run: |
+          cd testing-framework/terraform/eks
+          testing_id=$(terraform output -raw testing_id)
+          cd ../add_on/remote_state
```
jefchien, line 205: I think I must've missed the PR, but why are we using terraform to upload to S3? Wouldn't `aws s3 cp` do the same thing? Do you need to include the entire testing framework? We used to for the soaking test since it was possible to have backwards incompatible changes.

https://github.com/aws-observability/aws-otel-collector/blob/c926da59063d2173ced4094c4f58ccb335745af4/.github/workflows/soaking.yml#L189-L195

.github/workflows/CI.yml
```diff
@@ -682,6 +697,8 @@ jobs:
       - name: Set up terraform
         if: steps.e2etest-ec2-2.outputs.cache-hit != 'true'
         uses: hashicorp/setup-terraform@v1
+        with:
+          terraform_wrapper: false
```
jefchien, line 701: Why is this being disabled?

.github/workflows/CI.yml
khanhntd: Yes. As far as I concerned and tested it out. However, I can delete these format changes to avoid further problems though.

.github/workflows/CD.yml
khanhntd: No we don't need to include the entire workflow, only the terraform.state is fine to since we can destroy the resources by checking for the platform which the terraform.state is in. Mostly, everything should be managed by terraform for easier management and avoiding errors though.

.github/workflows/CI.yml
khanhntd: It would be failed when assigning  ```terraform output {{output variable}}``` to a variable since by using ```terraform_wrapper=true```, it will assign the variable the whole raw command instead of getting the output from it. More in depth read is [in this article ](https://blog.nillsf.com/index.php/2020/08/25/how-to-use-terraform-output-in-a-github-action/). Moreover, I tested this out with and without using this and seems the former works for me.

khanhntd:
Close on this since @jefchien suggested a better idea on this by using ```nick-revision/retry``` (https://github.com/aws-observability/aws-otel-collector/pull/919)

PR #898, opened 2022-01-18
Remove hourly soaking test terraform state cleaner.

jefchien:
**Description:** All of the old soaking test terraform states have been successfully destroyed and removed from S3. We no longer have the soaking tests running, so no more terraform states will be uploaded and this workflow can be removed.


khanhntd:
Any chances we could keep this since since it is needed for destroying uploaded terraform state in s3 though?

khanhntd (approved):
LGTM.

sethAmazon (approved):

PR #901, opened 2022-01-19
Delete launch configuration

khanhntd:

Currently, the integration test account's launch configuration has reached its quota. Therefore, deleting old launch configuration and auto-scaling group is compulsory.

N/A

Testing locally.

N/A

### DIFF
--- a/tools/workflow/clean_aws_resources.go
+++ b/tools/workflow/clean_aws_resources.go
@@ -16,6 +16,7 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


tools/workflow/clean_aws_resources.go
```diff
@@ -97,17 +105,172 @@ func terminateEc2Instances() {
...
+
+			//Skipping lc that does not older than 5 days
+			if !time.Now().UTC().Add(pastDayDeleteCalculation).After(*lc.CreatedTime) {
+				continue
+			}
+
+			//Split the instance profile to get the testing ID associate with the testing framework
+			splitLcIamInstanceProfile := strings.Split(*lc.IamInstanceProfile, "-")
+
+			if len(splitLcIamInstanceProfile) != 6 {
```
jefchien, line 214: 6 feels arbitrary. Make it a named const.

jefchien (commented):
This clean_aws_resources.go file is getting a bit large. Might be worth splitting it out.

tools/workflow/clean_aws_resources.go
khanhntd: Reasonable. Changed to ```requiredIAMInstanceProfileLength``` in newest commit. About splitting, agree. Still finding a boilerpoint to do it. Moreover, one problem to address when splitting the file is: should we still used go language and maintaining it dependencies (same as tool linters) or should we use bash? Personally, I would prefer bash but going to far to turn back to use bash now.

jefchien: It doesn't hurt to still use go. If we do split it up, it shouldn't be in this PR.

jefchien (approved):

PR #903, opened 2022-01-20
EKS ARM64 Test Case Matrix Creation

sethAmazon:
**Description:** 
Container insights prometheus test does not work on arm due to helm chart not supporting arm64. Thus removed from test in arm cluster.


khanhntd (approved):
LGTM.

jefchien (approved):

PR #907, opened 2022-01-21
Add release rollback workflow.

jefchien:
**Description:** With the added checks in the CD workflow, we're no longer able to rollback the latest on DockerHub/ECR by re-running the CD with an older version. This new workflow would be used in place of that. It basically checks to make sure that the DockerHub/ECR images for the rollback version exist along with the S3 objects and SSM documents. If those are all available, then it will retag the latest for the images, replace the S3 latest, and change the default version in SSM back to the rollback version.

It doesn't delete anything.

**Testing:** Tested all of the checks, but not the actual rollbacks. The image retagging was taken from the CD workflow. The S3 rollback is based off of the `s3_release` script. The SSM rollback was tested locally.

### DIFF
--- a/.github/workflows/rollback.yml
+++ b/.github/workflows/rollback.yml
@@ -1,0 +1,304 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+name: Release rollback
+
+on:
+  workflow_dispatch:
+    inputs:
+      rollback_version:
+        description: 'the version number to rollback to'
+        required: true
+
+env:
+  DOCKERHUB_REPO: amazon/aws-otel-collector
+  ECR_REPO: aws-observability/aws-otel-collector
+  S3_BUCKET: aws-otel-collector
+  SSM_RELEASE_PACKAGE_NAME: AWSDistroOTel-Collector
+  SSM_ARTIFACT_ROOT: rollback/ssm
+
+jobs:
+  image-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check version exists on Release ECR/DockerHub repos
+        run: |
+          docker manifest inspect public.ecr.aws/$ECR_REPO:${{ github.event.inputs.rollback_version }} > /dev/null
+          docker manifest inspect $DOCKERHUB_REPO:${{ github.event.inputs.rollback_version }} > /dev/null
+  s3-check:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      max-parallel: 5
+      matrix:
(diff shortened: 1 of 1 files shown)


sethAmazon:
How did you test this?

.github/workflows/rollback.yml
```diff
@@ -0,0 +1,307 @@
...
+          aws-secret-access-key: ${{ secrets.RELEASE_MXP_SECRET }}
+          aws-region: eu-south-1
+      - name: Configure AWS Credentials for CN
+        if: matrix.type == 'cn'
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.RELEASE_CN_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.RELEASE_CN_SECRET }}
+          aws-region: cn-north-1
+      - name: Check version exists in SSM
```
sethAmazon, line 138: Should this happen as a job before this one and have this job require this step?

.github/workflows/rollback.yml
```diff
@@ -0,0 +1,307 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - type: global
+            regions: "eu-north-1 ap-south-1 eu-west-3 us-east-2 eu-west-1 eu-central-1 sa-east-1 us-east-1 \
+                  ap-northeast-2 ap-northeast-3 eu-west-2 ap-northeast-1 us-west-2 us-west-1 ap-southeast-1 \
+                  ap-southeast-2 ca-central-1"
+          - type: hkg
```
sethAmazon, line 85: Why do we have to log in to specific region for these? I thought only gov and china required this.

jefchien:
> How did you test this?

Tested individual parts of this in different contexts. Mostly through a test branch on my fork. All the check steps have been tested. The `s3-rollback` and `ssm-rollback` haven't been fully tested. Will try it out with a different account.

.github/workflows/rollback.yml
jefchien: This is how the `ssm-release` workflow is set up as well. These are opt-in regions and part of different release accounts. We might not need to login for the `list-document-versions` since it's public, but I'm not sure. We definitely need to login for the `update-document-default-version`.

jefchien: Yeah just tried to run it in my branch with only the us-west-2 login and it wasn't authorized to `list-document-versions`.

jefchien: https://github.com/jefchien/aws-otel-collector/runs/4902645939?check_suite_focus=true

Aneurysm9 (approved):
Do we need to do anything about removing a version that we rolled back from?

jefchien:
> Do we need to do anything about removing a version that we rolled back from?

I'm not completely sure. Is there a standard practice for this sort of thing?

anuraaga (approved):
We shouldn't ever remove versions, so makes sense that we're only updating the floating tags here

.github/workflows/rollback.yml
```diff
@@ -0,0 +1,304 @@
...
+            else
+              echo "$region $document_version" >> ${{ env.SSM_ARTIFACT_ROOT }}/versions.txt
+            fi
+          done
+      - name: Upload versions artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.type }}-versions
+          path: ${{ env.SSM_ARTIFACT_ROOT }}
+  all-checked:
```
khanhntd, line 156: IMO should we leave this since these conditions and steps can be run on ```image-rollback``` jobs though?

khanhntd (approved):
LGTM.

PR #910, opened 2022-01-24
Separate documentation sync from performance test runs.

jefchien:
**Description:** Moved documentation sync out of performance test workflow. They aren't related.

**Testing:** Runs the same commands as the existing performance test workflow.

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -180,8 +180,6 @@ jobs:
       - name: Commit to gh-pages branch
         run: |
           git switch ${{ env.GH_PAGES_BRANCH }}
-          git restore --source=main -- README.md CONTRIBUTING.md RELEASING.md docs
-          git add README.md CONTRIBUTING.md RELEASING.md docs
           rsync -avv performance-report.md benchmark/report.md
           rsync -avv performance-data.json benchmark/data/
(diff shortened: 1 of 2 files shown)


sethAmazon (approved):

khanhntd:
Should the ```document sync``` workflow be sync after the performance test if I understand the ```document sync``` workflow correctly: to sync the performance table into the main branch ?

jefchien:
> Should the `document sync` workflow be sync after the performance test if I understand the `document sync` workflow correctly: to sync the performance table into the main branch ?

The document sync workflow is just to get the README and all other docs synced with the main branch. It has nothing to do with the performance tests or table, which is why I've separated it.

khanhntd:
> > Should the `document sync` workflow be sync after the performance test if I understand the `document sync` workflow correctly: to sync the performance table into the main branch ?
> 
> The document sync workflow is just to get the README and all other docs synced with the main branch. It has nothing to do with the performance tests or table, which is why I've separated it. This is necessary because if only document changes are made, the CI doesn't run and the performance tests won't get triggered, so the docs on the gh-pages site won't be up to date.

Reasonable. Understanding more about the doc now though.

khanhntd (approved):
LGTM

Issue #916, opened 2022-01-25
PR build and canary tests failing


jefchien:
**Describe the bug**
There's two parts to the issue. For the PR builds, the ECR pulls are rate limited to 1 per second since they are unauthenticated (https://docs.aws.amazon.com/AmazonECR/latest/public/public-service-quotas.html). We'll need to figure out a way to either stagger the test case runs or somehow increase our pull limit.

For both the PR builds and the canary, there are a couples tests that are failing consistently: `otlp_grpc_exporter_metric_mock` and `otlp_http_exporter_metric_mock`. There are also a couple tests that are now flakier and occasionally fail: `otlp_metric_mock`, `signalfx_exporter_metric_mock`.

**Steps to reproduce**
Run the PR build. Look at the canary.

Ran 4 attempts of the PR build https://github.com/aws-observability/aws-otel-collector/actions/runs/1742363763
2 of those failed due to rate limits and 2 of them failed due to the failing tests.

PR #917, opened 2022-01-25
Add retry to PR build with a random delay to prevent rate limiting

jefchien:
**Description:** Public ECR has an unauthenticated pull limit of 1 per second. We're having issues with multiple tests pulling the image at the same time. Trying to prevent that sort of collision with a delayed retry on failure.

Also fixed the build job so it actually uses the cache.

**Link to tracking Issue:** https://github.com/aws-observability/aws-otel-collector/issues/916

### DIFF
--- a/.github/workflows/PR-build.yml
+++ b/.github/workflows/PR-build.yml
@@ -58,128 +58,126 @@ jobs:
     steps:
     # Set up building environment, patch the dev repo code on dispatch events.  
     - name: Set up Go 1.x
-      uses: actions/setup-go@v2
       if: ${{ needs.changes.outputs.changed == 'true' }}
+      uses: actions/setup-go@v2
       with:
         go-version: 1.17
 
-
-    - uses: actions/checkout@v2
+    - name: Checkout
       if: ${{ needs.changes.outputs.changed == 'true' }}
+      uses: actions/checkout@v2
 
-    - name: Cache Build
+    - name: Cache go
+      if: ${{ needs.changes.outputs.changed == 'true' }}
       uses: actions/cache@v2
-      if: ${{ needs.changes.outputs.changed == 'true' }}
       with:
         path: |
           ~/go/pkg/mod
           ~/.cache/go-build
         key: v1-go-pkg-mod-${{ runner.os }}-${{ hashFiles('**/go.sum') }}
 
-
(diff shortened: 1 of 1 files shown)


sethAmazon (approved):

.github/workflows/PR-build.yml
```diff
@@ -209,44 +207,51 @@ jobs:
...
+        uses: nick-invision/retry@v2
+        with:
+          max_attempts: 3
+          timeout_minutes: 10
+          command: |
+            if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars" ; else opts="" ; fi
+            cd testing-framework/terraform/mock && terraform init && terraform apply -auto-approve -var="testcase=../testcases/${{ matrix.testcase }}" $opts
+          on_retry_command: |
+            cd testing-framework/terraform/mock && terraform destroy -auto-approve
+            sleep $((RANDOM % 10))
```
khanhntd, line 255: Instead of using sleep, it would be better to use [retry wait seconds](https://github.com/nick-invision/retry#retry_wait_seconds) though

khanhntd:
If it's good, adding retry_on_second for CI workflow would be a better idea too.

.github/workflows/PR-build.yml
jefchien: I looked at that, but the main purpose of this is to stagger the retries. The hope is that if two tests retry at the same time because they both hit the rate limit, they don't run at exactly the same time again. If I could set a random value to `retry_wait_seconds` then that would work. Not sure if I can.

khanhntd: Interesting. So random value for `retry_wait_seconds` is not possible and random % 10 is quite a small range though. Anyway we could use the image from docker hub instead of ECR image to avoid pulling quota?

jefchien: DockerHub has a rate limit as well. I think we're seeing if we can get a rate limit increase for the public ECR. A random 0-10 second delay should be good enough.

sethAmazon:
Going to merge this for now to unblock the ci. We can look for a solution to the ci backoff time in another pr.

PR #923, opened 2022-01-26
Add IAM role clean to cleaner.

jefchien:
**Description:** Cleans IAM roles starting with `terraform`, `aoc-eks-assume-role`, or `fargate-profile-role`.

**Testing:** `export AWS_SDK_LOAD_CONFIG=true && go run tools/workflow/clean_aws_resources.go` with other functions removed.

### DIFF
--- a/tools/workflow/clean_aws_resources.go
+++ b/tools/workflow/clean_aws_resources.go
@@ -25,6 +25,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/aws/aws-sdk-go/service/iam"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


tools/workflow/clean_aws_resources.go
```diff
@@ -322,3 +328,67 @@ func destroyLoadBalancerResource() {
...
+		}
+
+		for _, role := range lro.Roles {
+			var hasPrefix bool
+			for _, prefix := range roleNamePrefixes {
+				if hasPrefix = strings.HasPrefix(*role.RoleName, prefix); hasPrefix {
+					break
+				}
+			}
+			if hasPrefix && time.Now().UTC().Add(pastDayDeleteCalculation).After(*role.CreateDate) {
```
sethAmazon, line 358: We may want to also have it delete by date since last used

jefchien: For some reason this is always nil, but I can add it in on the off chance that it's available.

```go
if role.RoleLastUsed != nil {
  log.Printf("Last used %s %v", *role.RoleName, *role.RoleLastUsed.LastUsedDate)
}
```

khanhntd: Prefer last used date instead of create date.

jefchien: It's nil for all of the roles in our account. I've added it to the conditional in case it's ever not nil.

sethAmazon: nil I think means not used

sethAmazon (approved):

tools/workflow/clean_aws_resources.go
jefchien: It seems like a bug in the AWS SDK https://github.com/aws/aws-sdk/issues/24

PR #924, opened 2022-01-26
Refactor terraform apply steps in workflows.

jefchien:
**Description:** Remove check for parameters file since every testcase includes a parameters file and I can't see a case where that wouldn't be true. Used the `on_retry_command` to destroy the terraform state instead of the if else.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -639,19 +639,18 @@ jobs:
           max_attempts: 3
           timeout_minutes: 120
           command: |
-            opts=""
-            if [[ -f testing-framework/terraform/testcases/${{ matrix.testcase }}/parameters.tfvars ]] ; then opts="-var-file=../testcases/${{ matrix.testcase }}/parameters.tfvars" ; fi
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -1064,28 +1069,24 @@ jobs:
...
-      #This is here just in case workflow cancel
       - name: Destroy resources
-        if: ${{ cancelled() && steps.e2etest-eks-fargate.outputs.cache-hit != 'true' }}
+        if: ${{ always() && steps.e2etest-eks-fargate.outputs.cache-hit != 'true' }}
         uses: nick-invision/retry@v2
         with:
           max_attempts: 3
           timeout_minutes: 30
           retry_wait_seconds: 20
           command: |
```
khanhntd, line 1088: We don't need `|` if it is only one line though

khanhntd:
We should not merge this PR before we find out why ```nick-invision/retry``` not works  with ```terraform destroy``` since we convert to use always though.

PR #927, opened 2022-01-27
Refactor resource cleaners.

jefchien:
**Description:** Separated the cleaners into packages based on type. The code within the cleaners themselves is basically the same. Added two flags `keep` and `clean` to specify the days to keep the resources before cleaning and the cleaners to use respectively. Changed the workflow to run each cleaner separately.

**Testing:** `export AWS_SDK_LOAD_CONFIG=true && go run tools/workflow/cleaner/main.go -keep 10 -clean ec2,loadbalancer`


tools/workflow/cleaner/main.go
```diff
@@ -0,0 +1,97 @@
...
+	// flags
+	daysToKeep    int
+	cleanersToRun string
+
+	cleanerTypes   = []string{autoscaling.Type, ec2.Type, efs.Type, iam.Type, launchconfig.Type, loadbalancer.Type}
+	cleanerOptions = strings.Join(cleanerTypes, delimiter)
+)
+
+func init() {
+	flag.IntVar(&daysToKeep, "keep", defaultDaysToKeep, "Days to keep a resource before cleaning it.")
```
khanhntd, line 51: Nice on using flags though.

.github/workflows/weekly-resources-clean.yml
```diff
@@ -17,10 +17,29 @@ on:
   schedule:
     - cron: "0 0 * * SAT" # Run every saturday at midnight
   workflow_dispatch:
+    inputs:
+      daystokeep:
+        description: "days to keep a resource"
+        required: true
```
khanhntd, line 23: IMO, we don't need this to be required, set this as a default value by using ```github.event.inputs.daystokeep || 5``` would simpler though

jefchien: Sure. I think it's not a big deal since this is for manual runs, which shouldn't normally happen anyway.

jefchien: Feels a bit weird to have inputs without one of the fields required. At least to me.

khanhntd (approved):
LGTM.

khanhntd (approved):
LGTM. Merge now to unblock the CD workflow.

PR #942, opened 2022-01-31
Fix IAM cleaner so the policies, profiles, and roles are all deleted.

jefchien:
**Description:** Cleaners should log errors, but all should still run.

**Testing:** Ran the delete locally for IAM using the integ-test account.

### DIFF
--- a/tools/workflow/cleaner/iam/clean.go
+++ b/tools/workflow/cleaner/iam/clean.go
@@ -16,73 +16,157 @@
 package iam
 
 import (
+	"fmt"
 	"log"
+	"os"
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/iam"
 )
 
 const Type = "iam"
 
-var roleNamePrefixes = []string{"terraform", "aoc-eks-assume-role", "fargate-profile-role"}
+var (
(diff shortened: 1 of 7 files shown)


sethAmazon:
Why are so many files being touched? Can you explain this pr a bit more. Or maybe we can do a live review.

jefchien:
> Why are so many files being touched? Can you explain this pr a bit more. Or maybe we can do a live review.

The main change was to remove all `log.Fatalf` calls and have them return the error instead. Then, `main.go` logs the errors instead of exiting immediately. I guess this isn't completely necessary since the cleaner workflow runs each individually now.

tools/workflow/cleaner/iam/clean.go
```diff
@@ -16,73 +16,157 @@
...
-						if err != nil {
-							log.Fatalf("Failed to detach policy %s from %s because of %v", *policy.PolicyName, *role.RoleName, err)
-						}
-					}
-
-					if larpo.Marker == nil {
-						break
-					}
-					policiesMarker = larpo.Marker
+				if err = deleteProfiles(client, expirationDate, role.RoleName); err != nil {
```
khanhntd, line 66: I thought we would need to delete Profiles, then Roles, then Policies instead of this order though based on this error from the workflow https://github.com/aws-observability/aws-otel-collector/runs/5008125297?check_suite_focus=true

jefchien: That's the order I'm doing it in. I'm detaching the policies, deleting the profiles, deleting the roles, and then deleting the policies.

khanhntd (approved):
LGTM.

PR #944, opened 2022-01-31
Add linting check to PR build.

jefchien:


vasireddy99 (approved):

khanhntd:
Should we need this since the ```make build``` has included the link check though? https://github.com/aws-observability/aws-otel-collector/blob/main/Makefile#L79

jefchien:
> Should we need this since the `make build` has included the link check though? https://github.com/aws-observability/aws-otel-collector/blob/main/Makefile#L79

You're right. Don't think we need this then.

PR #946, opened 2022-01-31
Added skip rollback options to workflow.

jefchien:
**Description:** Fixed parameter expansion for AWS CLI commands.

### DIFF
--- a/.github/workflows/rollback.yml
+++ b/.github/workflows/rollback.yml
@@ -16,9 +16,21 @@ name: Release rollback
 on:
   workflow_dispatch:
     inputs:
-      rollback_version:
+      rollback-version:
         description: 'the version number to rollback to'
         required: true
+      skip-image-rollback:
+        description: 'skip image rollback'
+        required: false
+        type: boolean
+      skip-s3-rollback:
+        description: 'skip s3 rollback'
+        required: false
+        type: boolean
+      skip-ssm-rollback:
+        description: 'skip ssm rollback'
+        required: false
+        type: boolean
 
 env:
   DOCKERHUB_REPO: amazon/aws-otel-collector
@@ -26,17 +38,20 @@ env:
   S3_BUCKET: aws-otel-collector
   SSM_RELEASE_PACKAGE_NAME: AWSDistroOTel-Collector
   SSM_ARTIFACT_ROOT: rollback/ssm
+  SSM_VERSIONS_FILE_NAME: versions.txt
 
 jobs:
   image-check:
     runs-on: ubuntu-latest
+    if: github.event.inputs.skip-image-rollback == 'false'
     steps:
       - name: Check version exists on Release ECR/DockerHub repos
         run: |
-          docker manifest inspect public.ecr.aws/$ECR_REPO:${{ github.event.inputs.rollback_version }} > /dev/null
-          docker manifest inspect $DOCKERHUB_REPO:${{ github.event.inputs.rollback_version }} > /dev/null
+          docker manifest inspect public.ecr.aws/$ECR_REPO:${{ github.event.inputs.rollback-version }} > /dev/null
+          docker manifest inspect $DOCKERHUB_REPO:${{ github.event.inputs.rollback-version }} > /dev/null
   s3-check:
     runs-on: ubuntu-latest
+    if: github.event.inputs.skip-s3-rollback == 'false'
     strategy:
       fail-fast: true
       max-parallel: 5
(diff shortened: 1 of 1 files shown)


.github/workflows/rollback.yml
```diff
@@ -94,61 +110,63 @@ jobs:
             regions: "cn-north-1 cn-northwest-1"
     steps:
       - name: Configure AWS Credentials
-        if: matrix.type == 'global'
+        if: ${{ matrix.type == 'global' && github.event.inputs.skip-ssm-rollback == 'false' }}
```
bryan-aguilar, line 113: Is it possible to add this as a precondition to the job rather than have it in each step?

jefchien: Unfortunately not. If I have it at the job level, the job itself will count as cancelled and the dependent jobs like `all-checked` will treat that as a failure.

.github/workflows/rollback.yml
```diff
@@ -16,25 +16,39 @@ name: Release rollback
 on:
   workflow_dispatch:
     inputs:
-      rollback_version:
+      rollback-version:
         description: 'the version number to rollback to'
         required: true
+      skip-image-rollback:
```
khanhntd, line 22: Maybe used default false for boolean type when the required is false though.

.github/workflows/rollback.yml
```diff
@@ -214,18 +236,20 @@ jobs:
...
-          VERSION_KEY="${{ matrix.os }}/${{ matrix.arch }}/${{ github.event.inputs.rollback_version }}/aws-otel-collector.${{ matrix.type }}"
+          VERSION_KEY="${{ matrix.os }}/${{ matrix.arch }}/${{ github.event.inputs.rollback-version }}/aws-otel-collector.${{ matrix.type }}"
           LATEST_KEY="${{ matrix.os }}/${{ matrix.arch }}/latest/aws-otel-collector.${{ matrix.type }}"
-          VERISON_URI="s3://$S3_BUCKET/$VERSION_KEY"
-          LATEST_URI="s3://$S3_BUCKET/$LATEST_KEY"
-          aws s3 cp $VERSION_URI $LATEST_URI --acl public-read
+          VERISON_URI="s3://${S3_BUCKET}/${VERSION_KEY}"
+          LATEST_URI="s3://${S3_BUCKET}/${LATEST_KEY}"
+          aws s3 cp "${VERSION_URI}" "${LATEST_URI}" --acl public-read
   ssm-rollback:
```
khanhntd, line 253: This could use the job conditions instead of steps condition.

.github/workflows/rollback.yml
jefchien: Boolean types are false by default.

bryan-aguilar (approved):
LGTM thanks for the changes!

.github/workflows/rollback.yml
```diff
@@ -221,14 +244,15 @@ jobs:
...
     runs-on: ubuntu-latest
     needs: [all-checked]
+    if: always() && needs.all-checked.result == 'success' && github.event.inputs.skip-ssm-rollback == 'false'
```
khanhntd, line 255: Wait why would we want to use always with needs.all-checked.result though?

jefchien: It doesn't work otherwise. It seems like if a job is skipped, every dependent job is also skipped. So we need to get around this by having the always with the check to basically have the same effect as the normal needs.

khanhntd: Weird to me with the Github conditions. Cause it only needs the ```all-checked``` job and does not have any children job.

jefchien: I've tested this in a fork. Without the condition, it won't even run this job.

.github/workflows/rollback.yml
```diff
@@ -140,12 +156,12 @@ jobs:
           mkdir -p ${{ env.SSM_ARTIFACT_ROOT }}
           for region in ${{ matrix.regions }}; do
             document_versions=$(aws ssm list-document-versions --name ${SSM_RELEASE_PACKAGE_NAME} --region ${region})
```
khanhntd, line 158: Might be good to check the binaries that are released in S3 associate with SSM before checking version SSM thought.

jefchien: Added it.

bryan-aguilar:
I am going to merge this. We plan on testing the rollback workflow before February release.

PR #950, opened 2022-01-31
Add ECS cleaner.

jefchien:
**Description:** Add mount target deletion in EFS cleaner as well.

**Testing:** Ran both EFS and ECS cleaners.

### DIFF
--- a/tools/workflow/cleaner/autoscaling/clean.go
+++ b/tools/workflow/cleaner/autoscaling/clean.go
@@ -28,9 +28,9 @@ import (
 
 const Type = "autoscaling"
 
-var logger = log.New(os.Stdout, fmt.Sprintf("[%s] ", Type), log.Ldate)
+var logger = log.New(os.Stdout, fmt.Sprintf("[%s] ", Type), log.LstdFlags)
 
-func Clean(sess *session.Session, keepDuration time.Duration) error {
+func Clean(sess *session.Session, expirationDate time.Time) error {
 	logger.Printf("Begin to clean ECS AutoScaling")
 
 	autoscalingclient := autoscaling.New(sess)
@@ -54,7 +54,7 @@ func Clean(sess *session.Session, keepDuration time.Duration) error {
 
 		for _, asg := range describeAutoScalingOutputs.AutoScalingGroups {
 			//Skipping asg that does not older than 5 days
-			if !time.Now().UTC().Add(keepDuration).After(*asg.CreatedTime) {
+			if !expirationDate.After(*asg.CreatedTime) {
 				continue
 			}
 
--- a/tools/workflow/cleaner/ecs/clean.go
+++ b/tools/workflow/cleaner/ecs/clean.go
@@ -1,0 +1,136 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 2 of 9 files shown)


tools/workflow/cleaner/ecs/clean.go
```diff
@@ -0,0 +1,70 @@
...
+var logger = log.New(os.Stdout, fmt.Sprintf("[%s] ", Type), log.LstdFlags)
+
+func Clean(sess *session.Session, expirationDate time.Time) error {
+	logger.Printf("Begin to clean ECS Task Definitions")
+	client := ecs.New(sess)
+
+	var nextToken *string
+	for {
+		ltdi := &ecs.ListTaskDefinitionsInput{NextToken: nextToken, MaxResults: aws.Int64(10)}
+		ltdo, err := client.ListTaskDefinitions(ltdi)
```
khanhntd, line 44: Should we delete the services and task definition after de-registering it? In this way, we can delete the ECS too since the ECS does not have the creation time and we can use these two creation times to delete it.

tools/workflow/cleaner/autoscaling/clean.go
```diff
@@ -28,9 +28,9 @@ import (
 
 const Type = "autoscaling"
 
-var logger = log.New(os.Stdout, fmt.Sprintf("[%s] ", Type), log.Ldate)
+var logger = log.New(os.Stdout, fmt.Sprintf("[%s] ", Type), log.LstdFlags)
 
-func Clean(sess *session.Session, keepDuration time.Duration) error {
+func Clean(sess *session.Session, expirationDate time.Time) error {
```
bryan-aguilar, line 33: I would like to request a comment be added to explain the usage of expiration date and how it is utilized. Readability is a little hard since it is being compared to resource creation date.

tools/workflow/cleaner/ecs/clean.go
```diff
@@ -0,0 +1,137 @@
...
+		dco, err := client.DescribeClusters(dci)
+		if err != nil {
+			return err
+		}
+		for _, cluster := range dco.Clusters {
+			if strings.HasPrefix(*cluster.ClusterName, clusterPrefix) {
+				if *cluster.ActiveServicesCount > 0 {
+					allServicesDeleted, err := deleteServices(client, expirationDate, cluster.ClusterArn)
+					if err != nil {
+						return err
```
bryan-aguilar, line 59: Is returning an err here the desirable behavior? Wouldn't we want to iterate through all clusters even if one attempted deletion failed? Is there a specific scenario where we want to abort when one cluster fails to delete?

tools/workflow/cleaner/ecs/clean.go
```diff
@@ -0,0 +1,137 @@
...
+					}
+
+					if allServicesDeleted {
+						logger.Printf("Trying to delete cluster %s", *cluster.ClusterName)
+						dlci := &ecs.DeleteClusterInput{Cluster: cluster.ClusterArn}
+						if _, err = client.DeleteCluster(dlci); err != nil {
+							return err
```
bryan-aguilar, line 66: Same thought as returning an err here as above.

bryan-aguilar (approved):
LGTM

PR #952, opened 2022-02-01
Modify default version to previous version to avoid failing when deleting SSM version

khanhntd:

Currently, [the CD fails ](https://github.com/aws-observability/aws-otel-collector/runs/5014507743?check_suite_focus=true) if we delete the default version in SSM. Therefore, we need to make the second largest version in SSM as default and delete the current version if the validation test fails ( when triggering the roll back)
N/A
Testing locally
https://docs.aws.amazon.com/cli/latest/reference/ssm/update-document-default-version.html


tools/workflow/docker-version-compare.sh
```diff
@@ -98,3 +98,4 @@ echo "::set-output name=minor-update::${MINOR_UPDATE}"
 echo "::set-output name=patch-update::${PATCH_UPDATE}"
 echo "::set-output name=any-update::${ANY_UPDATE}"
 echo "::set-output name=same-version::${SAME_VERSION}"
+echo "::set-output name=latest-version::${LATEST_VERSION}"
```
jefchien, line 101: Oh this isn't going to work. The latest version is considered v0.16.0. It's just getting the largest version. It isn't checking the digest for both tags.

khanhntd: Yes. You got it right though. Will try to change this in the  next commit.

.github/workflows/CD.yml
```diff
@@ -639,9 +647,46 @@ jobs:
...
+          aws-region: us-west-2
+
+      - name: Restore cached packages
+        id: cache_packages
+        uses: actions/cache@v2
+        with:
+          key: ${{ env.PACKAGE_CACHE_KEY }}
+          path: ${{ env.PACKAGING_ROOT }}
+
+      - name: Update SSM default version
```
jefchien, line 672:
```suggestion
      - name: Rollback SSM default version
```

.github/workflows/CD.yml
```diff
@@ -639,9 +647,46 @@ jobs:
...
+      #-if the validation test failed, then we will delete the SSM package since we don't want to roll out the unstable version
+      #-if the validation successes, then we will trigger the release SSM package workflow to replace all of this with the appropriate region
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.RELEASE_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.RELEASE_SECRET }}
+          aws-region: us-west-2
+
+      - name: Restore cached packages
```
jefchien, line 665: What do you need the cached packages for?

tools/ssm/ssm_update_default_version.sh
```diff
@@ -0,0 +1,61 @@
...
+
+function update_second_largest_target_document_as_default() {
+    list_document_versions=$(aws ssm list-document-versions --name "${ssm_package_name}" --output json)
+    target_document_version=$(echo $list_document_versions | docker run --rm -i stedolan/jq -c -r ".DocumentVersions|.[]|select(.VersionName==\"${version}\")|.DocumentVersion")
+
+    if [[ -z $target_document_version ]]; then
+        error_exit "Document version not found for ${version}"
+    fi
+
+    second_largest_document_version="$((target_document_version - 1))"
```
jefchien, line 53: This might not always be true. Especially if we take into account patching.

.github/workflows/CD.yml
```diff
@@ -639,9 +647,46 @@ jobs:
...
+        id: cache_packages
+        uses: actions/cache@v2
+        with:
+          key: ${{ env.PACKAGE_CACHE_KEY }}
+          path: ${{ env.PACKAGING_ROOT }}
+
+      - name: Update SSM default version
+        if: steps.cache_packages.outputs.cache-hit == 'true'
+        run: ssm_package_name=${{ env.SSM_RELEASE_PACKAGE_NAME }} version=${{ needs.release-checking.outputs.version }}  tools/ssm/ssm_update_default_version.sh
```
jefchien, line 674: The script should probably mention that it's rolling back the default version instead of just updating it. From the name, it sounds like you're updating the default version to the version passed in.

.github/workflows/CD.yml
khanhntd: Reasonable. Fixed it in newest commit.

.github/workflows/CD.yml
```diff
@@ -544,6 +544,14 @@ jobs:
           key: release-latest-s3-${{ github.run_id }}
           path: VERSION
 
+      - name: Configure AWS Credentials
```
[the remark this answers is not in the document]
khanhntd: Nice catch. The merge did not show for me and forgot to check it. Thanks for pointing it out though.

.github/workflows/CD.yml
khanhntd: Mostly the cached packages are for checking if the SSM release is a success or not. However, add a condition and only trigger when```release-to-s3``` job success.

.github/workflows/CD.yml
khanhntd: Reasonable. Fixed it in newest commit.

khanhntd:
Hey @bryan-aguilar, no worries and I have updated with the conflicts. Let's me know if there are any reviews regarding this PR.

tools/ssm/ssm_rollback_default_version.sh
```diff
@@ -0,0 +1,75 @@
...
+    if [[ ${latest_document_version_name} == ${version} ]]; then
+      rollback_version_name="${second_latest_document_version_name}"
+    else
+      rollback_version_name="${latest_document_version_name}"
+    fi
+
+    echo "Roll-back the default SSM version to ${rollback_version_name}"
+
+    rollback_document_version=$(aws ssm describe-document --name "${ssm_package_name}" --version-name "${rollback_version_name}" --output json | docker run --rm -i stedolan/jq -c -r ".Document|.DocumentVersion")
+    aws ssm update-document-default-version --name ${ssm_package_name} --document-version "${rollback_document_version}"
```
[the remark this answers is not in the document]
khanhntd: For the first question, to ensure it would roll back with the latest default when manually set the older SSM version as default. Other than that, then yes we can exit if the mentioned scenario is a black swan case.
For the second question, the version as default will still be v0.17.0. The solution is simple, sorted `the document version name`, in your case, the latest will be v0.17.0; however, the target if v0.16.2 which is different from v0.17.0. Therefore, the rollback version would be v0.17.0.

bryan-aguilar (approved):

PR #955, opened 2022-02-01
Update flags.go to be in sync with upstream

### LINKED ISSUE #951: flag.go out of sync with upstream
**Describe the bug**
During the v0.16.0 release an upstream change to `flags.go` file was not identified and updated. Since the aws collector maintains a non tracked copy due to export incompatibilities this change was not automatically flagged. 

**Additional context**
[Upstream change](https://github.com/open-telemetry/opentelemetry-collector/commit/6ab3c983bbfbda567f4372774abc7238680fb22f#diff-28b816700c2be73261e9ca6234cbf5615f045121cf28d4eff7ffe5899e4ed2ae)

bryan-aguilar:
**Description:** During the v0.16.0 an updated to the unexported `flags.go` file was not recognized and updated. This PR updates `flags.go` to match upstream. The unit tests had to be slightly modified. The `configflag` was reset after each test case so that no flag was left in memory between test runs. 

**Link to tracking Issue:** closes #951 

**Testing:** `go test ./...` `make test` `make build`

**Documentation:** n/a

### DIFF
--- a/pkg/config/config_factory.go
+++ b/pkg/config/config_factory.go
@@ -32,8 +32,8 @@ func GetConfigProvider() service.ConfigProvider {
 	loc := getConfigFlag()
 	if configContent, ok := os.LookupEnv(envKey); ok {
 		log.Printf("Reading AOT config from environment: %v\n", configContent)
-		loc = "env:" + envKey
+		loc = append(loc, "env:"+envKey)
 	}
 
-	return service.NewDefaultConfigProvider([]string{loc}, getSetFlag())
+	return service.NewDefaultConfigProvider(loc, getSetFlag())
 }
--- a/pkg/config/config_factory.go
+++ b/pkg/config/config_factory.go
@@ -32,7 +32,7 @@ func GetConfigProvider() service.ConfigProvider {
 	loc := getConfigFlag()
 	if configContent, ok := os.LookupEnv(envKey); ok {
 		log.Printf("Reading AOT config from environment: %v\n", configContent)
-		loc = append(loc, "env:"+envKey)
+		loc = []string{"env:"+envKey}
 	}
 
 	return service.NewDefaultConfigProvider(loc, getSetFlag())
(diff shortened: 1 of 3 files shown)


pkg/config/config_factory.go
```diff
@@ -32,8 +32,8 @@ func GetConfigProvider() service.ConfigProvider {
 	loc := getConfigFlag()
 	if configContent, ok := os.LookupEnv(envKey); ok {
 		log.Printf("Reading AOT config from environment: %v\n", configContent)
-		loc = "env:" + envKey
+		loc = append(loc, "env:"+envKey)
```
Aneurysm9, line 35:
```suggestion
		loc = []string{"env:"+envKey}
```

This should replace any value coming from the CLI flag for now.  This ensures compatibility with the prior behavior.  Now that we have the ability to merge multiple configurations we should work out how to transition to allowing supplemental configuration through the environment, but for now let's not break expectations.

pkg/config/config_factory_test.go
```diff
@@ -35,6 +35,10 @@ func TestGetCfgFactoryConfig(t *testing.T) {
 		Factories: factories,
 	}
 
+	resetFlag := func() {
+		configFlag = new(stringArrayValue)
```
Aneurysm9, line 39:
```suggestion
		configFlag = new(stringArrayValue)
		setFlag    = new(stringArrayValue)
```

Not sure it's used in the tests, but probably best to reset both of these here to avoid confusion if someone adds a test for the set flag later.

pkg/config/config_factory_test.go
```diff
@@ -49,6 +53,7 @@ func TestGetCfgFactoryConfig(t *testing.T) {
 		provider := GetConfigProvider()
 		_, err = provider.Get(context.Background(), factories)
 		require.Error(t, err)
+		resetFlag()
```
Aneurysm9, line 56: I would suggest doing this before the `require.Error()` call, possibly using [`t.Cleanup()`](https://pkg.go.dev/testing#T.Cleanup).  I believe that on failure `require.*` will prevent continuation of the test and thus this cleanup won't happen.

Aneurysm9 (approved):

jefchien:
While you're at it, can you update the tests in `config_factory_test.go` to use [`t.Setenv()`](https://pkg.go.dev/testing#T.Setenv) instead of the os one? Just for consistency.

jefchien (approved):

Issue #878, opened 2022-01-12
Minimal dockerfile that requires no build tool chain


sethAmazon:
**Is your feature request related to a problem? 
Need a minimal dockerfile that requires no build tool chain. This is for testing only

**Describe the solution you'd like**
Similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/examples/tracing/Dockerfile

**Describe alternatives you've considered**
Just use the make docker commands and require user to insall go tool chain

PR #956, opened 2022-02-01
Change Dockerfile to be multi-stage with build args.

jefchien:
**Description:** Mostly based on https://github.com/docker/cli/issues/1134#issuecomment-406449342. The idea is basically to have a build stage and a copy stage. Only one of them will run based on the `BUILDMODE` build arg. By default, the steps will build the binary (takes forever to download the go mod) and skip the copy stage. For our CI and Makefile, we will pass in `BUILDMODE=copy`, which will skip the build stage.

**Link to tracking Issue:** Resolves https://github.com/aws-observability/aws-otel-collector/issues/878

**Testing:** Ran `make docker-build` and `docker buildx build --platform linux/amd64 --build-arg BUILDMODE=build --load -t awscollector:test -f ./cmd/awscollector/Dockerfile .`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,11 +115,11 @@ package-deb: build
 
 .PHONY: docker-build
 docker-build: amd64-build
-	docker buildx build --platform linux/amd64 -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
+	docker buildx build --platform linux/amd64 --build-arg BUILDMODE=copy --load -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
 
 .PHONY: docker-build-arm
 docker-build-arm: arm64-build
-	docker buildx build --platform linux/arm64 -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
+	docker buildx build --platform linux/arm64 --build-arg BUILDMODE=copy --load -t $(DOCKER_NAMESPACE)/$(COMPONENT):$(VERSION) -f ./cmd/$(COMPONENT)/Dockerfile .
 
 .PHONY: docker-push
 docker-push:
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -1,11 +1,48 @@
+# The build mode (options: build, copy) passed in as a --build-arg. If build is specified, then the copy
+# stage will be skipped and vice versa. The build mode builds the binary from the source files, while
+# the copy mode copies in a pre-built binary.
+ARG BUILDMODE=build
+
 ################################
-#	Building Stage         #	
+#	Certificate Stage      #
+#			       #
+################################
+FROM alpine:latest AS certs
+
+RUN apk --update add ca-certificates
+
+################################
+#	Build Stage            #
+#			       #
+################################
+FROM golang:1.17 AS prep-build
+
+ARG TARGETARCH
+
+# pass in the GOPROXY as a --build-arg (e.g. --build-arg GOPROXY=direct)
+ARG GOPROXY
+ENV GOPROXY=${GOPROXY}
+
+# download go modules ahead to speed up the building
+WORKDIR /workspace
+COPY go.mod .
+COPY go.sum .
+RUN go mod download -x
+
+# copy source
+COPY . .
+
+# build
+RUN make ${TARGETARCH}-build
+
+# move
+RUN mv /workspace/build/linux/$TARGETARCH/aoc /workspace/awscollector
+
+################################
+#	Copy Stage             #
 #			       #	
 ################################
-FROM alpine:latest as build 
-
-# update cert
-RUN apk --update add ca-certificates
+FROM scratch AS prep-copy
 
 WORKDIR /workspace
 
@@ -14,6 +51,13 @@ ARG TARGETARCH
 # copy artifacts
 # always assume binary is created
 COPY build/linux/$TARGETARCH/aoc /workspace/awscollector
+
+################################
+#	Packing Stage          #
+#			       #
+################################
+FROM prep-${BUILDMODE} AS package
+
 COPY config.yaml /workspace/config/otel-config.yaml
 COPY config/ /workspace/config/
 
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -99,6 +99,10 @@ arm64-build: install-tools lint multimod-verify
 amd64-build-only:
 	GOOS=linux GOARCH=amd64 $(GOBUILD) $(LDFLAGS) -o ./build/linux/amd64/aoc ./cmd/awscollector
 
+.PHONY: amd64-build-only
```
sethAmazon, line 102: amd with arm?

cmd/awscollector/Dockerfile
```diff
@@ -1,19 +1,59 @@
...
+
+ENV GOPROXY=direct
+
+ARG TARGETARCH
+
+# download go modules ahead to speed up the building
+WORKDIR /workspace
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
```
sethAmazon, line 26: Will this take a long time. @anuraaga suggestion was to move this another dockerfile.

sethAmazon: Have 1 for coping the binary and 1 where you can build the binary inside of the build container

cmd/awscollector/Dockerfile
```diff
@@ -1,19 +1,59 @@
...
-# update cert
 RUN apk --update add ca-certificates
 
+################################
+#	Building Stage         #
+#			       #
+################################
+FROM golang:1.17 AS prep-build
+
+ENV GOPROXY=direct
```
sethAmazon, line 18: Not really suppose to have this since it is only required by amazon employees but idc you can keep it

Makefile
jefchien: My bad.

cmd/awscollector/Dockerfile
jefchien: I think it's fine to keep it.

cmd/awscollector/Dockerfile
jefchien: This is just a way to have both in a single Dockerfile. We typically won't use the build option, so the stage will be skipped, but it's there for those that want it.

cmd/awscollector/Dockerfile
Aneurysm9: I think this is a significant contributor to the prolonged runtime of the build stages.  Is it possible to extract this to a runtime argument as well so that it can be used only when necessary?

jefchien: Yeah can do. I guess we'd have to use it, but others wouldn't.

cmd/awscollector/Dockerfile
```diff
@@ -1,19 +1,59 @@
...
+WORKDIR /workspace
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+# copy source
+COPY . .
+
+# build
+RUN make ${TARGETARCH}-build-only
```
Aneurysm9, line 32:
```suggestion
RUN make ${TARGETARCH}-build
```

Not sure we should encourage skipping the lint checks.  If that's really necessary then a binary can be built externally and copied in.

cmd/awscollector/Dockerfile
```diff
@@ -1,19 +1,59 @@
...
 
 # copy artifacts
 # always assume binary is created
 COPY build/linux/$TARGETARCH/aoc /workspace/awscollector
+
+################################
+#	Packing Stage          #
+#			       #
+################################
+FROM prep-${BIN_SRC_TYPE} AS package
```
Aneurysm9, line 55: Is this stage necessary?  Does this not work?

```DOCKERFILE
COPY --from=${BIN_SRC_TYPE} /workspace/awscollector /awscollector
COPY config.yaml /etc/otel-config.yaml
COPY config/ /etc/
```

cmd/awscollector/Dockerfile
Aneurysm9: Some documentation/description of that in this file might be good.  It's a bit opaque.

cmd/awscollector/Dockerfile
jefchien: It seems necessary and that does not work.
```
Dockerfile:72
--------------------
  70 |     
  71 |     COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
  72 | >>> COPY --from="prep-${BIN_SRC_TYPE}" /workspace/awscollector /awscollector
  73 |     COPY config.yaml /workspace/config/otel-config.yaml
  74 |     COPY config/ /workspace/config/
--------------------
error: failed to solve: failed to parse stage name "prep-${BIN_SRC_TYPE}": invalid reference format: repository name must be lowercase
```

Aneurysm9: booooooooooooooooo.

cmd/awscollector/Dockerfile
jefchien: Makes sense. Changed it back to `${TARGETARCH}-build`.

cmd/awscollector/Dockerfile
jefchien: Added it as an optional build arg with a comment above the declaration. We'll have to include `--build-arg GOPROXY=direct` to be able to build it locally.

cmd/awscollector/Dockerfile
jefchien: Added some documentation at the top of the Dockerfile where the ARG is declared.

Aneurysm9 (approved):

cmd/awscollector/Dockerfile
```diff
@@ -1,19 +1,63 @@
+# The build mode (options: build, copy) passed in as a --build-arg. If build is specified, then the copy
+# stage will be skipped and vice versa. The build mode builds the binary from the source files, while
+# the copy mode copies in a pre-built binary.
+ARG BUILDMODE=build
```
khanhntd, line 4: Should this be default as COPY for integration tests on CI? For docker-build and docker-build-arm should it be BUILDMODE=build (for first time user and it would be cache the layers after the first build)

jefchien: That could work too.

anuraaga (approved):
Nice pattern!

cmd/awscollector/Dockerfile
anuraaga: Oops had missed this comment - think the default should be build so the Dockerfile can be built and run easily by external tools, e.g., `docker run https://github.com/aws-observability/aws-otel-collector`

copy is an interaction with our Makefile but may as well make the docker default as "docker-native" as possible

jefchien: Oh didn't know you could do that. Doesn't look like `docker run https://github.com/aws-observability/aws-otel-collector` is valid, but I see your point.

bryan-aguilar (approved):

PR #978, opened 2022-02-17
CI Workflow for ADOT Operator releases

vasireddy99:
**Description:** 

This PR creates a seperate CI Workflow for ADOT operator releases. It included e2e tests for `adot-operator` 

Second Step : The next step is to eliminate the adot-operator tests from the[ CI workflow](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CI.yml#L1091)

*In-Progress*

 *In-Progress*

### DIFF
--- a/.github/workflows/CI-adot-operator.yml
+++ b/.github/workflows/CI-adot-operator.yml
@@ -15,22 +15,13 @@ name: C/I ADOT Operator
 on:
   push:
     branches:
-      - main
-      - release/v*
-      - dev
-      - test/*
+
     paths-ignore:
       - '.github/**'
       - '!.github/workflows/CI-adot-operator.yml'
       - '**.md'
 
-  # from collector and contrib repo
-  repository_dispatch:
-    types: [dependency-build, workflow-run]
-
 env:
-  IMAGE_NAME: aws-otel-collector
-  ECR_REPO: aws-otel-test/adot-collector-integration-test
   TF_VAR_aws_access_key_id: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
   TF_VAR_aws_secret_access_key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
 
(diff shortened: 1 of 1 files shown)


jefchien:
What is this for? It looks like it's redoing a lot of the steps in the CI.

vasireddy99:
> What is this for? It looks like it's redoing a lot of the steps in the CI.
This PR separates CI workflow for the ADOT operator, in process of decoupling the ADOT operator releases from the ADOT collector releases. I agree, yes the steps are referenced form CI.yml

jefchien:
I don't see the benefit of separating this part out of the CI. Decoupling the release should only involve changes to the CD. If the operator mirroring is being moved out of this repo, then it makes sense to have a separate CI for it in whatever repo it ends up in. Right now, this workflow doesn't seem to do anything different to justify this change.

bryan-aguilar:
> 



> I don't see the benefit of separating this part out of the CI. Decoupling the release should only involve changes to the CD. If the operator mirroring is being moved out of this repo, then it makes sense to have a separate CI for it in whatever repo it ends up in. Right now, this workflow doesn't seem to do anything different to justify this change.

My question is, does the collector CI need to rely on the operator tests passing when the ADOT CD mirroring is decoupled. If the answer is no then I believe that justifies separating them.

jefchien:
Even if the operator mirroring gets removed from this repo, the operator tests should still be done. They have nothing to do with the mirroring. That's done in the CD, which is after all these tests are run. It's testing against the current operator image.

vasireddy99:
Converting to Draft.

alolita:
Based on the asynchronous nature of upstream releases of the Collector and Operator, here are my thoughts and recommendations. There are two types of workflows for releasing the ADOT Operator.

1. **Standard** workflow to release adot-operator as a fast-follow on the latest ADOT Collector aws-otel-collector tested and released. E.g. adot-operator v 0.45 released in sync w aws-otel-collector v 0.17 (mapped to OTEL 0.45). This workflow would be the default release workflow for publishing both images synchronously.

2. **Ad-hoc** workflow to run adot-operator specific tests and released for a particular release of aws-otel-collector. E.g. adot-operator v 0.44 released after passing tests with aws-otel-collector v 0.44. This workflow would be run whenever an out-of-band release is needed for the ADOT Operator. This workflow would be run manually and would require an explicit collector release to be tested with.

Thoughts? @erichsueh3 @vasireddy99 @bryan-aguilar

anuraaga:
For reference, here's a chat I had with @erichsueh3 

To sum up, the operator is two components, the operator docker image (and supporting) and the helm charts to deploy them, upstream releases both at the same time. Similarly, I’d expect us to need to mirror the docker images and also distribute a helm chart pointing to the mirrored images to call it an “ADOT operator”. It would make sense for these two be in the same repo since they need to be released together. I’d expect that to be an aws-otel-operator repo really.

Also another thing to keep in mind is that the aws-otel-collector itself only needs to be tested on k8s once, the operator is just an abstraction of deployment to k8s. Since we already have k8s tests in the collector repo, I would remove the operator tests from it. You test the operator when updating the operator, not the collector.

The benefits to us / users is the aws-otel-collector CI becomes a bit less unwieldy so easier to work in (again, maybe it’s too unwieldy to matter) and to be able to release the operator images / helm chart together (assuming we are releasing helm charts which I would be surprised if we don’t)

It’s a lot like how we have a separate repo for the lambda layers since they use the collector, and the lambda layers aren’t part of the collector CI.

The design of the ADOT operator itself seems to be missing a doc

bryan-aguilar:
Is this necessary anymore since there is now a separate workflow and the operator release is toggleable in the CD workflow?

vasireddy99:
> Is this necessary anymore since there is now a separate workflow and the operator release is toggleable in the CD workflow?

Closing this PR since it is not necessary

PR #980, opened 2022-02-18
Separate CD for ADOT Operator mirroring

erichsueh3:
This PR adds a separate CD Workflow for ADOT Operator mirroring.

**Testing:**
In Progress

### DIFF
--- a/.github/workflows/CD-adot-operator.yml
+++ b/.github/workflows/CD-adot-operator.yml
@@ -1,0 +1,42 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+name: C/D ADOT Operator
+
+on: [workflow_call, workflow_dispatch]
+
+jobs:           
+  release-latest-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go 1.x
+        if: steps.release-latest-image.outputs.cache-hit != 'true'
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+
+      - name: Configure AWS Credentials
+        if: steps.release-latest-image.outputs.cache-hit != 'true'
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.RELEASE_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.RELEASE_SECRET }}
+          aws-region: us-west-2
+
+      - name: Mirror ADOT operator
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true' }}
+        run: cd tools/release/adot-operator-images-mirror && go run ./
+        env:
+          AWS_SDK_LOAD_CONFIG: true
--- a/.github/workflows/CD-adot-operator.yml
+++ b/.github/workflows/CD-adot-operator.yml
@@ -11,7 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-name: C/D ADOT Operator
+name: Mirror ADOT Operator
 
 on: [workflow_call, workflow_dispatch]
 
@@ -36,7 +36,6 @@ jobs:
           aws-region: us-west-2
 
       - name: Mirror ADOT operator
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true' }}
         run: cd tools/release/adot-operator-images-mirror && go run ./
         env:
           AWS_SDK_LOAD_CONFIG: true
--- a/.github/workflows/CD-adot-operator.yml
+++ b/.github/workflows/CD-adot-operator.yml
@@ -13,7 +13,7 @@
 
 name: Mirror ADOT Operator
 
-on: [workflow_call, workflow_dispatch]
+on: workflow_dispatch
 
 jobs:           
   release-latest-image:
--- a/.github/workflows/CD-adot-operator.yml
+++ b/.github/workflows/CD-adot-operator.yml
@@ -13,7 +13,12 @@
 
 name: Mirror ADOT Operator
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'the desired version number of the operator (ex. 0.45.0)'
+        required: true
 
 jobs:           
   release-latest-image:
@@ -22,20 +27,23 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set up Go 1.x
-        if: steps.release-latest-image.outputs.cache-hit != 'true'
         uses: actions/setup-go@v2
         with:
           go-version: 1.17
 
       - name: Configure AWS Credentials
-        if: steps.release-latest-image.outputs.cache-hit != 'true'
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-access-key-id: ${{ secrets.RELEASE_KEY_ID }}
           aws-secret-access-key: ${{ secrets.RELEASE_SECRET }}
           aws-region: us-west-2
 
+      - name: Version checking
+        id: version
+        run: VERSION="${{ github.event.inputs.version }}" bash tools/workflow/version-check.sh
+
       - name: Mirror ADOT operator
+        if: ${{ steps.version.outputs.update-operator == 'true' }}
         run: cd tools/release/adot-operator-images-mirror && go run ./
         env:
           AWS_SDK_LOAD_CONFIG: true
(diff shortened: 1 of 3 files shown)


erichsueh3:
Couple of notes:
1. Not sure if we want this to be able to be called manually only or also be able to be called when the main CD workflow is called
2. Not sure if setting up Go 1.x is necessary
3. Will eventually also remove this workflow from the main CD workflow
4. Can anyone provide insight into how to "test" this, as all it does is mirror the ADOT Operator image to public ECR?

bryan-aguilar:
In response to your comments
1. This should be able to be triggered by a dispatch. It may also be valid to trigger this during the Collector CD workflow also. Although I think it may be better to keep them as isolated CI/CD workflows that do not depend on each other. 
2. It is necessary to properly setup the environment before the `go run` is triggered. 
3. Agreed
4. Lets talk tomorrow. This is one of the issues that I am trying to solve at the moment.

bryan-aguilar:
Will this be published as a github release and require a tag? See this [job](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L642) that we currently use. Does this need to be applied here also? If release notes are not required then we can possibly trim that job down to just [this](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L654) step.

anuraaga:
We might also consider removing the existing charts there and renaming the repo to aws-otel-operator. IMO it seems a bit much to maintain charts in addition to the repo, documenting it all up to ask customers to choose makes things more complicated for them. Just having the operator as the canonical way of deploying ADOT to k8s will simplify the story a lot I think.

.github/workflows/CD-adot-operator.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Configure AWS Credentials
+        if: steps.release-latest-image.outputs.cache-hit != 'true'
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.RELEASE_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.RELEASE_SECRET }}
+          aws-region: us-west-2
+
+      - name: Mirror ADOT operator
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true' }}
```
jefchien, line 39: You are no longer running the version checking script and don't have a cache, so these conditions aren't needed.

jefchien:
@erichsueh3 You can test it in your fork by setting your own credentials as the `RELEASE_KEY_ID` and `RELEASE_SECRET`. After running it, you should see the two repos mirrored in your ECR.

If it's still going to be called by the CD and only manually run sometimes, then it makes sense to keep it in this repo. Otherwise, I agree with @anuraaga.

.github/workflows/CD-adot-operator.yml
```diff
@@ -0,0 +1,42 @@
...
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+name: C/D ADOT Operator
```
jefchien, line 14:
```suggestion
name: Mirror ADOT Operator
```

bryan-aguilar:
> Since the operator docker images are basically an implementation detail of the helm charts, which are the actual user-facing thing, should this mirroring happen in that repo? I think putting it here makes this repo's CI even more confusing (perhaps we've passed the cliff where it doesn't matter anymore though :P)
> 
> https://github.com/aws-observability/aws-otel-helm-charts

> We might also consider removing the existing charts there and renaming the repo to aws-otel-operator. IMO it seems a bit much to maintain collector-only deployment charts in addition to the operator, documenting it all up to ask customers to choose makes things more complicated for them. Just having the operator as the canonical way of deploying ADOT to k8s will simplify the story a lot I think.

I think this is a really good idea to explore. I had briefly thought that these operator actions should exist in a repo on their own. I think the pro/cons of moving these actions should be weighed before making a final decision to merge this. I think as a first step isolating operator CI/CD into it's own repository would make sense. A future action of simplifying existing charts should be the second step.

erichsueh3:
> Will this be published as a github release and require a tag? See this [job](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L642) that we currently use. Does this need to be applied here also? If release notes are not required then we can possibly trim that job down to just [this](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CD.yml#L654) step.

Don't think that we need to publish this as a release, but if anyone feels strongly the other way then we can

erichsueh3:
> > Since the operator docker images are basically an implementation detail of the helm charts, which are the actual user-facing thing, should this mirroring happen in that repo? I think putting it here makes this repo's CI even more confusing (perhaps we've passed the cliff where it doesn't matter anymore though :P)
> > https://github.com/aws-observability/aws-otel-helm-charts
> 
> > We might also consider removing the existing charts there and renaming the repo to aws-otel-operator. IMO it seems a bit much to maintain collector-only deployment charts in addition to the operator, documenting it all up to ask customers to choose makes things more complicated for them. Just having the operator as the canonical way of deploying ADOT to k8s will simplify the story a lot I think.
> 
> I think this is a really good idea to explore. I had briefly thought that these operator actions should exist in a repo on their own. I think the pro/cons of moving these actions should be weighed before making a final decision to merge this. I think as a first step isolating operator CI/CD into it's own repository would make sense. A future action of simplifying existing charts should be the second step.

I agree with both  of you, having the Operator CI/CD in a different repository would make a lot of sense. We can document the pros and cons of this, and also document what that process might look like, and have a discussion early next week about it before making any decisions

erichsueh3:
Update on this PR: the workflow (and the script to check if the image should be mirrored) work as intended. However, there is now the issue within the ADOT Operator mirroring tool itself; sometime in January, OTEL Operator images [stopped being mirrored](https://github.com/open-telemetry/opentelemetry-operator/pull/661/files) to quay.io, which is where the ADOT Operator is mirrored from. This means that both the release workflow and this new ad hoc workflow will not mirror images anymore, and there is now an extra step to change the tool to mirror from ghcr instead of quay which I am working on now.

bryan-aguilar:
Hi @erichsueh3,

That's an interesting find relating to the mirroring issue. I went ahead and converted this to a draft while the work is still in progress. Feel free to toggle it back when it is ready for review.

erichsueh3:
> Hi @erichsueh3,
> 
> That's an interesting find relating to the mirroring issue. I went ahead and converted this to a draft while the work is still in progress. Feel free to toggle it back when it is ready for review.

Hey @bryan-aguilar, sounds good. Will work on a separate PR for the mirroring tools

.github/workflows/CD-adot-operator.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:           
+  release-latest-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
```
bryan-aguilar, line 32:
```suggestion
          go-version: ^1.17.7
```

.github/workflows/CD-adot-operator.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.RELEASE_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.RELEASE_SECRET }}
+          aws-region: us-west-2
+
+      - name: Version checking
+        id: version
+        run: VERSION="${{ github.event.inputs.version }}" bash tools/workflow/version-check.sh
```
bryan-aguilar, line 43: Is this collector or operator version? Does this need to go into the CD workflow?

erichsueh3: This is the Operator version, and I don't think it needs to go in the CD workflow since it should only run on release and also checks collector version; also this should be `tools/release/adot-operator-images-mirror/version-check.sh`, fixed that just now.

.github/workflows/CD-adot-operator.yml
erichsueh3: Fixed

.github/workflows/CD.yml
```diff
@@ -634,7 +638,7 @@ jobs:
           aws-region: us-west-2
 
       - name: Mirror ADOT operator
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true' }}
+        if: ${{ github.event.inputs.skip-adot-operator-mirroring == 'false' && (steps.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true') }}
```
bryan-aguilar, line 641:
```suggestion
        if: ${{ github.event.inputs.skip-adot-operator-mirroring == 'false' && (steps.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.patch-update == 'false') }}
```

Can you verify that this logic is the same by referencing [this](https://github.com/aws-observability/aws-otel-collector/blob/main/tools/workflow/docker-version-compare.sh) script.

erichsueh3: Not sure why it checks for `steps.version.outputs.same-version == 'true'` so that should be removed; checking only `steps.version.outputs.patch-update == 'false'` doesn't account for if major/minor updates have happened. If we check it as `steps.version.outputs.patch-update == 'true'` though it should be equivalent, while also adding a check for patch updates (if we want that in adot operator images)

bryan-aguilar: I believe only `xxx-update` values can be true at once. I believe `any-update` can also be true but that is not referenced here.

bryan-aguilar:
The plan is to merge this PR now and investigate moving the ADOT operator releases to a separate repository in the futures. Merging this now will unblock the ADOT Operator being out of sync with upstream.

bryan-aguilar (approved):

PR #1001, opened 2022-02-28
Change canary tests to run every 4 hours

bryan-aguilar:
**Description:** Canary tests are taking longer than an hour but being queued on the hour. Change to run once every 4 hours.

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -14,7 +14,7 @@
 name: 'Canary Test'
 on:
   schedule:
-    - cron: '0 * * * *' # triggers the workflow every hour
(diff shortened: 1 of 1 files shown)


Aneurysm9 (approved):

vasireddy99 (approved):

anuraaga:
Do we also want to add a concurrency limit of 1?

https://docs.github.com/en/actions/using-jobs/using-concurrency

bryan-aguilar:
> Do we also want to add a concurrency limit of 1?
> 
> https://docs.github.com/en/actions/using-jobs/using-concurrency

That's a good suggestion, added.

PR #1012, opened 2022-03-05
Build and use dbotconf

bryan-aguilar:
**Description:** This PR adds two new make commands to utilize [`dbotconf`](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/dbotconf) located in [`opentelemetry-go-build-tools`](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/dbotconf). 
Adds 
`make dependabot-check`
`make dependabot-generate`

Creates a new make target of `install-dbotconf` so that it is only installed if not present. Improving the makefile so this logic is present for all tools should be tracked in another issue. 

**Testing:** Tested locally see new `dependabot.yml`

There is a bug currently where the `actions` label is being applied incorrectly but a [PR is already made upstream](https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/67) to fix this.


Makefile
```diff
@@ -204,6 +216,13 @@ install-tools:
 	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install honnef.co/go/tools/cmd/staticcheck@v0.2.0
 	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.0
 	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install mvdan.cc/sh/v3/cmd/shfmt@latest
+	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install go.opentelemetry.io/build-tools/dbotconf
+
+.PHONY: install-dbotconf
+install-dbotconf:
```
Aneurysm9, line 222: Is this needed?  Shouldn't it be covered by `install-tools`?

tools/linters/go.mod
```diff
@@ -4,9 +4,11 @@ go 1.17
 
 require (
 	github.com/golangci/golangci-lint v1.39.0
+	go.opentelemetry.io/build-tools/dbotconf v0.0.0-20220304232821-55b850922377
 	go.opentelemetry.io/build-tools/multimod v0.0.0-20210920164323-2ceabab23375
```
Aneurysm9, line 8: I'd suggest updating multimod to the same pseudoversion.  I don't think anything in their common dependencies will have broken multimod at this version if updated to the dbotconf version, but it's a possibility if they diverge.

bryan-aguilar: Thank you, that's a good point. Updated.

Makefile
Aneurysm9: I think I prefer the approach taken by the [opentelemetry-go repo](https://github.com/open-telemetry/opentelemetry-go/blob/main/Makefile#L31-L72).  This builds the specific version of the tool required in an ignored directory directly in the repo, which also allows for only building the tools that are needed.  I think it has the same issue with not detecting updates, but I think that can be fixed by depending on `internal/tools/go.sum`.

bryan-aguilar: I can see how this makes more sense and think we should adapt this into our process. Do you believe this is a blocker for this PR? We could open a new issue to track this improvement for a separate PR.

Aneurysm9: Let's tackle that on a separate PR.

Aneurysm9 (approved):

Issue #1006, opened 2022-03-03
Separate operator mirror into it's own job. 


bryan-aguilar:
I believe this step ([Mirror ADOT Operator](https://github.com/aws-observability/aws-otel-collector/blob/ba4622c9611e15b059e915bb01d85caa011615be/.github/workflows/CD.yml#L636)) should exist as it's own job in the CD workflow. This would allow it to fail in isolation and the publishing of other images success can be cached. This will allow the CD workflow to be safely re-ran in case the operator mirror fails in the future.

bryan-aguilar:
Closed with #980

PR #1032, opened 2022-03-08
Move ADOT Operator mirroring to own job in CD workflow

erichsueh3:
**Description:**
This PR moves the ADOT Operator mirroring step out of the `release-latest-image` job into its own `mirror-adot-operator` job in the CD workflow

**Link to tracking Issue:** 
#1006

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -631,24 +631,32 @@ jobs:
 
   mirror_adot_operator:
     runs-on: ubuntu-latest
-    needs: release-latest-image
+    needs: release-checking
     steps:
       - uses: actions/checkout@v2
 
+      - name: Cache if success
+        id: mirror_adot_operator
+        uses: actions/cache@v2
+        with:
+          key: mirror_adot_operator-${{ github.run_id }}
+          path: VERSION
+
       - name: Set up Go 1.x
+        if: steps.mirror_adot_operator.outputs.cache-hit != 'true'
         uses: actions/setup-go@v2
         with:
           go-version: ^1.17.7
 
       - name: Compare version with Dockerhub latest
         id: version
-        if: needs.release-latest-image.outputs.cache-hit != 'true'
+        if: steps.mirror_adot_operator.outputs.cache-hit != 'true'
         run: |
           TAG="${{ needs.release-checking.outputs.version }}"
           TARGET_VERSION=$TAG bash tools/workflow/docker-version-compare.sh
 
       - name: Configure AWS Credentials
-        if: needs.release-latest-image.outputs.cache-hit != 'true'
+        if: steps.mirror_adot_operator.outputs.cache-hit != 'true'
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-access-key-id: ${{ secrets.RELEASE_KEY_ID }}
@@ -656,7 +664,7 @@ jobs:
           aws-region: us-west-2
 
       - name: Mirror ADOT operator
-        if: ${{ github.event.inputs.skip-adot-operator-mirroring == 'false' && (needs.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true') }}
+        if: ${{ github.event.inputs.skip-adot-operator-mirroring == 'false' && (steps.mirror_adot_operator.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true') }}
         run: cd tools/release/adot-operator-images-mirror && go run ./
         env:
           AWS_SDK_LOAD_CONFIG: true
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -629,34 +629,34 @@ jobs:
           src: public.ecr.aws/${{ env.ECR_REPO }}:${{ needs.release-checking.outputs.version }}
           dst: ${{ env.IMAGE_NAMESPACE }}/${{ env.IMAGE_NAME }}:${{ needs.release-checking.outputs.version }}
 
-  mirror_adot_operator:
+  mirror-adot-operator:
     runs-on: ubuntu-latest
     needs: release-checking
     steps:
       - uses: actions/checkout@v2
 
       - name: Cache if success
-        id: mirror_adot_operator
+        id: mirror-adot-operator
         uses: actions/cache@v2
         with:
-          key: mirror_adot_operator-${{ github.run_id }}
+          key: mirror-adot-operator-${{ github.run_id }}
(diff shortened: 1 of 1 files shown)


erichsueh3:
Unsure if we need the checks for `release-latest-image.outputs.cache-hit != 'true'` if we already require that the `release-latest-image` job runs before this one

.github/workflows/CD.yml
```diff
@@ -629,16 +629,34 @@ jobs:
...
+      - name: Compare version with Dockerhub latest
+        id: version
+        if: needs.release-latest-image.outputs.cache-hit != 'true'
+        run: |
+          TAG="${{ needs.release-checking.outputs.version }}"
+          TARGET_VERSION=$TAG bash tools/workflow/docker-version-compare.sh
+
       - name: Configure AWS Credentials
-        if: steps.release-latest-image.outputs.cache-hit != 'true'
+        if: needs.release-latest-image.outputs.cache-hit != 'true'
```
bryan-aguilar, line 651: It does not look like you export this from the `release-latest-image` job.

.github/workflows/CD.yml
```diff
@@ -629,16 +629,34 @@ jobs:
...
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-access-key-id: ${{ secrets.RELEASE_KEY_ID }}
           aws-secret-access-key: ${{ secrets.RELEASE_SECRET }}
           aws-region: us-west-2
 
       - name: Mirror ADOT operator
-        if: ${{ github.event.inputs.skip-adot-operator-mirroring == 'false' && (steps.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true') }}
+        if: ${{ github.event.inputs.skip-adot-operator-mirroring == 'false' && (needs.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true') }}
```
bryan-aguilar, line 659: Same issue with this cache hit. I know you did not write this but can you explain the conditional logic behind the version comparison? Should the operator be bound to only be mirrored during major/minor or same collector version CD runs? Why do we not do it on patch releases? 

This may have been answered in the previous PR but is running the mirror script when ADOT Operator is the same version as Upstream Operator. The `docker-version-compare` compares the collector versions and I am wondering how that is related to operator versions. 

I suspect this could be tech debt and we should fully understand the reasoning for it before moving forward.

bryan-aguilar (commented):
Can you please outline how you tested these changes?

.github/workflows/CD.yml
```diff
@@ -629,16 +629,34 @@ jobs:
           src: public.ecr.aws/${{ env.ECR_REPO }}:${{ needs.release-checking.outputs.version }}
           dst: ${{ env.IMAGE_NAMESPACE }}/${{ env.IMAGE_NAME }}:${{ needs.release-checking.outputs.version }}
 
+  mirror_adot_operator:
+    runs-on: ubuntu-latest
+    needs: release-latest-image
```
bryan-aguilar, line 634: Why does this job require the `release-latest-image` job to be successful?

bryan-aguilar (commented):
We should also be caching this job if it's successful. I believe that should answer your cache question. Review the cache step in `release-latest-image`

.github/workflows/CD.yml
erichsueh3: Added cache step and changed conditional to check against that instead

.github/workflows/CD.yml
erichsueh3: Looks like it doesn't, but I've added that it needs `release-checking` to be successful instead as its output is needed to compare the docker version.

.github/workflows/CD.yml
erichsueh3: I don't understand the conditional logic either, I think it would make sense for the operator to be mirrored on major/minor/patch releases and NOT same Collector version CD runs. The `docker-version-compare` DOES compare collector versions, but if the idea is that if the ADOT Operator releases are still tied to ADOT Collector releases (+ the upstream OTEL Operator.Collector are in sync), then a check for collector version is still the same. Either way, when the ADOT Operator is already the same version as the upstream, no image ends up being mirrored in the end regardless

erichsueh3: Just changed the `same-version` check to a `patch-update` check as well

erichsueh3:
> Can you please outline how you tested these changes?

The changes were made on a branch of my own fork of the collector, with mirroring done to my own registry. The CD workflow was modified and stripped down to only contain what was needed to test the change (essentially the `release-checking` job and `mirror-adot-operator` job), see [here](https://github.com/erichsueh3/aws-otel-collector-1/actions/runs/1953625433) for a successful CD workflow run with the `mirror-adot-operator` now as a separate job

.github/workflows/CD.yml
```diff
@@ -629,16 +629,42 @@ jobs:
...
+        if: steps.mirror-adot-operator.outputs.cache-hit != 'true'
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-access-key-id: ${{ secrets.RELEASE_KEY_ID }}
           aws-secret-access-key: ${{ secrets.RELEASE_SECRET }}
           aws-region: us-west-2
 
       - name: Mirror ADOT operator
-        if: ${{ github.event.inputs.skip-adot-operator-mirroring == 'false' && (steps.release-latest-image.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.same-version == 'true') }}
+        if: ${{ github.event.inputs.skip-adot-operator-mirroring == 'false' && (steps.mirror-adot-operator.outputs.cache-hit != 'true' && (steps.version.outputs.major-update == 'true' || steps.version.outputs.minor-update == 'true') || steps.version.outputs.patch-update == 'true') }}
```
bryan-aguilar, line 667: Why is the ADOT Operator release tied to the collector version release anyways? I think we could just get rid of the whole version check anyways. We should maybe see if anyone else has input on this. 

Is there a situation, other than when it is explicitly disabled, that we do not want to mirror the ADOT operator?

erichsueh3: I'm okay with getting rid of the version check; the tool, when run, will not re-mirror existing Operator images regardless, but yeah some outside input might be helpful. The ADOT Operator release is tied to the Collector release because we would want to use the same version of the Operator/Collector (i.e 0.17 ADOT Collector (0.45 OTEL Collector) + 0.45 ADOT Operator)

There isn't really a situation where we would NOT want to mirror the ADOT Operator, especially when considering that it doesn't re-mirror images.

erichsueh3:
Closing this PR, instead of moving the ADOT Operator mirroring to its own job in the CD workflow, we are instead opting to remove it from the CD workflow entirely.

PR #1042, opened 2022-03-11
Update `timeout_minutes` for canary tests

vasireddy99:
**Description:** 

This PR updates `timeout_minutes` for the canary tests. Current canary tests are failing due to timeout issues. Increasing the timeout minutes will give enough time to run the tests.

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -131,7 +131,7 @@ jobs:
         uses: nick-invision/retry@v2
         with:
           max_attempts: 3
-          timeout_minutes: 60
+          timeout_minutes: 120
           retry_on: error
           retry_wait_seconds: 5
           command: |


bryan-aguilar:
Why are these tests suddenly taking 20 minutes more than there last three successful runs? I checked the `otlp_trace,canary linux` completion times and they were sub 50 minutes. I am not sure continuing to bump up the timeout time is the correct answer.

bryan-aguilar:
Does this timeout_duration apply to each separate job since it's a matrix configuration or the totality of all the jobs created under it? I could foresee that being a problem.

vasireddy99:
> Does this timeout_duration apply to each separate job since it's a matrix configuration or the totality of all the jobs created under it? I could foresee that being a problem.
> 
> Still though, filtering successful runs of the canary tests (seen [here](https://github.com/aws-observability/aws-otel-collector/actions/workflows/canary.yml?query=is%3Asuccess++)) are all sub 1hr (except 1) completion times. What has caused the time to increase so much? It is interesting through, if you go back to 9 days ago on that query you can see a batch of tests that took greater than 1hr to complete.

yeah, the tests took greater than 1 hr to complete previously as you mentioned and that's one of the reason we modified to 4 hours, if am not wrong. 
So from my understanding, there might be few cases such as,
 * Canary tests are passed only if after successful validation for 20 times. if one of the test case fail they. will go to the retry attempts.
 * Possibility of available GitHub runners (not exactly sure how that works), similar to what we experienced in last week.

vasireddy99:
My understanding is with these tests are failing when add 
* https://github.com/aws-observability/aws-otel-collector/pull/1035
the `timeout_minutes` from when its hitting 3600000ms it is failing with that error. 
Although the validation took only[ 32  minutes](https://github.com/aws-observability/aws-otel-collector/runs/5513633755?check_suite_focus=true#step:8:1776) but the job `Run canary test` exceeded the time.

bryan-aguilar:
> > Does this timeout_duration apply to each separate job since it's a matrix configuration or the totality of all the jobs created under it? I could foresee that being a problem.
> > Still though, filtering successful runs of the canary tests (seen [here](https://github.com/aws-observability/aws-otel-collector/actions/workflows/canary.yml?query=is%3Asuccess++)) are all sub 1hr (except 1) completion times. What has caused the time to increase so much? It is interesting through, if you go back to 9 days ago on that query you can see a batch of tests that took greater than 1hr to complete.
> 
> yeah, the tests took greater than 1 hr to complete previously as you mentioned and that's one of the reason we modified to 4 hours, if am not wrong. So from my understanding, there might be few cases such as,
> 
>     * Canary tests are passed only if after successful validation for 20 times. if one of the test case fail they. will go to the retry attempts.
> 
>     * Possibility of available GitHub runners (not exactly sure how that works), similar to what we experienced in last week.
> 
>     * Possibility of taking extra time when creating the aws resources.

Now that you mentioned it I think it could be a safe assumption that the batch of > 1hr tests were due to a lack of github runners because we were running CI continuously. So that makes a lot of sense.

It sounds like you are saying that the `timeout_duration` applies to the entire workflow even though it is only added to one job. Is this due to the workflow using a matrix for their test cases? This does not seem ideal and is there anyway around this? I would assume that we would want to apply a `timeout_duration` to each individual test case.

vasireddy99:
> > > Does this timeout_duration apply to each separate job since it's a matrix configuration or the totality of all the jobs created under it? I could foresee that being a problem.
> > > Still though, filtering successful runs of the canary tests (seen [here](https://github.com/aws-observability/aws-otel-collector/actions/workflows/canary.yml?query=is%3Asuccess++)) are all sub 1hr (except 1) completion times. What has caused the time to increase so much? It is interesting through, if you go back to 9 days ago on that query you can see a batch of tests that took greater than 1hr to complete.
> > 
> > 
> > yeah, the tests took greater than 1 hr to complete previously as you mentioned and that's one of the reason we modified to 4 hours, if am not wrong. So from my understanding, there might be few cases such as,
> > ```
> > * Canary tests are passed only if after successful validation for 20 times. if one of the test case fail they. will go to the retry attempts.
> > 
> > * Possibility of available GitHub runners (not exactly sure how that works), similar to what we experienced in last week.
> > 
> > * Possibility of taking extra time when creating the aws resources.
> > ```
> 
> Now that you mentioned it I think it could be a safe assumption that the batch of > 1hr tests were due to a lack of github runners because we were running CI continuously. So that makes a lot of sense.
> 
> It sounds like you are saying that the `timeout_duration` applies to the entire workflow even though it is only added to one job. Is this due to the workflow using a matrix for their test cases? This does not seem ideal and is there anyway around this? I would assume that we would want to apply a `timeout_duration` to each individual test case.
> 
> If this is not possible, to apply the `timeout_duration`, to each test case then should we consider removing the `timeout_duration` limit?
> 
> This also brings up another question, if one test case fails will it trigger all of the test cases in the matrix again?

So, this particular `timeout_minutes` form [here](https://github.com/marketplace/actions/retry-step#timeout_minutes). It is a required parameter when using the retry and we cannot remove that.

If a test case fails, it will not trigger the whole workflow, it will be just the failed test case

bryan-aguilar:
> If a test case fails, it will not trigger the whole workflow, it will be just the failed test case

Can you provide a reference or test that verified this? Because right now it appears `timeout_duration` applies to the cumulative time it takes to run all test cases not each individual job. So by that logic it seems plausible that the retry could apply to the sum of all test cases instead of each individual job.

vasireddy99:
> > If a test case fails, it will not trigger the whole workflow, it will be just the failed test case
> 
> Can you provide a reference or test that verified this? Because right now it appears `timeout_duration` applies to the cumulative time it takes to run all test cases not each individual job. So by that logic it seems plausible that the retry could apply to the sum of all test cases instead of each individual job.

sure, I have followed this [reference link](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes). which applies to our use case.

bryan-aguilar:
> sure, I have followed this [reference link](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes). which applies to our use case. rather than the job level, it applies at the `step` level.

That is not the same as the previous link you shared which is [this ](https://github.com/marketplace/actions/retry-step#timeout_minutes). It is not clear whether `timeout-minutes` and `timeout_minutes` are the same. I think if no better documentation exists we need to build a quick test case in a fork with a matrix type setup similar the canary tests.

vasireddy99:
I agree, it is bit unclear to believe between `timeout-minutes` and `timeout_minutes`, based on the [error](https://github.com/aws-observability/aws-otel-collector/runs/5516364612?check_suite_focus=true#step:8:2144) we have. 
`Error: Timeout of 3600000ms hit` which I think is caused for `timeout_minutes = 60` that is applicable at step level. `60 *6000 = 360000`

I did quick test earlier to understand the functionality of it so I have used `timeout_seconds: 1` as `timeout_seconds` and `timeout_minutes` are relative
image

which led to this expected error

image

this gave me some understanding and although few other questions have popped which might not be related to this :-), can discuss offline a bit if that works better and move forward.

bryan-aguilar:
closed due to #1043

Issue #1049, opened 2022-03-14
Canary test timeout after successful validation. 


bryan-aguilar:
**Describe the bug**
Some canary tests are now timing out after successful validation. The second attempt is always successful because terraform does not need to apply any new changes. A few canary tests jobs are consistently timing out such as [`[otlp_trace,canary_linux]`](https://github.com/aws-observability/aws-otel-collector/runs/5536970369?check_suite_focus=true#step:8:1799). This could be due to an issue on how the terraform module is exiting. Similar retry logic is implemented in the CI tests but they do not exhibit this timeout behavior. This bug is causing tests to take an addition ~25minutes to complete. It has moved our average successful completion time for the canary test from approximately ~45 minutes to ~75minutes. 

**Steps to reproduce**
Run canary tests

**What did you expect to see?**
no timeout on successful validation

**What did you see instead?**
timeout

**Environment**
github action

**Additional context**
n/a

bryan-aguilar:
The same thing happens in the [`destroy resources`](https://github.com/aws-observability/aws-otel-collector/runs/5571198991?check_suite_focus=true) steps of CI. This is adding significant time to the CI process.

PR #1063, opened 2022-03-17
Reduce timeout duration in CI

bryan-aguilar:
**Description:** The PR reduces the timeout duration for e2e test in the CI workflow. These tests have no reason to be taking longer than an hour. Currently certain tests can hang such as `red_hat` test cases. This would cause the CI to hang for 2 hours while the test timed out. This brings it down to an hour which still may be too long but further analysis should be done.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -659,7 +659,7 @@ jobs:
         uses: nick-invision/retry@v2
         with:
           max_attempts: 3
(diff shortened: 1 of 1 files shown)


vasireddy99 (approved):
LGTM. 
the discussion/observation about the time tests are taking to run at the moment might also be related/tracked on below issue I guess.
https://github.com/aws-observability/aws-otel-collector/issues/1049

Issue #1008, opened 2022-03-03
Publish latest image sanity check


bryan-aguilar:
A sanity check should be done on the the step [`Pull image from integration test ECR, tag as latest and push to public release ECR and DockerHub`](https://github.com/aws-observability/aws-otel-collector/blob/4c46b62f95c55464bf619dcc48e00ec2428e28ef/.github/workflows/CD.yml#L612) in the CD workflow. It is not entirely clear what would happen if we patch a previous minor version of the ADOT Collector. A high level overview of the [version checking logic](https://github.com/aws-observability/aws-otel-collector/blob/main/tools/workflow/docker-version-compare.sh) and the step conditions make it appear that the latest tag is always pushed. The latest tag should not be pushed when a patch version is not associated with the latest minor version.

bryan-aguilar:
What about the `release-latest-s3` job?

bryan-aguilar:
We also release to SSM. I don't think we have a latest for that but did we verify?

bryan-aguilar:
I guess that was a bit outside of the scope of this issue since this only brings up the `image`. If you think it would be better to open a separate issue for these items then please do so. If not maybe we can change the description/title of this issue.

Kausik-A:
As this issue was addressing one specific step of the CD workflow, I think its better we continue this discussion on another issue, while also keeping a reference to #1008. I will start a new one !

PR #1096, opened 2022-03-29
Better code redability in docker-version-compare.sh

Kausik-A:
**Description:** 

Changed the part of setting up the appropriate  `ANY_UPDATE ` boolean value in accordance to the target update for better code readability. 
The unlikely but possible problem with the old approach is if the ANY_UPDATE=true command were to fail, then ANY_UPDATE=false would be executed 

**Link to tracking Issue:** 
https://github.com/aws-observability/aws-otel-collector/issues/1008

**Testing:** 
Ran the bash scripts locally to verify the output.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -628,7 +628,7 @@ jobs:
 
       - name: Login to Public Release ECR
         id: login-ecr
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && steps.version.outputs.LATEST_OR_NEWER == 'true' }}
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && steps.version.outputs.latest-or-newer == 'true' }}
         uses: docker/login-action@v1
         with:
           registry: public.ecr.aws
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -628,7 +628,7 @@ jobs:
 
       - name: Login to Public Release ECR
         id: login-ecr
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && steps.version.outputs.any-update == 'true' }}
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && steps.version.outputs.LATEST_OR_NEWER == 'true' }}
         uses: docker/login-action@v1
         with:
           registry: public.ecr.aws
@@ -638,7 +638,7 @@ jobs:
           AWS_REGION: us-east-1
 
       - name: Pull image from integration test ECR, tag as latest and push to public release ECR and DockerHub
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.any-update == 'true' }}
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.LATEST_OR_NEWER == 'true' }}
         uses: akhilerm/tag-push-action@v2.0.0
         with:
           src: public.ecr.aws/${{ env.ECR_REPO }}:${{ needs.release-checking.outputs.version }}
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -638,7 +638,7 @@ jobs:
           AWS_REGION: us-east-1
 
       - name: Pull image from integration test ECR, tag as latest and push to public release ECR and DockerHub
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.LATEST_OR_NEWER == 'true' }}
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.latest-or-newer== 'true' }}
         uses: akhilerm/tag-push-action@v2.0.0
         with:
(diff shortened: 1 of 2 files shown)


Aneurysm9:
Would `LATEST_OR_NEWER` be more descriptive?  It seems like the logic is determining whether the target version is either the current latest version or a newer version than that.  It would be false if the target version is less than the current latest version.

Kausik-A:
This sounds right. NOTE: As we change the variable name, we also have to change its reference in CD.yaml. So is there a way to test the CD.yaml workflow after the changes are being made ?

bryan-aguilar:
> This sounds right. NOTE: As we change the variable name, we also have to change its reference in CD.yaml. So is there a way to test the CD.yaml workflow after the changes are being made ?

The short answer is, not easily. You can safely assume that any reference to `ANY_UPDATE` needs to be changed to reflect the new variable name.

tools/workflow/docker-version-compare.sh
```diff
@@ -91,10 +91,17 @@ elif [ "${TARGET_MAJOR}" -eq "${LATEST_MAJOR}" ]; then
     fi
 fi
 
-[ ${MAJOR_UPDATE} == "true" ] || [ ${MINOR_UPDATE} == "true" ] || [ ${PATCH_UPDATE} == "true" ] || [ ${SAME_VERSION} == "true" ] && ANY_UPDATE=true || ANY_UPDATE=false
+# If any of the updates are true, then LATEST_OR_NEWER = true
+# LATEST_OR_NEWER would be false if the target version is less than the current latest version or vice cersa
```
bryan-aguilar, line 95:
```suggestion
# LATEST_OR_NEWER would be false if the target version is less than the current latest version
```

tools/workflow/docker-version-compare.sh
```diff
@@ -91,10 +91,17 @@ elif [ "${TARGET_MAJOR}" -eq "${LATEST_MAJOR}" ]; then
...
+    LATEST_OR_NEWER=true
+else
+    LATEST_OR_NEWER=false
+fi
 
 echo "::set-output name=major-update::${MAJOR_UPDATE}"
 echo "::set-output name=minor-update::${MINOR_UPDATE}"
 echo "::set-output name=patch-update::${PATCH_UPDATE}"
-echo "::set-output name=any-update::${ANY_UPDATE}"
+echo "::set-output name=any-update::${LATEST_OR_NEWER}"
```
bryan-aguilar, line 105:
```suggestion
echo "::set-output name=latest-or-newer::${LATEST_OR_NEWER}"
```

.github/workflows/CD.yml
```diff
@@ -628,7 +628,7 @@ jobs:
 
       - name: Login to Public Release ECR
         id: login-ecr
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && steps.version.outputs.any-update == 'true' }}
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && steps.version.outputs.LATEST_OR_NEWER == 'true' }}
```
bryan-aguilar, line 631:
```suggestion
        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true' && steps.version.outputs.latest-or-newer == 'true' }}
```

.github/workflows/CD.yml
```diff
@@ -638,7 +638,7 @@ jobs:
           AWS_REGION: us-east-1
 
       - name: Pull image from integration test ECR, tag as latest and push to public release ECR and DockerHub
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.any-update == 'true' }}
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.LATEST_OR_NEWER == 'true' }}
```
bryan-aguilar, line 641:
```suggestion
        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.latest-or-newer== 'true' }}
```

tools/workflow/docker-version-compare.sh
```diff
@@ -26,7 +26,7 @@ set -e
 #
 # Step outputs
 # (All booleans: true/false)
-# 1. any-update
+# 1. LATEST_OR_NEWER
```
bryan-aguilar, line 29:
```suggestion
# 1. latest-or-newer
```

tools/workflow/docker-version-compare.sh
Kausik-A: Good catch, I still need to extra cautious with bash

.github/workflows/CD.yml
```diff
@@ -638,7 +638,7 @@ jobs:
           AWS_REGION: us-east-1
 
       - name: Pull image from integration test ECR, tag as latest and push to public release ECR and DockerHub
-        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.any-update == 'true' }}
+        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.latest-or-newer== 'true' }}
```
bryan-aguilar, line 641:
```suggestion
        if: ${{ steps.release-latest-image.outputs.cache-hit != 'true'  && steps.version.outputs.latest-or-newer == 'true' }}
```

bryan-aguilar (approved):

PR #1103, opened 2022-03-31
Use latest version tag in go install for workflow

bryan-aguilar:
**Description:** Go install requires a version tag when not installing from with a module. 

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -82,7 +82,7 @@ jobs:
         with:
           go-version: '^1.17.7'
       - name: Install Go tools
-        run: cd /tmp && go install golang.org/x/tools/cmd/goimports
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -82,7 +82,7 @@ jobs:
         with:
           go-version: '^1.17.7'
       - name: Install Go tools
-        run: cd /tmp && go install golang.org/x/tools/cmd/goimports
+        run: cd /tmp && go install golang.org/x/tools/cmd/goimports@latest
```
Aneurysm9, line 85: Let's fix this at[ the source](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/cmd/aotutil/Makefile#L4).  Not sure why that is set up to format before building.  Things should be formatted before being committed and building shouldn't need to format again.

PR #1277, opened 2022-06-01
Update OTel-http port to match the specification

vasireddy99:
**Description:** 

This PR updates the OTel http port to `4318`, it includes the changes that might have missed in previous PR's

Reference:

 * https://github.com/open-telemetry/opentelemetry-collector/pull/3743

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -7,7 +7,7 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:4317
       http:
-        endpoint: 0.0.0.0:55681
+        endpoint: 0.0.0.0:4318
   awsxray:
     endpoint: 0.0.0.0:2000
     transport: udp
--- a/config/ecs/container-insights/otel-task-metrics-config.yaml
+++ b/config/ecs/container-insights/otel-task-metrics-config.yaml
@@ -7,7 +7,7 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:4317
       http:
-        endpoint: 0.0.0.0:55681
+        endpoint: 0.0.0.0:4318
   awsxray:
     endpoint: 0.0.0.0:2000
     transport: udp
--- a/config/ecs/ecs-default-config.yaml
+++ b/config/ecs/ecs-default-config.yaml
@@ -7,7 +7,7 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:4317
       http:
-        endpoint: 0.0.0.0:55681
+        endpoint: 0.0.0.0:4318
   awsxray:
     endpoint: 0.0.0.0:2000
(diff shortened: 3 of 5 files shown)


config/ecs/ecs-default-config.yaml
```diff
@@ -7,7 +7,7 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:4317
       http:
-        endpoint: 0.0.0.0:55681
+        endpoint: 0.0.0.0:4318
```
bryan-aguilar, line 10:
```suggestion
```

config/ecs/ecs-default-config.yaml
```diff
@@ -7,7 +7,7 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:4317
```
bryan-aguilar, line 8:
```suggestion
```

config/ecs/container-insights/otel-task-metrics-config.yaml
```diff
@@ -7,7 +7,7 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:4317
```
bryan-aguilar, line 8:
```suggestion
```

config.yaml
```diff
@@ -7,7 +7,7 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:4317
```
bryan-aguilar, line 8:
```suggestion
```

bryan-aguilar (commented):
We should be consistent in whether or not we provide default points. I suggest we do not provide the `endpoint` field at all for any `otlp` receivers.

vasireddy99:
> We should be consistent in whether or not we provide default points. I suggest we do not provide the `endpoint` field at all for any `otlp` receivers.

We are currently providing default end points in all of the configs such as in [here](https://github.com/aws-observability/aws-otel-collector/blob/main/config/ecs/ecs-amp-xray-prometheus.yaml#L10), since default will likely to not change much often from now, I think it will be okay?.

bryan-aguilar:
> > We should be consistent in whether or not we provide default points. I suggest we do not provide the `endpoint` field at all for any `otlp` receivers.
> 
> We are currently providing default end points in all of the configs such as in [here](https://github.com/aws-observability/aws-otel-collector/blob/main/config/ecs/ecs-amp-xray-prometheus.yaml#L10), since default will likely to not change much often from now, I think it will be okay?.

Ahh I see, so the `apprunner` config is the only one that does not specify defaults. So this is merely a preference of styling. I could go either way on this and don't know if there is a good argument for keeping or leaving them.

bryan-aguilar (approved):

PR #1278, opened 2022-06-02
Add create-test-ref need in CI-Batched job `release/v0.17.x` branch

bryan-aguilar:
**Description:** Adds missing need to ensure the correct ref is checked out in `validate-test-passed` job in `CI-Batched

### DIFF
--- a/.github/workflows/CI-Batched.yml
+++ b/.github/workflows/CI-Batched.yml
@@ -664,7 +664,7 @@ jobs:
 
   validate-all-tests-pass:
     runs-on: ubuntu-latest
-    needs: [run-batch-job,e2etest-preparation]
+    needs: [run-batch-job,e2etest-preparation,create-test-ref]
     outputs:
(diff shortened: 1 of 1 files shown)


vasireddy99 (approved):

.github/workflows/CI-Batched.yml
```diff
@@ -664,7 +664,7 @@ jobs:
 
   validate-all-tests-pass:
     runs-on: ubuntu-latest
-    needs: [run-batch-job,e2etest-preparation]
+    needs: [run-batch-job,e2etest-preparation,create-test-ref]
```
vasireddy99, line 667: Looks like we already have this [need](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CI-Batched.yml#L582) for `run-batch-job` and since run-batch-job is already a need for this step. is this a required change.

PR #1304, opened 2022-06-17
Separate Collector and Operator CI

erichsueh3:
This PR removes the Operator tests from the Collector CI. Additionally, a new workflow, `CI-Operator`, is created, that includes only the Operator CI tests. It runs one batch tests, the `EKS_ADOT_OPERATOR` tests, and is triggered manually. 

### DIFF
--- a/.github/workflows/CI-Operator.yml
+++ b/.github/workflows/CI-Operator.yml
@@ -1,0 +1,150 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+name: C/I-Operator
+on: workflow_dispatch
+
+env:
+  EKS_ARM_64_AMP_ENDPOINT: "https://aps-workspaces.us-east-2.amazonaws.com/workspaces/ws-1de68e95-0680-42bb-8e55-67e7fd5d0861"
+  EKS_ARM_64_CLUSTER_NAME: "integ-test-arm64-cluster"
+  EKS_ARM_64_REGION: "us-east-2"
+  TESTING_FRAMEWORK_REPO: aws-observability/aws-otel-test-framework
+  # DDB_TABLE_NAME: BatchTestCache
+  NUM_BATCHES: 1
+
+
+concurrency:
+  group: ci-batched${{ github.ref_name }}
+  cancel-in-progress: true  
+
+jobs:
+  create-test-ref:
+    runs-on: ubuntu-latest
+    outputs:
+      testRef: ${{ steps.setRef.outputs.ref }}
+    steps:
+      - name: Set testRef output
+        id: setRef
+        run: |
+          if [[ ${{ github.ref_name }} == release/v* ]]; then 
+            echo "::set-output name=ref::${{github.ref_name}}"
+          else
+            echo "::set-output name=ref::terraform"
+          fi
+
+  get-testing-suites: 
+    runs-on: ubuntu-latest
+    needs: [create-test-ref]
+    outputs:
+      test-case-batch-key: ${{ steps.set-batches.outputs.batch-keys }}
+      test-case-batch-value: ${{ steps.set-batches.outputs.batch-values }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ env.TESTING_FRAMEWORK_REPO }}
+          path: testing-framework
+          ref: ${{ needs.create-test-ref.outputs.testRef }}
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: '~1.17.7'
+
+      - name: Create test batch key values
+        id: set-batches
+        run: |
+          cd testing-framework/tools/batchTestGenerator
+          go build
+          ./batchTestGenerator github --testCaseFilePath=./testcases.json --maxBatch=${{ env.NUM_BATCHES }} \
+            --eksarm64amp=${{ env.EKS_ARM_64_AMP_ENDPOINT }} \
+            --eksarm64region=${{ env.EKS_ARM_64_REGION }} \
+            --eksarm64cluster=${{ env.EKS_ARM_64_CLUSTER_NAME }} \
+            --include=EKS_ADOT_OPERATOR
+      - name: List testing suites
+        run: |
+          echo ${{ steps.set-batches.outputs.batch-keys }}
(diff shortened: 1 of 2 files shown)


.github/workflows/CI-Operator.yml
```diff
@@ -0,0 +1,150 @@
...
+
+name: C/I-Operator
+on: workflow_dispatch
+
+env:
+  EKS_ARM_64_AMP_ENDPOINT: "https://aps-workspaces.us-east-2.amazonaws.com/workspaces/ws-1de68e95-0680-42bb-8e55-67e7fd5d0861"
+  EKS_ARM_64_CLUSTER_NAME: "integ-test-arm64-cluster"
+  EKS_ARM_64_REGION: "us-east-2"
+  TESTING_FRAMEWORK_REPO: aws-observability/aws-otel-test-framework
+  # DDB_TABLE_NAME: BatchTestCache
```
Aneurysm9, line 22: Is this not needed here?

.github/workflows/CI-Operator.yml
```diff
@@ -0,0 +1,150 @@
...
+  NUM_BATCHES: 1
+
+
+concurrency:
+  group: ci-batched${{ github.ref_name }}
```
Aneurysm9, line 27:
```suggestion
  group: ci-operator${{ github.ref_name }}
```

Does this need to have the same concurrency group as the other workflow?

.github/workflows/CI-Operator.yml
erichsueh3: Meant to take this out, as this is used in the `validate-all-tests-pass` job which is used to set `release-candidate-ready`, but if `run-batch-jobs` is green that should indicate all tests passing as well

.github/workflows/CI-Operator.yml
erichsueh3: no, will change

.github/workflows/CI-Operator.yml
```diff
@@ -0,0 +1,149 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+name: C/I-Operator
+on: workflow_dispatch
+
+env:
+  EKS_ARM_64_AMP_ENDPOINT: "https://aps-workspaces.us-east-2.amazonaws.com/workspaces/ws-1de68e95-0680-42bb-8e55-67e7fd5d0861"
+  EKS_ARM_64_CLUSTER_NAME: "integ-test-arm64-cluster"
+  EKS_ARM_64_REGION: "us-east-2"
```
bryan-aguilar, line 20:
```suggestion
```

.github/workflows/CI-Operator.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Create test batch key values
+        id: set-batches
+        run: |
+          cd testing-framework/tools/batchTestGenerator
+          go build
+          ./batchTestGenerator github --testCaseFilePath=./testcases.json --maxBatch=${{ env.NUM_BATCHES }} \
+            --eksarm64amp=${{ env.EKS_ARM_64_AMP_ENDPOINT }} \
+            --eksarm64region=${{ env.EKS_ARM_64_REGION }} \
+            --eksarm64cluster=${{ env.EKS_ARM_64_CLUSTER_NAME }} \
```
bryan-aguilar, line 71:
```suggestion
```

.github/workflows/CI-Operator.yml
```diff
@@ -0,0 +1,149 @@
...
+        run: echo "::set-output name=ttldate::$(date -u -d "+7 days" +%s)"
+
+      # version used is the same as determined during 'e2etest-preparation' to ensure same 'adot-collector-integration-test' image is used
+      - name: Versioning for testing
+        id: versioning
+        run: |
+          version="$(cat VERSION)-$(git rev-parse --short HEAD)"
+          echo $version > $PACKAGING_ROOT/VERSION
+          cat $PACKAGING_ROOT/VERSION
+          echo "::set-output name=version::$version"
```
bryan-aguilar, line 128: This creates a direct dependency on the Collector CI being ran so that this image is created. Is this a coupling we want to accept? 

Also, don't we want to guarantee that we are testing upstream collector version against operator version. This versioning strategy does not guarantee that.

bryan-aguilar (changes requested):
`DDB_TABLE_NAME` does need to be included for this as the [`execute`](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/executeTerraformTest.sh) script utilizes that env variable to query the cache. See the linked file for required env vars.

.github/workflows/CI-Operator.yml
erichsueh3: Wouldn't we want to test against the downstream collector version that is being released, i.e. this image that's being created by Collector CI? Are we releasing the Operator on a different cadence than the Collector?

bryan-aguilar: I believe this could be okay, as the CI for releases would be done on a release branch and I do not think collector version drift would be an issue there. I do wonder if there are edge cases we haven't quite considered.

bryan-aguilar:
I also am wondering if choosing an `on-demand` CI for the operator with `workflow_dispatch` is the best idea. This means that we will only see failures during release time and lose visibility of continuously testing. Have we considered the effect of switching to this model of testing?

erichsueh3:
> I also am wondering if choosing an `on-demand` CI for the operator with `workflow_dispatch` is the best idea. This means that we will only see failures during release time and lose visibility of continuously testing. Have we considered the effect of switching to this model of testing?

I've made it so this workflow runs on push so we don't lose visibility there. The option to run the CI on demand still exists, and additionally includes the ability to specify a Collector version to test against, so that Operator CI can be ran without being dependent on the Collector CI in the case of an ad hoc release for the Operator.

.github/workflows/CI-Operator.yml
```diff
@@ -0,0 +1,166 @@
...
+      - name: Get TTL_DATE for cache
+        id: date
+        run: echo "::set-output name=ttldate::$(date -u -d "+7 days" +%s)"
+
+      # version used is the same as determined during 'e2etest-preparation' to ensure same 'adot-collector-integration-test' image is used
+      # if version is specified during manual run of workflow, then that version is used.
+      - name: Versioning for testing
+        id: versioning
+        run: |
+          if [[ ${{ github.event.inputs.version }} == '' ]]; then
```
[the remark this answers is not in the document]
erichsueh3: Updated - will check for on `push` or for no version provided (meaning `dispatch`, but still needs a version)

.github/workflows/CI-Operator.yml
```diff
@@ -0,0 +1,166 @@
...
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+
+      - name: Set up terraform
+        uses: hashicorp/setup-terraform@v2
+
+      - name: Configure AWS Credentials
+        if: steps.e2etest-eks.outputs.cache-hit != 'true'
```
bryan-aguilar, line 108:
```suggestion
```

bryan-aguilar (approved):

.github/workflows/CI-Operator.yml
erichsueh3: good catch, fixed

Aneurysm9 (approved):

PR #1305, opened 2022-06-17
Add processors consistent with opentelemetry-lambda upstream

kolanos:
**Description:** <Describe what has changed.>

Adds processors consistent with the [opentelemetry-lambda upstream](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/lambdacomponents/default.go#L54).

**Link to tracking Issue:** <Issue number if applicable>

aws-observability/aws-otel-lambda#104

**Testing:** <Describe what testing was performed and which tests were added.>

Added tests consistent with existing component tests.

**Documentation:** <Describe the documentation added.>

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1312, opened 2022-06-23
Update upstream to v0.54.0

### LINKED ISSUE #1311: Upgrade collector version to v0.54.0

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

**Link to tracking Issue:** Fixes #1311


bryan-aguilar:
Missing updates to pkg/lambdacomponents. Are we skipping that for this PR?

Aneurysm9:
> Missing updates to pkg/lambdacomponents. Are we skipping that for this PR?

I thought we were updating those separately, but I can do it here if it should be done now.  I think the upstream Lambda layer needs updates, too, though.

bryan-aguilar:
> > Missing updates to pkg/lambdacomponents. Are we skipping that for this PR?
> 
> I thought we were updating those separately, but I can do it here if it should be done now. I think the upstream Lambda layer needs updates, too, though.

Historically we have been updating those dependencies at the same time. That go.mod file is just  the list of default components for the ADOT lambda layer to include so it seems reasonable to update those dependencies at the same time.

kolanos:
@bryan-aguilar Can #1305 be included in this update by any chance?

PR #1344, opened 2022-07-08
Set operator image from go.mod collector dependency

erichsueh3:
This PR follows [PR 750](https://github.com/aws-observability/aws-otel-test-framework/pull/750) in the test-framework repo that allows the Operator image to be set as a variable. This PR sets the variable tag to be the same as the collector dependency taken from the `go.mod` file. 

### DIFF
--- a/.github/workflows/CI-Operator.yml
+++ b/.github/workflows/CI-Operator.yml
@@ -162,6 +162,7 @@ jobs:
         run: |
           export TTL_DATE=${{ steps.date.outputs.ttldate }}
           export TF_VAR_aoc_version=${{ steps.versioning.outputs.version }}
+          export TF_VAR_operator_tag=$(cat go.mod | grep "go.opentelemetry.io/collector " | cut -d " " -f2)
           cd testing-framework/terraform
           make execute-batch-test
           


bryan-aguilar:
Should the workflow fail if that tag doesn't exist upstream? This would cause the tests to continually run and fail during operator deployment because it cannot deploy the image during the test.

Aneurysm9:
> Should the workflow fail if that tag doesn't exist upstream? This would cause the tests to continually run and fail during operator deployment because it cannot deploy the image during the test.

Probably?  I think this means that updating the collector dependency before the corresponding operator release upstream would cause operator CI failures.  That would be expected as we couldn't test an operator that doesn't exist.  We can make it a policy to not update the collector dependency until the corresponding operator is released if we find it to be problematic.

One thing that might not be good about that approach is what happens when the collector makes a patch release we need to incorporate but the operator doesn't?

erichsueh3:
> Probably? I think this means that updating the collector dependency before the corresponding operator release upstream would cause operator CI failures. That would be expected as we couldn't test an operator that doesn't exist. We can make it a policy to not update the collector dependency until the corresponding operator is released if we find it to be problematic.


This makes sense, and I don't think it would be too problematic either since it would only affect Operator releases and not Collector releases now that those processes are decoupled; we can't release an ADOT Operator without an upstream Operator release either

> One thing that might not be good about that approach is what happens when the collector makes a patch release we need to incorporate but the operator doesn't?

Currently thinking of ways to get around this, I think the easiest way is to check for a patch version before setting the image, and if it is then set the image to `vx.x.0` instead, will test it out and update here

erichsueh3:
Another edge case to consider is when a patch release of the upstream Operator has been done, and we need to test using that Operator image; just another shortcoming of this approach, but working on mitigating both issues currently

tools/release/adot-operator-images-mirror/get-operator-tag.sh
```diff
@@ -0,0 +1,58 @@
...
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+COLLECTOR_TAG=$(cat go.mod | grep "go.opentelemetry.io/collector " | cut -d " " -f2)
+COLLECTOR_PATCH_VER=${COLLECTOR_TAG: -1}
+COLLECTOR_TAG_WITHOUT_PATCH_VER=$(echo $COLLECTOR_TAG | sed 's/.$//')
```
Aneurysm9, line 18:
```suggestion
COLLECTOR_TAG_WITHOUT_PATCH_VER=$(echo $COLLECTOR_TAG | sed 's/\..*$//')
```
Is this meant to remove the last `.` and everything following?  If so, it's probably not doing what you expect.  Instead, it's removing the last character.

tools/release/adot-operator-images-mirror/get-operator-tag.sh
```diff
@@ -0,0 +1,58 @@
...
+
+OPERATOR_TAGS=$(curl https://api.github.com/repos/open-telemetry/opentelemetry-operator/tags)
+
+if [[ $COLLECTOR_PATCH_VER == "0" ]]; then
+    # Check if any Operator image exists for the Collector image
+    if echo $OPERATOR_TAGS | grep -q $COLLECTOR_TAG_WITHOUT_PATCH_VER; then
```
Aneurysm9, line 24: I would recommend using `jq` instead of `grep` to work with JSON documents.

tools/release/adot-operator-images-mirror/get-operator-tag.sh
```diff
@@ -0,0 +1,58 @@
...
+        # NO COLLECTOR PATCH, NO OPERATOR IMAGE - FAIL WORKFLOW
+        echo "NO OPERATOR IMAGE EXISTS FOR THIS COLLECTOR IMAGE"
+        exit 1
+    fi
+else
+    # Check if any Operator image exists for the Collector image
+    if echo $OPERATOR_TAGS | grep -q $COLLECTOR_TAG_WITHOUT_PATCH_VER; then
+        # Checks if there is more than one operator tag that has the same major/minor version; if only one, then there are no operator patches
+        if [[ $(echo "$OPERATOR_TAGS" | grep "\"name\": \"$COLLECTOR_TAG_WITHOUT_PATCH_VER" -c) == "1" ]]; then 
+            # COLLECTOR PATCH, NO PATCH FOR OPERATOR IMAGE - IMAGE TAG USED == vx.x.0
```
Aneurysm9, line 45: This should find the most recent operator patch version for the corresponding major.minor.  Don't use `x.y.0` if `x.y.1` exists, even if the collector is on `x.y.2`.

tools/release/adot-operator-images-mirror/get-operator-tag.sh
erichsueh3: This logic really only needs to remove the last character as `COLLECTOR_TAG_WITHOUT_PATCH_VER` is mostly used in `grep` checks where the extra `.` won't change things - fixed regardless though, for better clarity

tools/release/adot-operator-images-mirror/get-operator-tag.sh
erichsueh3: Changed

tools/release/adot-operator-images-mirror/get-operator-tag.sh
erichsueh3: Will work on fixing this logic - I was too focused on patch vs no patch, when in reality it could be something like you said, with `x.y.2` vs `x.y.1` - it should just default to using the latest available Operator image for `x.y`

erichsueh3:
After considering @Aneurysm9's comment I realized the logic for the script could be massively simplified; we will always want the latest operator patch version for a given major.minor version, and so the script has been simplified to, from a list of operator tags, get a subset of tags that correspond to the major.minor version of a collector, and return the first tag (which will always be the latest patch version). If there is no major.minor version of the operator for the given collector major.minor version, the workflow will fail

tools/release/adot-operator-images-mirror/get-operator-tag.sh
```diff
@@ -0,0 +1,27 @@
...
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+COLLECTOR_TAG=$(cat go.mod | grep "go.opentelemetry.io/collector " | cut -d " " -f2)
+COLLECTOR_TAG_WITHOUT_PATCH_VER=$(echo $COLLECTOR_TAG | sed 's/\..$//')
+
+OPERATOR_TAGS=$(echo $(curl https://api.github.com/repos/open-telemetry/opentelemetry-operator/tags) | jq  '.[] | .name')
+OPERATOR_TAG=$(echo "$OPERATOR_TAGS" | grep "$COLLECTOR_TAG_WITHOUT_PATCH_VER" -m 1 | sed 's/"//g')
```
Aneurysm9, line 20:
```suggestion
OPERATOR_TAG=$(curl https://api.github.com/repos/open-telemetry/opentelemetry-operator/tags | jq  "map(select( .name | startswith(\"${COLLECTOR_TAG_WITHOUT_PATCH_VER}\"))) | first | .name")
```

Let `jq` do the heavy lifting for you.  I'm not 100% certain the escaping on this is correct, but the `jq` pipeline works.  It `map`s each object in the array, `select`ing those whose `.name` `startswith` the desired value, then takes the `first` item in the resulting array and extracts its `.name`.

tools/release/adot-operator-images-mirror/get-operator-tag.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+COLLECTOR_TAG=$(cat go.mod | grep "go.opentelemetry.io/collector " | cut -d " " -f2)
+COLLECTOR_TAG_WITHOUT_PATCH_VER=$(echo $COLLECTOR_TAG | sed 's/\..$//')
```
Aneurysm9, line 17:
```suggestion
COLLECTOR_TAG=$(cat go.mod | grep "go.opentelemetry.io/collector " | cut -d " " -f2 | sed 's/\..$//)
```

I don't think the full tag is used anywhere anymore, so might as well collapse this.  

Also the regex needs to catch all characters after the final `.`.  Right now it will break if we ever get a `vX.Y.10`.  If it were PCRE then `s/\.[^.]+$//` would work, but not sure `sed` supports that negative range format.  If you can make it work quickly, great, otherwise leave a comment and maybe create an issue to tack the tech debt.  I doubt we're going to get ten patch releases on a single version of the collector any time soon.

erichsueh3: Collapsed, and changed the `sed` to another `cut` (`cut -d "." -f 1,2`) so that it instead delimits by `.` and gets the first two fields (major.minor) but not the third (patch)

Aneurysm9: now you're thinking with pipelines! :)

tools/release/adot-operator-images-mirror/get-operator-tag.sh
```diff
@@ -0,0 +1,25 @@
...
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+COLLECTOR_TAG=$(cat go.mod | grep "go.opentelemetry.io/collector " | cut -d " " -f 2 | cut -d "." -f 1,2)
+
+OPERATOR_TAG=$(curl https://api.github.com/repos/open-telemetry/opentelemetry-operator/tags | jq  "map(select( .name | startswith(\"${COLLECTOR_TAG}\"))) | first | .name" | sed 's/"//g')
```
Aneurysm9, line 18:
```suggestion
OPERATOR_TAG=$(curl https://api.github.com/repos/open-telemetry/opentelemetry-operator/tags | jq -r "map(select( .name | startswith(\"${COLLECTOR_TAG}\"))) | first | .name"')
```

`jq -r` to get the raw, unquoted value and then there's no need to strip quotes with `sed`.

Aneurysm9 (approved):

tools/release/adot-operator-images-mirror/get-operator-tag.sh
```diff
@@ -0,0 +1,25 @@
...
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+COLLECTOR_TAG=$(cat go.mod | grep "go.opentelemetry.io/collector " | cut -d " " -f 2 | cut -d "." -f 1,2)
+
+OPERATOR_TAG=$(curl https://api.github.com/repos/open-telemetry/opentelemetry-operator/tags | jq -r "map(select( .name | startswith(\"${COLLECTOR_TAG}\"))) | first | .name")
```
bryan-aguilar, line 18: Is the order of the tags always guaranteed to be the same?

erichsueh3: The tags are always returned as a list ordered from latest --> oldest, but technically there is no guarantee on that

Aneurysm9: can add `sort_by(.name)` to ensure it, but it appeared that they did come in order.

tools/release/adot-operator-images-mirror/get-operator-tag.sh
erichsueh3: Fixed

PR #1365, opened 2022-07-18
Add operator arm64 tests

erichsueh3:
This PR adds testcases to be tested by the Operator with arm64. This also adds these tests to the Operator CI

### DIFF
--- a/.github/workflows/CI-Operator.yml
+++ b/.github/workflows/CI-Operator.yml
@@ -36,9 +36,6 @@ env:
   TESTING_FRAMEWORK_REPO: aws-observability/aws-otel-test-framework
   DDB_TABLE_NAME: BatchTestCache
   NUM_BATCHES: 1
-  EKS_ARM_64_AMP_ENDPOINT: "https://aps-workspaces.us-east-2.amazonaws.com/workspaces/ws-1de68e95-0680-42bb-8e55-67e7fd5d0861"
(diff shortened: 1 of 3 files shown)


bryan-aguilar:
My previous comment is not entirely true. The test case generator is still used for Canary tests. The Collector and Operator CI utilize the batch test generator though.

erichsueh3:
> My previous comment is not entirely true. The python test case generator is still used for Canary tests. The Collector and Operator CI utilize the batch test generator though.

Good call, will adjust this PR to be only the change to the Operator CI

bryan-aguilar:
This also doesn't add the necessary flags for cluster name, region, and AMP ws

erichsueh3:
> This also doesn't add the necessary flags for cluster name, region, and AMP ws

[This comment](https://github.com/aws-observability/aws-otel-test-framework/pull/767#discussion_r932757347) explains it briefly, but we do not need to set the endpoint/cluster/region arm64 flags for the operator arm64 tests. This is because the operator arm64 cluster lives in the default region uses the default amp endpoint. The cluster is then set in the `executeTerraformTest.sh` script in a `-var=` flag, which takes precedence over all other methods of defining a terraform variable (The way non-arm64 operator tests set a cluster name is by a `.tfvars` file which has higher precedence than the default variable value from variables.tf)

bryan-aguilar (approved):

PR #1497, opened 2022-09-16
Use IAM Role instead of User creds in GHA workflows

bryan-aguilar:
**Description:** Use an IAM Role that utilizes the GH OIDC provider to configure AWS credentials. If `role-duration-seconds` is not specified then it will default to one hour. 

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -123,7 +123,7 @@ jobs:
           role-to-assume: ${{ secrets.COLLECTOR_ASSUMABLE_ROLE_ARN }}
           aws-region: us-west-2
           # 3 hours
-          role-duration-seconds: 108000
+          role-duration-seconds: 10800
           
       - name: Set up JDK 11
         uses: actions/setup-java@v3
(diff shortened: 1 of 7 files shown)


vasireddy99 (approved):

.github/workflows/perf.yml
```diff
@@ -123,9 +120,10 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-access-key-id: ${{ secrets.INTEG_TEST_AWS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.INTEG_TEST_AWS_KEY_SECRET }}
+          role-to-assume: ${{ secrets.COLLECTOR_ASSUMABLE_ROLE_ARN }}
           aws-region: us-west-2
+          # 3 hours
+          role-duration-seconds: 108000
```
Aneurysm9, line 126:
```suggestion
          role-duration-seconds: 10800
```

I think that was 30h.

Aneurysm9 (approved):

PR #1528, opened 2022-09-29
Add APIGateway cleaner to resource cleaner

bryan-aguilar:
**Description:** Add API Gateway cleaner to clean up leftover api gateways from lambda tests. 

**Testing:** I wrote a unit test to verify the regex works. 

### DIFF
--- a/tools/workflow/cleaner/apigw/clean.go
+++ b/tools/workflow/cleaner/apigw/clean.go
@@ -40,21 +40,20 @@ func Clean(sess *session.Session, expirationDate time.Time) error {
 			break
 		}
 		nextToken = getRestApisOutput.Position
+	}
+	if len(deleteGatewayIds) < 1 {
+		logger.Printf("No API Gateways to delete")
+		return nil
+	}
 
-		if len(deleteGatewayIds) < 1 {
-			logger.Printf("No API Gateways to delete")
-			return nil
+	for _, id := range deleteGatewayIds {
+		terminateGatewayInput := &apigateway.DeleteRestApiInput{RestApiId: id}
+		if _, err := apigwclient.DeleteRestApi(terminateGatewayInput); err != nil {
+			return fmt.Errorf("unable to delete gateway: %w", err)
 		}
-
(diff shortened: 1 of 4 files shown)


tools/workflow/cleaner/apigw/clean.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		for _, id := range deleteGatewayIds {
+			terminateGatewayInput := &apigateway.DeleteRestApiInput{RestApiId: id}
+			if _, err := apigwclient.DeleteRestApi(terminateGatewayInput); err != nil {
+				return fmt.Errorf("unable to delete workspace: %w", err)
+			}
+			// 30 second request limit to delete ap
+			// https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html#api-gateway-control-service-limits-table
+			time.Sleep(31 * time.Second)
```
rapphil, line 56: I'm just curios if deleting while you still paginating over the existing API Gateways will not causes any problem in the pagination.

I think it is safer to paginate and collect the full list and then delete?

tools/workflow/cleaner/apigw/clean.go
```diff
@@ -0,0 +1,80 @@
...
+				logger.Printf("Try to delete gateway %s created-at %v", *gw.Id, gw.CreatedDate)
+				deleteGatewayIds = append(deleteGatewayIds, gw.Id)
+			}
+		}
+		if getRestApisOutput.Position == nil {
+			break
+		}
+		nextToken = getRestApisOutput.Position
+
+		if len(deleteGatewayIds) < 1 {
```
Aneurysm9, line 44: Is this right?  What if the first page of results produced nothing to delete but subsequent pages would?

It feels like the `for` block containing this should end just before this line and the function from here out executed once and only once when the full list of gateways to delete has been constructed.

bryan-aguilar:
@Aneurysm9 @rapphil Thank you, good catch. That was definitely an oversight.

Aneurysm9 (approved):

PR #1541, opened 2022-10-06
Additional IAM role and policy to the resource cleaner

vasireddy99:
**Description:** 
This PR is created to include iam role and policies of lambda in resource cleaner


bryan-aguilar:
Should we move toward regex based filtering instead of relying on a prefix alone?

Aneurysm9:
> Should we move toward regex based filtering instead of relying on a prefix alone?

Sounds reasonable, but let's do this as a separate issue.

Aneurysm9 (approved):

PR #1630, opened 2022-11-13
Allow default config to be overridden in Dockerfile

lukestoward:
**Description:**
I've been exploring options to customise the collector config used in the AWS provided  Docker image. At present, it does not seem possible to alter the configurations baked into the Docker image, meaning you have to use either the default config or those available in the `config` directory. If this is possible, please do let me know!

One option I have found is to [build the Docker image yourself](https://github.com/aws-observability/aws-otel-collector/blob/main/docs/developers/build-docker.md) and manually insert custom configs into the `config` directory prior to building the Docker image. Said custom configs would be copied into the image and usable by overriding the `--config` option when running the image.

Another option, would be to make the default config file selectable when building the Docker image. This would enable people to build the Docker image themselves but reference a custom config to be used by default. This PR implements this option and introduces an `ARG` into the `cmd/awscollector/Dockerfile` that allows users to specify which config file should be copied into the image and used by default. This option can be used in combination with the first option described above or instead of.

### DIFF
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -3,17 +3,19 @@
 # the copy mode copies in a pre-built binary.
 ARG BUILDMODE=build
 
+# The default configuration file. A custom configuration
+# file can be used by setting the --build-arg.
+ARG CONFIGFILE=config.yaml
+
 ################################
-#	Certificate Stage      #
-#			       #
+#	Certificate Stage          #
 ################################
 FROM alpine:latest AS certs
 
 RUN apk --update add ca-certificates
 
 ################################
-#	Build Stage            #
-#			       #
+#	Build Stage                #
 ################################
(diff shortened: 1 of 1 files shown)


lukestoward:
Any chance of a review?

Aneurysm9:
Thank you for your interest in contributing to the ADOT Collector.  Unfortunately, I don't believe this contribution is in line with the design intent of the ADOT Collector container image.  A default configuration is provided mostly to ensure that someone starting a container using this image for the first time without additional configuration will not encounter an immediate abend.  

We do not expect or anticipate that users will rebuild the container image in order to change the effective configuration.  Indeed, much of the testing we perform is post-build and we would not be able to offer support to users who build their own container images.  Instead, configuration should be made available to the running container through mounted volumes, whether that is a reference to a local filesystem or something like a k8s projected volume backed by a config map.

PR #1646, opened 2022-11-17
Change Collector CD workflow to trigger on new tag push

humivo:
**Description:** Changed the ADOT Collector CD workflow to be triggered on a new tag push.  Version number and commit hash is grabbed from Github Contexts.

**Testing:** Tested in the sample workflow in my own forked repo.

**Documentation:** documentation for collector release and patch release will be updated to reflect the changes.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -14,14 +14,9 @@
 name: C/D
 
 on:        
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'the version number to release'
-        required: true
-      sha:
-        description: 'the github sha to release'
-        required: true
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+*"
     
 env:
   IMAGE_NAME: aws-otel-collector
@@ -61,7 +56,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_RELEASE_TOKEN }}
 
       - name: Download candidate
-        run: aws s3 cp "s3://aws-otel-collector-release-candidate/${{ github.event.inputs.sha }}.tar.gz" candidate.tar.gz
+        run: aws s3 cp "s3://aws-otel-collector-release-candidate/${{ github.sha }}.tar.gz" candidate.tar.gz
 
       - name: Uncompress the candidate package
         run: tar zxvf candidate.tar.gz
@@ -73,12 +68,12 @@ jobs:
           version_in_release_candidate=`cat $PACKAGING_ROOT/TESTING_VERSION`
           sha_in_candidate=`cat $PACKAGING_ROOT/GITHUB_SHA`
 
-          if [ $version_in_release != ${{ github.event.inputs.version }} ]; then
-            echo "::error::Wrong version is detected: $version_in_release != ${{ github.event.inputs.version }}"
+          if [ $version_in_release != ${{ github.ref_name }} ]; then
+            echo "::error::Wrong version is detected: $version_in_release != ${{ github.ref_name }}"
             exit 1
           fi
-          if [ $sha_in_candidate != ${{ github.event.inputs.sha }} ]; then
-            echo "::error::Wrong SHA is detected: $sha_in_candidate != ${{ github.event.inputs.sha }}"
+          if [ $sha_in_candidate != ${{ github.sha }} ]; then
+            echo "::error::Wrong SHA is detected: $sha_in_candidate != ${{ github.sha }}"
             exit 1
           fi
 
@@ -199,7 +194,7 @@ jobs:
         with:
           workflow: release SSM package
           token: ${{ secrets.REPO_WRITE_ACCESS_TOKEN }}
-          inputs: '{ "version": "${{ needs.release-checking.outputs.version }}", "sha": "${{ github.event.inputs.sha }}", "public": "true", "pkgname": "${{ env.SSM_RELEASE_PACKAGE_NAME }}" }'
+          inputs: '{ "version": "${{ needs.release-checking.outputs.version }}", "sha": "${{ github.sha }}", "public": "true", "pkgname": "${{ env.SSM_RELEASE_PACKAGE_NAME }}" }'
 
   release-to-github:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


vasireddy99:
LGTM !, documentation is the key 🗝️. 😅

.github/workflows/CD.yml
```diff
@@ -14,14 +14,9 @@
...
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'the version number to release'
-        required: true
-      sha:
-        description: 'the github sha to release'
-        required: true
+  push:
+    tags:
```
bryan-aguilar, line 18: We also need to add a regex to ignore lambda tags

bryan-aguilar (changes requested):

.github/workflows/CD.yml
bryan-aguilar: Nvm, I got myself confused. This will only execute with tags pushed for this regex correct?

.github/workflows/CD.yml
```diff
@@ -14,14 +14,9 @@
...
+      - "v[0-9]+.[0-9]+.[0-9]+*"
```
bryan-aguilar, line 19:
```suggestion
      - "?v[0-9]+.[0-9]+.[0-9]+*"
```
Do we need this to indicate start of string?

.github/workflows/CD.yml
humivo: Yes this should only execute for tags pushed matching this pattern

vasireddy99:
Just wondering, if the workflow has to re-triggered for some reason. The tag needs to be deleted and pushed again?

.github/workflows/CD.yml
humivo: Hmmm I'm not sure but this is the pattern than upstream repositories use

.github/workflows/CD.yml
bryan-aguilar: Does `pkg/v1.23.0/test` trigger this regex?

.github/workflows/CD.yml
bryan-aguilar: Nvm I think I'm off base here since this is glob matching and not regex. See [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) cheat sheet for ref.

.github/workflows/CD.yml
bryan-aguilar: Disregard the last comment.

humivo: Just tested on my fork and it did not trigger on that regex.  Is that the intended behavior?

.github/workflows/CD.yml
bryan-aguilar, line 19:
```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+"
```
Do we have a need for the `*` wildcard? Currently we do not publish tags other than `v1.23.0`.

Aneurysm9:
> Just wondering, if the workflow has to re-triggered for some reason. The tag needs to be deleted and pushed again?

No.  We should never delete tags once they reach this repo.  We need to assume that someone has already pulled and is now relying on them.  If the workflow needs to be retriggered we need an additional mechanism to do so.

.github/workflows/CD.yml
humivo: If that's the case, then I think it's safe to remove the wildcard

Aneurysm9: I'm pretty sure that's the only pattern we want to match.  We will publish additional tags, but they'd have additional content at the front, not the end, and we don't want them to trigger a release.

Aneurysm9 (approved):
Good once the tag regex is fixed.

bryan-aguilar:
@humivo and I had briefly talked about a scenario like this. Is there a requirement to have a manual escape hatch? Currently the workflow file that is used will be the one that is contained in the tag commit. This means that with this method we could not use an update workflow file with an older commit. 

From my experience combining a `workflow_dispatch` with inputs with `push` triggers can get messy. The `github.event.input` context is not available on the `push` triggered workflow. If a manual CD escape hatch is desired we may be best splitting CD up into three (maybe two) workflows. Something like:
1. `CD-Core` has the core CD logic, called by other workflows
2. `CD-Tag` run on tag push, will call `CD-Core`. 
3. `CD-Dispatch` - manual dispatch, will call `CD-Core`. Maybe not necessary.

bryan-aguilar (commented):
Do updates to `GITHUB_ENV` persist through different jobs? The [documentation](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) relating to that seem to indicate that it is not persisted. Would it be safer to use `GITHUB_OUTPUT` instead?
```
The path on the runner to the file that sets environment variables from workflow commands. This file is unique to the current step and changes for each step in a job. For example, /home/runner/work/_temp/_runner_file_commands/set_env_87406d6e-4979-4d42-98e1-3dab1f48b13a. For more information, see "[Workflow commands for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable)."
```

humivo:
> Do updates to `GITHUB_ENV` persist through different jobs? The [documentation](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) relating to that seem to indicate that it is not persisted. Would it be safer to use `GITHUB_OUTPUT` instead?
> 
> ```
> The path on the runner to the file that sets environment variables from workflow commands. This file is unique to the current step and changes for each step in a job. For example, /home/runner/work/_temp/_runner_file_commands/set_env_87406d6e-4979-4d42-98e1-3dab1f48b13a. For more information, see "[Workflow commands for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable)."
> ```

Ah I think you're right the `GITHUB_ENV` does not persist through to the other job.  I just tested using the `GITHUB_OUTPUT` instead and that seems to work better than having to set the env every job so I will update my PR accordingly

bryan-aguilar (approved):

PR #1652, opened 2022-11-22
Use different test framework branch for dev branch

rapphil:
Use different test framework branch for dev branch

Signed-off-by: Raphael Silva <rapphil@gmail.com>

### DIFF
--- a/.github/workflows/CI-Operator.yml
+++ b/.github/workflows/CI-Operator.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Set testRef output
         id: setRef
         run: |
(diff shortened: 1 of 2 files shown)


Aneurysm9 (approved):

bryan-aguilar:
I believe we can do all these changes the workflow file on the `dev` branch and it would pick it up. Do we want to make these changes the workflow file on the main branch. The only change we would have to make to the main branch workflow file would be to remove the `dev` branch from the list of targets. I think this is how it would work.

bryan-aguilar (approved):

PR #1678, opened 2022-12-01
Update terraform cleanup steps on cancel

bryan-aguilar:
**Description:** Update terraform cleanup steps to use bash with no input args. [Default shell](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference) uses `-e` which will cause the run command to exit early if the `kill` command fails. 


bryan-aguilar:
Marking as draft until current CI run is done.

Aneurysm9 (approved):

PR #1745, opened 2023-01-06
[release/v0.25.x] run pre-build checks against all go source files

bryan-aguilar:
**Description:** Lambda packages were not updated to use the latest packages. This was not caught because the makefile did not run the test during PR build. This PR updates the methods used and also adjusts the makefile to correctly test all source files on `make test`. 

To ensure that all intra repository go modules are correctly tested a common makefile was added that utilizes delegation targets. This is a mirror of the upstream repository implementation. 

When running the linter/static checker against the new set of targets a few new errors were triggered. These errors have been corrected in this PR. 


Aneurysm9 (approved):

Makefile
```diff
@@ -228,17 +209,17 @@ multimod-tags: multimod-verify
 
 .PHONY: install-tools
 install-tools:
-	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install golang.org/x/tools/cmd/goimports
-	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install go.opentelemetry.io/build-tools/multimod
-	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install honnef.co/go/tools/cmd/staticcheck
-	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install github.com/golangci/golangci-lint/cmd/golangci-lint
-	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install mvdan.cc/sh/v3/cmd/shfmt
-	cd $(TOOLS_MOD_DIR) && GOBIN=$(TOOLS_BIN_DIR) go install go.opentelemetry.io/build-tools/dbotconf
```
Aneurysm9, line 236: I'm not sure we want to make this change.  This will result in using system-install versions of the tools rather than installing them in `.tools` and using from there.  Installing them locally ensures that we're using the desired version of the tools.

bryan-aguilar: I may have a solution that utilizes a local tools installation though. Should be best of both worlds.

PR #1764, opened 2023-01-17
Support version upgrade for MSI packages

Kausik-A:
**Description:** 
Previously installing a new version of ADOT collector, did not automatically uninstall the older version. 
This PR adds support version upgrade for MSI build packages. 

**Testing:**  Tested it locally in a windows build system


tools/packaging/windows/aws-otel-collector.wxs
```diff
@@ -20,6 +20,12 @@ SummaryCodepage='1252'
 InstallScope="perMachine"
 Platform="x64"
     />
+
+    <MajorUpgrade
+    AllowDowngrades="no" DowngradeErrorMessage="A later version of ADOT Collectot is already installed. Setup will now exit"
```
vasireddy99, line 25: Small nit - ADOT Collector

tools/packaging/windows/aws-otel-collector.wxs
```diff
@@ -20,6 +20,12 @@ SummaryCodepage='1252'
...
+    AllowSameVersionUpgrades="yes"
```
vasireddy99, line 26: Do we have to set this to no. As we don't usually look after the fourth version field and may not be useful setting.

'''
This is useful when two product versions differ only in the fourth version field. MSI specifically ignores that field when comparing product versions, so two products that differ only in the fourth version field are the same product and need this attribute set to yes to be detected.
'''

tools/packaging/windows/aws-otel-collector.wxs
Kausik-A: fixed

tools/packaging/windows/aws-otel-collector.wxs
```diff
@@ -20,6 +20,12 @@ SummaryCodepage='1252'
 InstallScope="perMachine"
 Platform="x64"
     />
+
+    <MajorUpgrade
+    AllowDowngrades="no" DowngradeErrorMessage="A later version of ADOT Collector is already installed. Setup will now exit"
+    AllowSameVersionUpgrades="yes"
```
bryan-aguilar, line 26: After reading the docs for this field [here](https://wixtoolset.org/docs/v3/xsd/wix/majorupgrade/), I am wondering if this is the correct solution. I do not think we should be in a situation where we need to perform upgrades for the same version. We also do not use the fourth version field. Would a viable solution, which also my be required anyways, be to ensure that [this](https://github.com/aws-observability/aws-otel-collector/blob/main/tools/packaging/windows/aws-otel-collector.wxs#L6) version field is upgraded before each collector release? 

I think a time when this COULD happen would be if the end user was building their own windows packages. This would cause the same version to be different. This should never be true for our official released assets MSIs though.

bryan-aguilar (commented):
Can you please provide a bit more elaboration on the testing that was done manually. Preferably with a list of steps and desired outcomes.  What was the problem exactly? Was the problem replicated first? What was the desired outcome? Was the desired outcome reached after this change was made?

tools/packaging/windows/aws-otel-collector.wxs
bryan-aguilar: I also think [this](https://wixtoolset.org/docs/v3/howtos/updates/major_upgrade/) doc could be relevant here.

Kausik-A: Yeah, that is a good point. My initial thinking after reading the [docs](https://wixtoolset.org/docs/v3/howtos/updates/major_upgrade/) was on updating release version after each CD trigger. But it looks to be tricky as this needs to be passed as a env variable while running this [file](https://github.com/aws-observability/aws-otel-collector/blob/main/tools/packaging/windows/create_msi.ps1). We use the cached msi packages from successful CI run which are then used by our CD workflow. So one possible solution is to rebuild the MSI package during our CD workflow to make sure we have access to the version parameter. 

This approach needs a bit of refactor, so I temporary set `AllowSameVersionUpgrades="yes"` to make sure customers can upgrade the collector even-though all the release versions are tagged at `0.0.1`.

Kausik-A: @vasireddy99 looping you in for the comment you left above

bryan-aguilar: I think updating the version should be a step in the release process before triggering the CD. It may have to be done manually or the [`bump-version-create-release-note`](https://github.com/aws-observability/aws-otel-collector/blob/main/tools/release/bump-version-and-create-release-note.sh) script could be updated. I created an [issue](https://github.com/aws-observability/aws-otel-collector/issues/1765) here to track this requirement.

Issue #1767, opened 2023-01-19
[windows][CI/CD] ADOT collector delayed start


Kausik-A:
**Describe the bug**
Sets ADOT collector agent as Automatic (delayed start) services to mitigate known go windows issues with 1.9.2: https://github.com/golang/go/issues/23479

Services would not restart across reboots on Automatic services, they would timeout before coming up and the service control manager would give up spawning them. This workaround will be reverted back once's it is believed to addressed in the future golang releases

Kausik-A:
Updates:

ADOT collector agent has been crashing sometimes at windows system reboot. This behaviour has been experienced by other projects and it seems to be a bug in golang. The issue is currently being tracked https://github.com/golang/go/issues/23479

Testing

This issue has been tough to replicate as it is a probabilistic failure which occurs to some windows instances from a large fleet of machines. This limitation on not able replicate the issue every-time has been mentioned on the linked issue too. SO we need to come up with a testing strategy to set up a mock env with these condition to run across these failures.

Proposed Solution

Out of all proposed workaround mentioned in the issue thread, one approach found that having delayed start to the application has decreased the probability of the agent failing to start up (Still not 0% but better than before ).This is implemented in #1788 .  Considering this change just delays the start of the ADOT collector, it has a comparatively low blast radius/affect on our customer.
Also additionally have some more investigation into other possible solutions.

javasorn:
I using a  WinSW to handling service ADOTCollector service. Its working in windows-aws-ecs-fargate.

https://github.com/aws-observability/aws-otel-collector/pull/1788#issuecomment-1428855205

PR #1788, opened 2023-01-27
[windows][CI/CD] ADOT collector delayed start

Kausik-A:
**Description:** 

Sets ADOT collector agent as Automatic (delayed start) services to mitigate known go windows issues with 1.9.2: https://github.com/golang/go/issues/23479

Services would not restart across reboots on Automatic services, they would timeout before coming up and the service control manager would give up spawning them. 

**Link to tracking Issue:** https://github.com/aws-observability/aws-otel-collector/issues/1767

### DIFF
--- a/tools/packaging/windows/aws-otel-collector.wxs
+++ b/tools/packaging/windows/aws-otel-collector.wxs
@@ -64,6 +64,7 @@ Account="LocalSystem"
 ErrorControl="normal"
 Arguments=" --config=&quot;[INSTALLDIR]config.yaml&quot;"
 Interactive="no">
+<ServiceConfig DelayedAutoStart="yes" OnInstall="yes" OnReinstall ="yes" />
     </ServiceInstall>
     <ServiceControl
 Id="StartService"


Kausik-A:
Updates:

ADOT collector agent has been crashing sometimes at windows system reboot. This behaviour has been experienced by other projects and it seems to be a bug in golang. The issue is currently being tracked [here ](https://github.com/golang/go/issues/23479)

Testing

This issue has been tough to replicate as it is a probabilistic failure which occurs to some windows instances from a large fleet of machines. This limitation on not able replicate the issue every-time has been mentioned on the linked issue too. SO we need to come up with a testing strategy to set up a mock env with these condition to run across these failures.

Proposed Solution

Out of all proposed workaround mentioned in the issue thread, one approach found that having delayed start to the application has decreased the probability of the agent failing to start up (Still not 0% but better than before ). Considering this change just delays the start of the ADOT collector, it has a comparatively low blast radius/affect on our customer.  
Also additionally have some more investigation into other possible solutions.

bryan-aguilar (changes requested):
Based on the comment on the linked issue it appears that more investigation/testing is required on whether this proposed fix is the best path forward.

DevOpsFu:
FWIW, I have seen the same issue running the OTel collector contrib build on Windows (also packaged up and configured on the target system with WiX). 

We already run this with the `Automatic (Delayed Start)` startup type, and we still see this issue on reboots, particularly after Windows updates. So unfortunately this probably won't mitigate the issue.

bryan-aguilar:
@DevOpsFu thanks for the insight! Do any other mitigations for lessen the effect? We are still exploring solutions around this problem and any feedback would help.

DevOpsFu:
> @DevOpsFu thanks for the insight! Do any other mitigations for lessen the effect? We are still exploring solutions around this problem and any feedback would help.

I've not yet found anything that would mitigate the issue. At least nothing that isn't a horrible hack and I wouldn't consider using, e.g.:

- Setting the service to manual startup and starting it WAY after system boot to give the system time to settle down first.
- Tweaking the [ServicesPipeTimeout](https://learn.microsoft.com/en-us/troubleshoot/windows-server/system-management-components/service-not-start-events-7000-7011-time-out-error) system registry value.

Even the Windows service recovery options to automatically restart the service on failure do not work - presumably because the service needs to at least start up successfully once and then die for the recovery measures to kick in, and this isn't happening here.

Here are some useful links though:

1. A good summary of what might be the root cause: https://github.com/shirou/gopsutil/issues/570#issue-350712862
2. An example of how the code might be refactored to move the loading of DLLs into an `init` function rather than global scope: https://github.com/shirou/gopsutil/issues/570#issuecomment-413874069
3. The core golang issue for this bug: https://github.com/golang/go/issues/23479

I've not yet delved deeply into the Otel collector code (I'm not a seasoned Go developer either), but it seems to make sense to start by looking for any areas around the Windows-specific code that might be loading DLLs outside of a function declaration. I had a very quick look earlier today and couldn't find anything using GitHub's search functions. Maybe my searches were too specific though - or perhaps this isn't the issue at all. 

From my limited understanding of the root cause though, the problem seems to be that something is making startup take a long time on first invocation of the executable. Normally this isn't a problem and probably is never visible on a fast system, but if you have the perfect storm of a system that has:

- Fewer resources to start with
- Just been rebooted after applying a load of OS updates, causing a load of I/O contention

Then it creates the perfect set of conditions for the startup to take longer than 30s, and then the Windows SCM considers the service to be dead.

DevOpsFu:
Having done some more digging into this tonight, it seems to me that it might not be possible to solve this one within the OTel collector itself without changing the service startup code. [This issue](https://github.com/grafana/agent/issues/2060) describes the problem really nicely.

A viable workaround which would not require any hacks is to remove the responsibility of communicating with the Windows SCM from the OTel collector entirely - i.e. run it in interactive mode via a service wrapper such as NSSM or [WinSW](https://github.com/winsw/winsw).  

I prefer WinSW myself, and it lends itself well to being packaged up and installed by something like WiX. Rather than using the built in `install` command to register the service, you can continue to set it up using a `<ServiceInstall>` block like you're already doing. In the associated XML file for the service, make sure to set the `NO_WINDOWS_SERVICE` environment variable to something nonzero so that the collector will definitely not start in service mode.

An example WinSW XML file might look like this:

```
<service>
  <id>otel-collector</id>
  <name>Open Telemetry Collector Service</name>
  <description>Some interesting description</description>
  <env name="NO_WINDOWS_SERVICE" value="1" />
  <logpath>%BASE%\logs</logpath>
  <log mode="append" />

  <executable>%BASE%\otelcol.exe</executable>
  <arguments>
    --config="%BASE%\otel-config.yaml"
  </arguments>
</service>
```


Hope this helps!

DevOpsFu:
Sorry for spamming this issue (plus sorry for not making these comments against the issue rather than this PR, I only just realised! 🤦 )

I tested the OTel collector on a Windows system this morning under heavy CPU load. Using the binary directly as a service yielded the issue with the service failing to start up quickly enough for the Windows SCM. 

Testing the OTel collector running under WinSW was successful; the service responded to the Windows SCM quickly and the service status was shown as running. The OTel collector itself took about 7 minutes to fully start up in the background (this was on a system with the CPU load at a constant 100%). Right now, in the absence of any solution in the initialization code in the OTel collector, I'd say that this is the best mitigation for this issue.

bryan-aguilar:
@DevOpsFu Thank you very very much for sharing your learning with us. I know we appreciate this and will definitely be working of your initial research as we look for the correct long term solution. I have had the initialization shared with me from other team members and it is something that we are looking at. 

We do have control over that in the ADOT distribution but the effort required to implement it is not clear to me. In the prometheus example the implementation was pretty straight forward so I think we could try to take a stab at that. I'm hoping with a significant amount testing we can reduce this issue. The good thing about the initialization fix proposal is that this would be something that we could contribute back to upstream if it works out well for the ADOT distribution.

Issue #1636, opened 2022-11-14
Re-enable containerinsight_ecs_prometheus test case


straussb:
In https://github.com/aws-observability/aws-otel-collector/pull/653/ the `containerinsight_ecs_prometheus` test case was removed pending https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5373 .  A workaround has now been suggested there.  ADOT developers should evaluate whether the workaround is viable and implement if so, and re-enable the test case as well.

vasireddy99:
Thanks for creating this issue to track and providing more info. The workaround suggested seems to be valid, but may not be viable for ADOT to add the test case, hence will be tracking towards the fix in upstream.

PaurushGarg:
The relevant PR (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/21680) of this scope has been merged to `opentelemetry-collector-contrib` and the feature shall be available with the next release of `contrib` and the corresponding release of ADOT Collector.

PaurushGarg:
Closing this issue as the `ecs_observer` testcase has been enabled (https://github.com/aws-observability/aws-otel-collector/pull/2227) and the relevant PR (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/21680) for the ecs_observer fix has been merged.

PR #1811, opened 2023-02-03
Modify Dockerfile to include target file.

vasireddy99:
**Description:**
This PR is created as a workaround for [ecsobserver extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/observer/ecsobserver) to auto-discover Prometheus targets. More details here :  [#5373](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5373). By creating the file `ecs_sd_targets.yaml` at this path, the collector can auto-discover Prometheus targets and scrape them.

**Link to tracking Issue:** #1636 

**Testing:** 

This PR is dependent and to be merged after https://github.com/aws-observability/aws-otel-test-framework/pull/1069

### DIFF
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -11,6 +11,7 @@ FROM alpine:latest AS certs
 
 RUN apk --update add ca-certificates
 
+# Create ecs_sd_targets.yaml for prometheus receiver to use ecsoberver extension
 FROM alpine:latest AS ecstarget
 RUN touch ecs_sd_targets.yaml
 
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -11,6 +11,9 @@ FROM alpine:latest AS certs
 
 RUN apk --update add ca-certificates
 
+FROM alpine:latest AS ecstarget
+RUN touch ecs_sd_targets.yaml
+
 ################################
 #	Build Stage            #
 #			       #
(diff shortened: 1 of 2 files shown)


bryan-aguilar:
Can this be done upstream instead of in our distribution? This appears to be an issue that would be faced by all uses of the processor. I believe @Kausik-A was looking at something similar at one point also.

cmd/awscollector/Dockerfile
```diff
@@ -11,6 +11,9 @@ FROM alpine:latest AS certs
 
 RUN apk --update add ca-certificates
 
+FROM alpine:latest AS ecstarget
+RUN touch ecs_sd_targets.yaml
```
bryan-aguilar, line 15: Can we just have a blank yaml file committed to the repo that can be copied in the copy stage?

e2etest/testcases.json
```diff
@@ -47,6 +47,10 @@
 		"case_name": "otlp_trace_resourcedetection_ecs",
 		"platforms": ["ECS"]
 	},
+	{
+		"case_name": "containerinsight_ecs_prometheus",
+		"platforms": ["ECS"]
```
bryan-aguilar, line 52: Please ensure this change is also made in the `testcases.json` file in the test-framework

bryan-aguilar:
Should we merge this to dev branch first before committing it to main?

cmd/awscollector/Dockerfile
```diff
@@ -72,6 +76,7 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 COPY --from=package /workspace/awscollector /awscollector
 COPY --from=package /workspace/config/ /etc/
 COPY --from=package /workspace/healthcheck /healthcheck
+COPY --from=ecstarget ecs_sd_targets.yaml /etc/
```
rapphil, line 79: this same workaround could be implemented directly in the ecsobserver extension with the advantage that it would be future proof. This workaround is fragile and will only work if customers use this exact location for the sd_targets file.

Please consider fixing this upstream by creating an empty file based on the configuration.
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/observer/ecsobserver/extension.go#L42

rapphil (changes requested):
I don't think this should be merged in main given that this workaround is fragile and it will mess with the final docker image that we publish.

PR #1846, opened 2023-02-14
Add support to remote uris in service control scripts in Windows and Linux

rapphil:
**Description:** It is now possible to pass remote uris such as http, https and s3 to the control scripts in Windows an Linux.

**Testing:** Tested changes locally

### DIFF
--- a/tools/ctl/linux/aws-otel-collector-ctl
+++ b/tools/ctl/linux/aws-otel-collector-ctl
@@ -44,7 +44,7 @@ UsageString="
   status:                                 get the status of the agent process.
 
   -c: configuration
-  <config-uri>:                           - file path on the host. E.g.: /tmp/config.yaml
+  <config-uri>:                           - file path on the host. E.g.: /tmp/config.yaml or file:/tmp/config.yaml
                                           - http uri. E.g.: http://example.com/config
                                           - https uri. E.g.: https://example.com/config
                                           - s3 uri. E.g.: s3://bucket/config
@@ -59,6 +59,9 @@ aoc_config_remote_uri() {
 
 aoc_config_local_uri() {
     config="${1:-}"
+
+    # Strip the file scheme in case it is present
+    config="${config#file:}"
 
     echo "config=\"--config /opt/aws/aws-otel-collector/etc/config.yaml\"" > $ENV_FILE
 
--- a/tools/ctl/linux/aws-otel-collector-ctl
+++ b/tools/ctl/linux/aws-otel-collector-ctl
@@ -19,16 +19,16 @@ set -u
 readonly AGENTDIR="/opt/aws/aws-otel-collector"
 readonly CMDDIR="${AGENTDIR}/bin"
 readonly CONFDIR="${AGENTDIR}/etc"
+readonly ENV_FILE="${CONFDIR}/.env"
 readonly DFT_CONFDIR="${AGENTDIR}/var"
 readonly LOGDIR="${AGENTDIR}/logs"
-readonly RESTART_FILE="${CONFDIR}/restart"
 readonly VERSION_FILE="${CMDDIR}/VERSION"
 
 SYSTEMD='false'
 
 UsageString="
 
-  usage: aws-otel-collector-ctl -a stop|start|status| [-c file:<file-path>]
+  usage: aws-otel-collector-ctl -a stop|start|status| [-c <config-uri>]
 
   e.g.
   1. start collector on onPermise host with a custom .yaml config file:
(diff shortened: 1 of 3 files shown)


tools/ctl/linux/aws-otel-collector-ctl
```diff
@@ -19,16 +19,16 @@ set -u
 readonly AGENTDIR="/opt/aws/aws-otel-collector"
 readonly CMDDIR="${AGENTDIR}/bin"
 readonly CONFDIR="${AGENTDIR}/etc"
+readonly ENV_FILE="${CONFDIR}/.env"
 readonly DFT_CONFDIR="${AGENTDIR}/var"
 readonly LOGDIR="${AGENTDIR}/logs"
-readonly RESTART_FILE="${CONFDIR}/restart"
```
bryan-aguilar, line 24: Why was this removed?

tools/ctl/linux/aws-otel-collector-ctl
```diff
@@ -44,20 +44,41 @@ UsageString="
   status:                                 get the status of the agent process.
 
   -c: configuration
-  file:<file-path>:                       file path on the host
+  <config-uri>:                           - file path on the host. E.g.: /tmp/config.yaml
```
bryan-aguilar, line 47:
```suggestion
  <config-uri>:                           - file path on the host. E.g.: file:/tmp/config.yaml
```
Is the conf map provider required for file uris still?

tools/ctl/linux/aws-otel-collector-ctl
```diff
@@ -44,20 +44,41 @@ UsageString="
...
     fi
+}
+
+aoc_start() {
+    config="${1:-}"
+
+    if [[ $config =~ ^(https|http|s3)://.* ]]; then
+        aoc_config_remote_uri $config
+    else
+        aoc_config_local_uri $config
```
bryan-aguilar, line 80: Why not use the `file:` [config map provider](https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/provider/fileprovider/provider.go) instead of copying?

bryan-aguilar (commented):
This may be scope creeping this PR but how can end users provide multiple configurations to the collector with this implementation?

tools/ctl/linux/aws-otel-collector-ctl
rapphil: because I don't want to change the previous behavior. customers might be copying files themselves directly to that location and just restarting the service.

tools/ctl/linux/aws-otel-collector-ctl
rapphil: it was unused.

tools/ctl/linux/aws-otel-collector-ctl
rapphil: no. that is optional.

rapphil:
> This may be scope creeping this PR but how can end users provide multiple configurations to the collector with this implementation?

@bryan-aguilar we don't support multiple configs today anywhere. I think this can be a separated issue no?

bryan-aguilar:
> > This may be scope creeping this PR but how can end users provide multiple configurations to the collector with this implementation?
> 
> @bryan-aguilar we don't support multiple configs today anywhere. I think this can be a separated issue no?

I think this is true for everything but container images. I think you could provide multiple configs to a container image command.

tools/ctl/linux/aws-otel-collector-ctl
```diff
@@ -44,20 +44,41 @@ UsageString="
...
+                                          - https uri. E.g.: https://example.com/config
+                                          - s3 uri. E.g.: s3://bucket/config
 
   "
 
-aoc_start() {
+aoc_config_remote_uri() {
     config="${1:-}"
 
+    echo "config=\"--config ${config}\"" > $ENV_FILE
```
Aneurysm9, line 57: This will overwrite the entire `$ENV_FILE`.  Is that desirable?

tools/ctl/linux/aws-otel-collector-ctl
```diff
@@ -44,20 +44,41 @@ UsageString="
...
 
     if [ ! -f $CONFDIR/config.yaml ]; then
         cp $DFT_CONFDIR/.config.yaml $CONFDIR/config.yaml
     fi
+}
+
+aoc_start() {
+    config="${1:-}"
+
+    if [[ $config =~ ^(https|http|s3)://.* ]]; then
```
Aneurysm9, line 77: This seems brittle.  We'll have to change this regex every time a new config provider is added.

tools/ctl/linux/aws-otel-collector-ctl
vasireddy99: can we use >> instead?

Aneurysm9: Appending will result in an ever-growing file setting the same variable many times.  Either this value shouldn't come from that file or this will need logic to detect whether the value is already being set in the file and react accordingly.

rapphil: > This will overwrite the entire $ENV_FILE. Is that desirable?

Yes, that is the intention.

tools/ctl/linux/aws-otel-collector-ctl
rapphil: I will replace by a more relaxed regex.

bryan-aguilar (approved):

PR #1911, opened 2023-03-16
Create default components after initializing feature gates

bryan-aguilar:
**Description:** This PR enables the usage of feature gates to block the initialization of components. This PR is an in place change and does not add or subtract any functionality. The ordering of initialization steps is adjusted so that the feature gate global registry is created before default components are initialized. 

This allows feature gates to be added in the `components` package. Components are moved into a separate list so that they can be appended to conditionally based on feature gates. This needs to happen before a factory map is created. 


cmd/awscollector/main.go
```diff
@@ -17,6 +17,7 @@ package main // import "aws-observability.io/collector/cmd/awscollector"
 
 import (
 	"fmt"
+	"github.com/aws-observability/aws-otel-collector/pkg/defaultcomponents"
```
Aneurysm9, line 20: This should go back where it came from.  Imports should generally be in three groups: stdlib, other sources, packages from this repo.

bryan-aguilar, line 20: Is there a reason goland would automatically move this import in this order? This was done automatically and not by me.

pkg/defaultcomponents/defaults.go
```diff
@@ -68,21 +68,24 @@ import (
 func Components() (otelcol.Factories, error) {
 	var errs error
 
-	extensions, err := extension.MakeFactoryMap(
+	var extensionsList = []extension.Factory{
```
Aneurysm9, line 71:
```suggestion
	extensionsList := []extension.Factory{
```

Any reason not to do short inline declaration here (and the other places in this file)?

pkg/defaultcomponents/defaults.go
```diff
@@ -68,21 +68,24 @@ import (
...
 		ecsobserver.NewFactory(),
 		healthcheckextension.NewFactory(),
 		pprofextension.NewFactory(),
 		sigv4authextension.NewFactory(),
 		zpagesextension.NewFactory(),
 		ballastextension.NewFactory(),
+	}
+	extensions, err := extension.MakeFactoryMap(
+		extensionsList...,
 	)
```
Aneurysm9, line 82: Why is this on multiple lines but not the other instances of the same pattern?

bryan-aguilar: fixed

cmd/awscollector/main.go
bryan-aguilar: https://github.com/aws-observability/aws-otel-collector/pull/1911#discussion_r1139301364

cmd/awscollector/main.go
```diff
@@ -51,12 +51,6 @@ func main() {
 	}
 
 	logger.SetupErrorLogger()
-
-	factories, err := defaultcomponents.Components()
```
rapphil, line 55: I'm almost sure it is not relevant, but this code was always executed as root, while by moving it here, it might be executed as the aoc user https://github.com/aws-observability/aws-otel-collector/pull/1911/files#diff-54474a2e8bb6289b7148377c1a16d0ba92e5347637414fbf582cffdc0b0b9e88R115

erichsueh3 (approved):

cmd/awscollector/main.go
bryan-aguilar: I can't think of a reason why this would be an issue. Constructing the factory map does no require elevated privileges.

PR #1921, opened 2023-03-21
Create flagset and config provider for windows svc

bryan-aguilar:
**Description:** Ensure that we use our distributions config provider before creating the windows service. This ensure that we use our distributions set of defined config map provider settings. 


cmd/awscollector/main_windows.go
```diff
@@ -44,6 +44,8 @@ func run(params otelcol.CollectorSettings) error {
 
 func runService(params otelcol.CollectorSettings) error {
 	// do not need to supply service name when startup is invoked through Service Control Manager directly
+	flagSet := config.Flags(featuregate.GlobalRegistry())
```
Aneurysm9, line 47: Can this be hoisted even further into `main()` so that `params.ConfigProvider` has the correct value from the start?

bryan-aguilar: That's on the TODO list but I wasn't sure how safe it would be to do that. I would prefer to do this incrementally while we also look at further improvements to the collector startup sequence.

rapphil (approved):

PR #1961, opened 2023-04-10
[chore] Update go version in dockerfile

bryan-aguilar:
**Description:**  Updates go version used in dockerfile

### DIFF
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -15,7 +15,7 @@ RUN apk --update add ca-certificates
 #	Build Stage            #
 #			       #
 ################################
-FROM golang:1.18 AS prep-build
+FROM golang:1.20 AS prep-build
 
 ARG TARGETARCH
 


vasireddy99 (approved):

cmd/awscollector/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apk --update add ca-certificates
 #	Build Stage            #
 #			       #
 ################################
-FROM golang:1.18 AS prep-build
+FROM golang:1.20 AS prep-build
```
rapphil, line 18: Can we try to use that dependabot update action? It seems that it works for dockerfiles.

bryan-aguilar: We use [dbotconf](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/dbotconf) to autogenerate dependabot files. those should not be edited manually. I think we could look at dbotconf to see how easy it would be to add `Dockerfile` detection for this. It may be useful for others as well.

rapphil: It looks like it supports. https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/dbotconf/internal/mods.go#L51
can you try?

.github/dependabot.yml
```diff
@@ -19,9 +28,6 @@ updates:
     schedule:
       interval: weekly
       day: sunday
-    ignore:
-      - dependency-name: "github.com/open-telemetry/opentelemetry-collector-contrib/*"
-      - dependency-name: "go.opentelemetry.io/collector*"
```
bryan-aguilar, line 24: These were removed because they were not generated by `dbotconf`. I believe the spirit of these ignores would prevent extra noise from these PRs. I think it's acceptable to let dependabot create these PRs now. It will act as notice to that the dependencies need to be updated. Dependabot go updates are less painful now that we have the dependabot script.

cmd/awscollector/Dockerfile
bryan-aguilar: Done

PR #2009, opened 2023-04-25
[chore] Update workflow to use `setup-go` native caching. 

bryan-aguilar:
**Description:** Setup-go v4 [added native caching](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs) to it's action. This PR leverages this instead of using the cache action directly. For some of this to take effect we had to ensure that the repository was checked out before the `setup-go` action was called. 


rapphil:
right after the release is a good time to put this in place. :)

rapphil (approved):

Issue #2043, opened 2023-05-05
Warning: ADOT Collector v0.31.0 breaking change


bryan-aguilar:
The ADOT team is aware of a breaking change made in [`v0.76.3`](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.76.3) of the upstream OpenTelemetry collector-contrib package. This breaking change affects users of the `prometheus` and `prometheusremotewrite` exporter. Starting in version `v0.76.3` the `pkg.translator.prometheus.NormalizeName` feature gate is enabled by default. The enabling of this feature gate results in the exporter using full prometheus metric name normalization rather than simple normalization. To read more about metric name normalization please refer to the normalization package documentation [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/prometheus). 

The enabling of this feature gate could result in a change to the metric names exported by Collector metric pipelines using the `prometheus` receiver,`prometheusremotewrite` exporter, or `prometheus` exporter. 

The ADOT team has made the decision to return the default state of this feature gate to disabled in ADOT Collector minor versions `v0.29.x` and `v0.30.x`. This will set the prometheus and prometheusremotewrite exporter to use simple metric name normalization by default. 

Full metric name normalization can still be enabled by providing the `--feature-gates=+pkg.translator.prometheus.NormalizeName` flag at startup. 

We **highly advise** that testing is done to determine if you are affected by this change before ADOT Collector release `v0.31.0` is released. 

Please feel free to ask any questions!

bryan-aguilar:
The ADOT team is actively investigating on whether or not this should be a configurable behavior in the listed components. Please see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/21743 issue upstream on updates.

rapphil:
To enable the feature flag in ec2 instances, you have to update the `/opt/aws/aws-otel-collector/etc/.env` manually to add the `--feature-gates=+pkg.translator.prometheus.NormalizeName`

The content should look like 
```
config="--config /opt/aws/aws-otel-collector/etc/config.yaml --feature-gates=+pkg.translator.prometheus.NormalizeName" 
```

To enable the feature flag in EC2 and ECS, you can pass the `--feature-gates=+pkg.translator.prometheus.NormalizeName` as a container argument/command line parameter.

PR #2044, opened 2023-05-05
Disable pkg.translator.prometheus.NormalizeName feature gate by default

bryan-aguilar:
**Description:** Disables the `pkg.translator.prometheus.NormalizeName` feature gate by default. 
Issue #2043 


cmd/awscollector/main.go
```diff
@@ -92,6 +92,15 @@ func main() {
 // our default component list. Flags also need to be parsed before creating the config provider.
 func buildAndParseFlagSet(featgate *featuregate.Registry) (*flag.FlagSet, error) {
 	flagSet := config.Flags(featgate)
+
+	// TODO: remove after ADOT Collector v0.30.0 is released
+	if err := featgate.Set("pkg.translator.prometheus.NormalizeName", false); err != nil {
```
rapphil, line 97: how users can opt-in into using the normalized names?

bryan-aguilar: The instructions are in the linked README. Do you think we need to duplicate it here?

cmd/awscollector/main.go
```diff
@@ -92,6 +92,15 @@ func main() {
...
+		return nil, err
+	}
+	log.Printf("attn: users of the prometheus or prometheusremotewrite exporter please refer to " +
```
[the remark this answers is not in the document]
bryan-aguilar: Yes this is a good suggestion. I will update the README. Thank you!

humivo (approved):

cmd/awscollector/main.go
rapphil: I'm asking if it is possible to override this from the command line, since this is hard coded. is it?

bryan-aguilar: Yes it is possible. The feature gate flags are parsed later on in collector initialization. If a user provides `--featuregates=+pkg.translator.prometheus.NormalizeName` at startup it will be overridden.

PR #2133, opened 2023-06-12
Use Equivalent Public ECR Images

vasireddy99:
**Description:** 

Uses Official Public ECR Image for `nginx:latest` .


deployment-template/ecs/aws-otel-ec2-sidecar-deployment-cfn.yaml
```diff
@@ -131,7 +131,7 @@ Resources:
           Links:
             - aws-collector
         - Name: traffic-generator-xray
-          Image: 'curlimages/curl:latest'
+          Image: 'public.ecr.aws/itx-devops/curl:latest'
```
bryan-aguilar, line 134: Who owns `itx-devops`?

bryan-aguilar (changes requested):

deployment-template/ecs/aws-otel-ec2-sidecar-deployment-cfn.yaml
vasireddy99: Its a mirror image from the Inbound Transportation Experience team, but didn't have full context over ownership.

bryan-aguilar: I do not want to introduce a dependency to this ECR repo. If we require the `curl` image I want to depend on officially support images and tags.

vasireddy99: The is no available officially supported tags in the Public ECR. I Think this image should be fine as its well used with 50M+ downloads, if not only then only way to use it by mirroring the official image 
image

bryan-aguilar (approved):

PR #2157, opened 2023-06-20
Use kube-rbac-proxy image from Quay.io

vasireddy99:
**Description:** 
Uses Kube-rbac-proxy image from [quay.io](https://quay.io/repository/brancz/kube-rbac-proxy?tab=tags), modifies the mirroring accordingly to mirror the image from [quay api](https://access.redhat.com/documentation/en-us/red_hat_quay/3/html-single/red_hat_quay_api_guide/index#listrepotags)

**Testing**
image

Tested Locally.

### DIFF
--- a/tools/release/image-mirror/config.yaml
+++ b/tools/release/image-mirror/config.yaml
@@ -5,11 +5,6 @@ sourceRepos:
     host: quay.io
     allowed_tags:
       - v0.14.2
-      - v0.14.1
-      - v0.13.0
-      - v0.11.0
-      - v0.8.0
-      - v0.5.0
 
 targetRepos:
   - registry: public.ecr.aws/aws-observability
--- a/tools/release/image-mirror/config.yaml
+++ b/tools/release/image-mirror/config.yaml
@@ -1,8 +1,8 @@
 ## The source repositories and target repositories are one-to-one correspondence.
 sourceRepos:
-  - registry: kubebuilder
+  - registry: brancz
     name: kube-rbac-proxy
-    host: gcr.io
+    host: quay.io
     allowed_tags:
       - v0.14.2
       - v0.14.1
--- a/tools/release/image-mirror/mirror.go
+++ b/tools/release/image-mirror/mirror.go
@@ -15,8 +15,7 @@ import (
 )
 
 const (
-	ghcr                 = "ghcr.io"
-	gcr                  = "gcr.io"
+	quay				 = "quay.io"
 	defaultSleepDuration = 60 * time.Second
 )
 
@@ -24,19 +23,9 @@ var (
 	httpClient = &http.Client{Timeout: 10 * time.Second}
 )
 
-// GHCRTagsResponse contains the tags' information of the HTTP get response from ghcr.io.
-type GHCRTagsResponse struct {
-	Tags []string `json:"tags"`
-}
-
-// GCRTagsResponse contains the tags' information of the HTTP get response from gcr.io.
-type GCRTagsResponse struct {
-	Tags []string `json:"tags"`
-}
-
-// GHCRToken contains the necessary token information to get an HTTP response from ghcr.io
-type GHCRToken struct {
-	Token string
+// QuayTagsResponse contains the tags' information of the HTTP get response from quay.io.
+type QuayTagsResponse struct {
+	Tags []RepositoryTag `json:"tags"`
 }
 
 // RepositoryTag holds the individual tag for the requested repository.
@@ -82,10 +71,8 @@ func (m *mirror) work() {
 func (m *mirror) getRemoteTags() {
 	var url string
 	switch m.sourceRepo.Host {
-	case ghcr:
-		url = fmt.Sprintf("https://ghcr.io/v2/%s/tags/list", m.sourceRepositoryName())
-	case gcr:
-		url = fmt.Sprintf("https://gcr.io/v2/%s/tags/list", m.sourceRepositoryName())
+	case quay:
+		url = fmt.Sprintf("https://quay.io/api/v1/repository/%s/tag", m.sourceRepositoryName())
 	}
 
 	if err := backoff.Retry(m.getTagResponseBackoff(url), backoffSettings); err != nil {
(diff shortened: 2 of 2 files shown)


tools/release/image-mirror/config.yaml
```diff
@@ -1,8 +1,8 @@
 ## The source repositories and target repositories are one-to-one correspondence.
 sourceRepos:
-  - registry: kubebuilder
+  - registry: brancz
     name: kube-rbac-proxy
-    host: gcr.io
```
bryan-aguilar, line 5: I am worried that this may overwrite previously mirrored image tags. Should we instead create a new `sourceRepo` entry where the host is different and remove the `v0.14.2` tag from this registry entry?

tools/release/image-mirror/mirror.go
```diff
@@ -82,10 +71,8 @@ func (m *mirror) work() {
 func (m *mirror) getRemoteTags() {
 	var url string
 	switch m.sourceRepo.Host {
-	case ghcr:
-		url = fmt.Sprintf("https://ghcr.io/v2/%s/tags/list", m.sourceRepositoryName())
-	case gcr:
-		url = fmt.Sprintf("https://gcr.io/v2/%s/tags/list", m.sourceRepositoryName())
```
bryan-aguilar, line 88: Why is support for ghcr and gcr being removed?

bryan-aguilar (changes requested):

tools/release/image-mirror/mirror.go
vasireddy99: We removed the mirroring from the repository at the time of decoupling the operator releases, I think. I am not able to find the use cases for ghcr and gcr in this repo for mirroring tool, so removing the unused code.

bryan-aguilar: I don't think we need to remove this at the moment. Is there a use case where we may need to mirror an image from gcr or ghcr in the future?

tools/release/image-mirror/config.yaml
vasireddy99: Oh, this is a good point. In this case, I would prefer that we only include `v0.14.2` and remove the [rest of tags](https://github.com/aws-observability/aws-otel-collector/blob/main/tools/release/image-mirror/config.yaml#L8-L12) so it will only mirror `0.14.2` leaving the rest of them not impacted.

tools/release/image-mirror/mirror.go
bryan-aguilar: Ok

tools/release/image-mirror/mirror.go
```diff
@@ -210,9 +163,9 @@ func getSleepTime(rateLimitReset string, now time.Time) time.Duration {
 	return calculatedSleepTime
 }
 
-func tagInAllowlist(tag string, allowlist []string) bool {
+func tagInAllowlist(tag RepositoryTag, allowlist []string) bool {
```
bryan-aguilar, line 166: This change doesn't have to be made. Just pass in `tag.Name` instead.

bryan-aguilar: The point being don't change the API of a function if you don't have to. I don't think this is a good practice.

tools/release/image-mirror/mirror.go
bryan-aguilar: One final comment on this. Just because this isn't actively used doesn't make it "dead code". You are removing functionality of a tool here. If we ever needed to mirror GHCR or GCR images again the next person would have to re-add these bits that you have removed.

I am not willing to block the PR for this but I don't think it's fair to classify this as `dead code` in practice.

tools/release/image-mirror/mirror.go
vasireddy99: I Agree

tools/release/image-mirror/mirror.go
```diff
@@ -153,29 +118,17 @@ func (m *mirror) getTagResponse(url string) error {
...
-				// Check if the ghcr image is on allowlist
-				if tagInAllowlist(tag, m.sourceRepo.AllowedTags) {
-					allTags = append(allTags, RepositoryTag{
-						Name: tag,
-					})
-				}
-			}
-		case gcr:
-			var tags GCRTagsResponse
+		case quay:
```
bryan-aguilar, line 121: There is only one host here. I think this switch statement can be removed.

tools/release/image-mirror/mirror.go
```diff
@@ -82,10 +71,8 @@ func (m *mirror) work() {
 func (m *mirror) getRemoteTags() {
 	var url string
 	switch m.sourceRepo.Host {
-	case ghcr:
-		url = fmt.Sprintf("https://ghcr.io/v2/%s/tags/list", m.sourceRepositoryName())
-	case gcr:
-		url = fmt.Sprintf("https://gcr.io/v2/%s/tags/list", m.sourceRepositoryName())
+	case quay:
```
bryan-aguilar, line 74: Only one case here. This switch statement can be removed.

tools/release/image-mirror/mirror.go
```diff
@@ -15,28 +15,17 @@ import (
...
-type GCRTagsResponse struct {
-	Tags []string `json:"tags"`
-}
-
-// GHCRToken contains the necessary token information to get an HTTP response from ghcr.io
-type GHCRToken struct {
-	Token string
+// QuayTagsResponse contains the tags' information of the HTTP get response from quay.io.
+type QuayTagsResponse struct {
+	Tags []RepositoryTag `json:"tags"`
```
bryan-aguilar, line 28: Why doesn't this follow the same structure as the GHCR or GRC response structs?

vasireddy99: Yeah, the api response is different from others as it is giving the [repo tags](https://access.redhat.com/documentation/en-us/red_hat_quay/3/html-single/red_hat_quay_api_guide/index#listrepotags) which is leading to `cannot unmarshal object` error

bryan-aguilar (approved):

PR #2183, opened 2023-07-03
Update development branch from main

rapphil:
**Description:** Update development branch from main


.github/workflows/CI.yml
```diff
@@ -723,13 +748,50 @@ jobs:
           cd testing-framework/terraform
           make checkCacheHits
 
+  release-candidate:
+    runs-on: ubuntu-latest
```
bryan-aguilar, line 752: I think this should be removed right?

bryan-aguilar: We could also consider adding an additional parameter to the `if` statement so that we only run this job on `main` and `release/` branches. This may make the updating of dev branch easier. We would have to do this on main branch first.

rapphil: https://github.com/aws-observability/aws-otel-collector/pull/2184/files

bryan-aguilar (approved):

PR #2191, opened 2023-07-05
added file log receiver, cwl exporter, and storage extension

ZahidMirza95:
**Description:** Added [File Log Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver), [AWS CloudWatch Logs Exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/awscloudwatchlogsexporter), and [File Storage Extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage) to the ADOT Collector. These components are behind a feature gate that is disabled by default (i.e. Stage Alpha).

**Link to tracking Issue:**

**Testing:** Added on to one of the unit tests to make sure all these components are not included by default. Separate unit tests were created to ensure each component was added to the ADOT Collector when their corresponding feature gate is enabled. These unit tests are located pkg/defaultcomponents/defaults_test.go

**Documentation:** No documentation was added since these components are still being tested.

### DIFF
--- a/pkg/defaultcomponents/defaults.go
+++ b/pkg/defaultcomponents/defaults.go
@@ -16,6 +16,7 @@
 package defaultcomponents // import "aws-observability.io/collector/defaultcomponents
 
 import (
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/awscloudwatchlogsexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/awsemfexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/awsxrayexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/datadogexporter"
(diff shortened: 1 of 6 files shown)


pkg/defaultcomponents/defaults.go
```diff
@@ -68,10 +72,21 @@ import (
 	"go.uber.org/multierr"
 )
 
+var fileLogReceiverFeatureGate = featuregate.GlobalRegistry().MustRegister("adot.filelog.enabled",
```
bryan-aguilar, line 75:
```suggestion
var fileLogReceiverFeatureGate = featuregate.GlobalRegistry().MustRegister("adot.filelog",
```
I believe putting the `.enabled` in the feature gate name is not necessary and possibly confusing. The enabling of a feature gate is defined by it's stage and also whether or not the user provides `+` or `-` at collector startup. 

I also wonder if we should consider namespacing this feature gate further. `adot.receiver.filelog`?

pkg/defaultcomponents/defaults.go
```diff
@@ -68,10 +72,21 @@ import (
...
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("Enables the File Log Receiver"))
```
bryan-aguilar, line 77:
```suggestion
	featuregate.WithRegisterDescription("Enables the File Log Receiver"))
```
nit: Enabling this feature gate does not `enable` the file log receiver. Enabling the feature gate allows the user to configure and start a collector with the file log receiver. Can we find better wording for this and the other feature gates?

pkg/defaultcomponents/defaults.go
ZahidMirza95: Sure I can namespace them like that, thanks for the suggestion.

pkg/defaultcomponents/defaults.go
ZahidMirza95: How about "Allows for the ADOT Collector to be configured and started with the File Log Receiver"?

pkg/defaultcomponents/defaults.go
ZahidMirza95: Or actually, to be consistent with how these components are named in the opentelemetry contrib repo, I'll namespace them like `adot.filelog.receiver`

ZahidMirza95: Resolving this conversation as I'm assuming my suggestion for the namespace change is fine.

Issue #2222, opened 2023-07-27
Ability to run the otel collector as non root user


massi-ang:
**Is your feature request related to a problem? Please describe.**
I would like to run the AWS OTEL collector as non root user, which is possible with the OpenTelemetry stock collector
I am trying to manually run the collector as non-root user, but the process exits with an exit code
`1` error (`2023/07/27 08:14:45 ADOT Collector version: v0.31.0` is printed on the console, and the error is accessed via `echo $?`)

**Describe the solution you'd like**
I would like to be able to run the command as non root user and also would like to be able to override the currently hardcoded path for extraconfig:
`"/opt/aws/aws-otel-collector/etc/extracfg.txt"`

**Describe alternatives you've considered**

**Additional context**

When running the stock otel collector with a non-root user I get:
```
# bin/otelcorecol_linux_arm64 --config ./examples/local/otel-config.yaml 
2023-07-27T08:02:49.963Z        info    service/telemetry.go:84 Setting up own telemetry...
...
```

A similar output is obtained from AWS otel collector when running with sudo:

```
# sudo ~/aws-otel-collector/bin/aws-otel-collector --config /tmp/config.yaml 
2023/07/27 08:17:06 ADOT Collector version: v0.31.0
2023-07-27T08:17:06.548Z        info    service/telemetry.go:81 Setting up own telemetry...
...
```

But when running aws collector without sudo, I get:

```
# ~/aws-otel-collector/bin/aws-otel-collector --config /tmp/config.yaml 
2023/07/27 08:20:02 ADOT Collector version: v0.31.0
# echo $?
1
```

## WORKAROUND

You can run the aws otel collector with whatever user, by setting the env variable `RUN_IN_CONTAINER=True`

eg:

```
# RUN_IN_CONTAINER=True ~/aws-otel-collector/bin/aws-otel-collector --config /tmp/config.yaml 
```

mhausenblas:
Thanks for creating this feature request @massi-ang and this is certainly something we have on our radar.

massi-ang:
Hi @mhausenblas, I just found a valid workaround. Maybe it is enough to add this information to the documentation without any code change

PaurushGarg:
Closing this issue as the ADOT Collector [v0.35.0](https://github.com/aws-observability/aws-otel-collector/releases/tag/v0.35.0) has been released with non-root user image.

PR #2259, opened 2023-08-09
[aoc-collector non-root user]: Removing code that changes user from root to aoc in Linux

PaurushGarg:
**Description:** <Describe what has changed.>
Removing Code that changes user from root to aoc in linux for non-containers
Haven't removed [RUN_IN_CONTAINER](https://github.com/aws-observability/aws-otel-collector/blob/main/pkg/extraconfig/container.go#L21) environment variable as it is getting used at [this](https://github.com/aws-observability/aws-otel-collector/blob/8343b25181940ee656138f42986462500f623672/pkg/logger/logger.go#L41) place as well.

**Link to tracking Issue:** <Issue number if applicable>
https://github.com/aws-observability/aws-otel-collector/issues/2222

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2260, opened 2023-08-09
[aoc-collector non-root user]: Create and use a new user for the container image

PaurushGarg:
Adding `aoc` user to docker file for dev branch as done in [this](https://github.com/aws-observability/aws-otel-collector/pull/1549) PR.
Haven't removed [RUN_IN_CONTAINER](https://github.com/aws-observability/aws-otel-collector/blob/main/pkg/extraconfig/container.go#L21) environment variable as it is getting used at [this](https://github.com/aws-observability/aws-otel-collector/blob/8343b25181940ee656138f42986462500f623672/pkg/logger/logger.go#L41) place as well.

https://github.com/aws-observability/aws-otel-collector/issues/2222

### DIFF
--- a/cmd/awscollector/Dockerfile
+++ b/cmd/awscollector/Dockerfile
@@ -4,16 +4,30 @@
 ARG BUILDMODE=build
 
 ################################
-#	Certificate Stage      #
-#			       #
+#	       Base Stage          #
+#			                   #
 ################################
-FROM alpine:latest AS certs
+FROM alpine:latest AS base
+
+ARG USERNAME=aoc
+ARG USER_UID=4317
+
+RUN addgroup \
+    -g $USER_UID \
+    $USERNAME && \
+    adduser \
+    -D \
+    -g $USERNAME \
+    -h "/home/${USERNAME}"\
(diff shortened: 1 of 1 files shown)


bryan-aguilar (approved):

rapphil:
the tests relying on containerinsights will fail if we merge without modifications in the test framework. Lets merge the ec2 pr first before merging this one, just to reduce the noise.

bryan-aguilar:
> the tests relying on containerinsights will fail if we merge without modifications in the test framework. Lets merge the ec2 pr first before merging this one, just to reduce the noise.

I actually don't mind to see a failure first on dev branch. Seeing a failure is a validation that this change and the fix will work as expected.

rapphil:
@bryan-aguilar 
>I actually don't mind to see a failure first on dev branch. Seeing a failure is a validation that this change and the fix will work as expected.
sure, but my point is that https://github.com/aws-observability/aws-otel-collector/pull/2259 will also affect tests that run in containers. If we merge them both at the same time we might not understand what was the reason for failure in containers.

PR #2328, opened 2023-09-08
Fix invocation of the perf workflow

rapphil:
**Description:** Fix invocation of the perf workflow

The [documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) was not clear that inherit would only work in the job level.


rapphil:
Closing as this will not solve the problem the way we want. we want to not block the CI on the execution of the perf tests.

PR #2418, opened 2023-10-13
Update OTel Collector to v0.86.0

vasireddy99:
**Description:** 
Update OTel collector to v0.86.0, fix the path for logs test bed.


.github/workflows/CI.yml
```diff
@@ -747,7 +747,8 @@ jobs:
       - name: run tests
         run: |
           export TEST_IMAGE=public.ecr.aws/${{ env.ECR_REPO }}:${{ needs.e2etest-preparation.outputs.version }}
-          cd test-framework/adot-testbed
+          cd testing-framework/adot-testbed
+          gradle wrapper
```
rapphil, line 751: why do you need this?

rapphil: the wrapper is already here: https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/adot-testbed/gradlew

So I'm not following why we need this step.

vasireddy99: In My local I am starting to get this

```
Error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain
Caused by: java.lang.ClassNotFoundException: org.gradle.wrapper.GradleWrapperMain
```
So have added this command here as i was thinking it could be the same case here.

rapphil (approved):

.github/workflows/CI.yml
```diff
@@ -743,11 +743,13 @@ jobs:
 
       - name: Setup Gradle
         uses: gradle/gradle-build-action@v2
+        with:
+          cache-disabled: true
```
bryan-aguilar, line 747: Why?

vasireddy99: added this changed from the work around - https://github.com/gradle/gradle-build-action/security/advisories/GHSA-h3qr-39j9-4r5v
`If the Gradle project does not opt-in to using the configuration cache, then it is not vulnerable.`

bryan-aguilar: You can re-enable the cache. We do not need to disable it globally for that alert.

PR #2735, opened 2024-05-08
[Release/v0.39.x] - Add third Party components behind featuregateBeta

vasireddy99:
**Description:** 

This PR adds third-party exporters behind the feature gate beta stage
Removes deprecated `configProvider` setting and use `configProviderSettings` .

### DIFF
--- a/pkg/config/config_factory.go
+++ b/pkg/config/config_factory.go
@@ -35,7 +35,7 @@ const (
 	envKey = "AOT_CONFIG_CONTENT"
 )
 
-func GetConfigProvider(flags *flag.FlagSet) otelcol.ConfigProvider {
+func GetConfigProviderSettings(flags *flag.FlagSet) otelcol.ConfigProviderSettings {
 	// aws-otel-collector supports loading yaml config from Env Var
 	// including SSM parameter store for ECS use case
 	loc := getConfigFlag(flags)
@@ -59,7 +59,7 @@ func GetConfigProvider(flags *flag.FlagSet) otelcol.ConfigProvider {
 	}
 
 	// create Config Provider Settings
-	settings := otelcol.ConfigProviderSettings{
+	configProviderSettings := otelcol.ConfigProviderSettings{
 		ResolverSettings: confmap.ResolverSettings{
 			URIs:       loc,
 			Providers:  mapProviders,
@@ -68,11 +68,11 @@ func GetConfigProvider(flags *flag.FlagSet) otelcol.ConfigProvider {
 	}
 
 	// get New config Provider
-	config_provider, err := otelcol.NewConfigProvider(settings)
+	_, err := otelcol.NewConfigProvider(configProviderSettings)
 
 	if err != nil {
 		log.Panicf("Err on creating Config Provider: %v\n", err)
 	}
 
-	return config_provider
+	return configProviderSettings
 }
--- a/pkg/defaultcomponents/defaults.go
+++ b/pkg/defaultcomponents/defaults.go
@@ -28,6 +28,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/sapmexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/signalfxexporter"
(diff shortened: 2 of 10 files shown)


pkg/config/config_factory.go
```diff
@@ -68,11 +68,11 @@ func GetConfigProvider(flags *flag.FlagSet) otelcol.ConfigProvider {
 	}
 
 	// get New config Provider
-	config_provider, err := otelcol.NewConfigProvider(settings)
+	_, err := otelcol.NewConfigProvider(configProviderSettings)
```
Aneurysm9, line 71: This isn't useful anymore, is it?

pkg/defaultcomponents/defaults.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/sapmexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/signalfxexporter"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/splunkhecexporter"
```
Aneurysm9, line 31: Why are we adding a new component here?  This doesn't seem right at all.

pkg/defaultcomponents/defaults.go
```diff
@@ -72,6 +74,26 @@ import (
 	"go.uber.org/multierr"
 )
 
+var datadogExporterFeatureGate = featuregate.GlobalRegistry().MustRegister("adot.exporter.datadogexporter",
+	featuregate.StageBeta,
```
Aneurysm9, line 78: Feature gates need to start at the alpha stage.  The name should probably be something like `adot.exporter.<name>.deprecation` so that the "enabled" state is removing the component.

pkg/defaultcomponents/defaults.go
```diff
@@ -141,18 +163,28 @@ func Components() (otelcol.Factories, error) {
...
-		logzioexporter.NewFactory(),
 		loggingexporter.NewFactory(),
 		otlpexporter.NewFactory(),
 		otlphttpexporter.NewFactory(),
 		awsxrayexporter.NewFactory(),
 		loadbalancingexporter.NewFactory(),
 		awscloudwatchlogsexporter.NewFactory(),
 	}
-
+	if datadogExporterFeatureGate.IsEnabled() {
```
Aneurysm9, line 173: See above about naming.  This condition should be inverted:

```suggestion
	if !datadogExporterFeatureGate.IsEnabled() {
```

so that in the alpha stage where the gate is disabled by default the component is available.

Aneurysm9 (changes requested):

pkg/defaultcomponents/defaults.go
vasireddy99: yeah, confused with previous commits in test framework, removed the change

pkg/defaultcomponents/defaults.go
vasireddy99: I was initially trying to have component by default and hense started from reverse fashion, like from beta, then alpha -> removal. instead the actual should alpha -> deprecated -> removal. 

It makes sense. Understood the idea now.

pkg/defaultcomponents/defaults.go
vasireddy99: modified the gate with `deprecation` suffix.

pkg/defaultcomponents/defaults.go
```diff
@@ -72,6 +73,22 @@ import (
 	"go.uber.org/multierr"
 )
 
+var datadogExporterFeatureGateDeprecation = featuregate.GlobalRegistry().MustRegister("adot.exporter.datadogexporter.deprecation",
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("Allows for the ADOT Collector to be configured and started with the Datadog Exporter"))
```
Aneurysm9, line 78:
```suggestion
	featuregate.WithRegisterDescription("Removes the Datadog exporter from the set of configurable exporters"))
```

This should be written describing what happens when the gate is enabled.  The same change applies to all of these descriptions.

Aneurysm9 (approved):

PR #2855, opened 2024-09-26
Add debug exporter to component factory list

Aneurysm9:
**Description:** Adds the `debugexporter` component, which will replace the `loggingexporter`, and converts in-tree configs to use `debug` instead of `logging`.


pkg/defaultcomponents/defaults.go
```diff
@@ -158,6 +159,7 @@ func Components() (otelcol.Factories, error) {
 		prometheusexporter.NewFactory(),
 		fileexporter.NewFactory(),
 		kafkaexporter.NewFactory(),
+		debugexporter.NewFactory(),
 		loggingexporter.NewFactory(),
```
alexperez52, line 163: should loggingexporter.NewFactory() be removed?

Aneurysm9: No, we want to keep that around so that existing configurations do not break.  We could introduce a config processor that would translate the configuration from one to the other, but that's a next step.

alexperez52 (approved):

PR #3071, opened 2025-07-23
Set Env and display aws version at the time of signing for visibility

vasireddy99:
**Description:** Set Env and display aws version at the time of signing for visibility


anant10 (approved):
LGTM!

PR #3139, opened 2025-12-09
[release/v0.46.x] Prepping for release 

Kausik-A:
**Description:** 

This PR upgrades OpenTelemetry Collector dependencies from v0.131.0 to v0.141.0 and updates Go to version 1.24.11 to address security vulnerabilities (GO-2025-4175, GO-2025-4155) in the crypto/x509 standard library.

Key changes:

- Upgraded all OTEL collector-contrib dependencies from v0.131.0 to v0.141.0
- Updated Go version from 1.24.5 to 1.24.11 across all go.mod files (root, testbed, tools/release/image-mirror, tools/workflow/cleaner, tools/workflow/linters)
- Removed patches/awscontainerinsightreceiver.patch (fix now included upstream)
- Updated patches/signalfxexporter.patch to align with new OTEL version
- Updated docker dependency to v28.5.2+incompatible in tools/release/image-mirror
- Minor code adjustments in pkg/extraconfig, pkg/logger, pkg/defaultcomponents, and cmd/healthcheck to accommodate API changes

**Testing:** 

- make gomod-tidy - passed
- make test - all tests passed
- govulncheck ./... - verified CVEs are resolved with Go 1.24.11

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
be the bug and how this fixes the issue.

PR #3142, opened 2025-12-10
Bump OpenTelemetry Collector and collector contrib dependencies to v0.141.0 

Kausik-A:
NOTE: Commits are ported over from draft PR #3139 as we need to raise a PR with remote branch rather than the fork branch for github actions to run with the right permissions. 

Description:

This PR upgrades OpenTelemetry Collector dependencies from v0.131.0 to v0.141.0 and updates Go to version 1.24.11 to address security vulnerabilities (GO-2025-4175, GO-2025-4155) in the crypto/x509 standard library.

Key changes:

- Upgraded all OTEL collector-contrib dependencies from v0.131.0 to v0.141.0
- Updated Go version from 1.24.5 to 1.24.11 across all go.mod files (root, testbed, tools/release/image-mirror, tools/workflow/cleaner, tools/workflow/linters)
- Removed patches/awscontainerinsightreceiver.patch (fix now included upstream)
- Updated patches/signalfxexporter.patch to align with new OTEL version
- Updated docker dependency to v28.5.2+incompatible in tools/release/image-mirror
- Minor code adjustments in pkg/extraconfig, pkg/logger, pkg/defaultcomponents, and cmd/healthcheck to accommodate API changes

Testing:

    make gomod-tidy - passed
    make test - all tests passed
    govulncheck ./... - verified CVEs are resolved with Go 1.24.11

be the bug and how this fixes the issue.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -24,6 +24,10 @@ concurrency:
   group: codeql-${{ github.ref_name }}
   cancel-in-progress: true
 
+permissions:
+  security-events: write
+  contents: read
+
 jobs:
   analyze:
     runs-on: ubuntu-22.04
--- a/.github/workflows/pr-body-validate.yml
+++ b/.github/workflows/pr-body-validate.yml
@@ -17,6 +17,9 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, edited, ready_for_review]
 
+permissions:
+  contents: read
+
 env:
   LICENSE_TEXT: By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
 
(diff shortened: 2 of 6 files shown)


pkg/extraconfig/extraconfig.go
```diff
@@ -78,7 +78,7 @@ func GetExtraConfig() (*ExtraConfig, error) {
 		case "awsCredentialFile":
 			extraConfig.AwsCredentialFile = val
 		default:
-			os.Setenv(key, val)
+			_ = os.Setenv(key, val)
```
vasireddy99: i think lint errors have started to come up after v2 bump, it makes sense for most of these, maybe we can check and return error if we find anywhere useful

PR #3187, opened 2026-05-15
ci: improve Go module caching with restore-keys and conditional skip

okankoAMZ:
- Disable setup-go built-in cache in favor of explicit actions/cache
- Add restore-keys fallback for partial cache hits on dep changes
- Skip Go cache download in build job when binary cache already hit
- Apply consistent caching pattern across all 4 Go jobs: build-aotutil, build, get-testing-suites, validate-all-tests-pass

**Description:** The `setup-go@v5` built-in cache does not support `restore-keys` (partial cache fallback) and always downloads the Go module cache even when the full binary cache has already hit. This PR replaces it with explicit `actions/cache@v3` steps that:
1. Use `restore-keys` so partial dep changes still get a warm cache instead of a cold 6GB download
2. Gate the Go cache step in the `build` job behind `cached_binaries.outputs.cache-hit != 'true'` so re-runs within the same workflow skip the unnecessary cache restore entirely

Pattern is consistent with [amazon-cloudwatch-agent/test-build.yml](https://github.com/aws/amazon-cloudwatch-agent/blob/main/.github/workflows/test-build.yml).

**Link to tracking Issue:** N/A

**Testing:** YAML validated locally. The caching behavior will be verified on the first CI run of this branch — expect cache misses on first run (populating), then hits on subsequent runs.

**Documentation:** No documentation changes needed — this is a CI-internal optimization.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -70,8 +70,8 @@ jobs:
       - name: Login Dockerhub
         uses: docker/login-action@v3
         with:
-          username: ${{ secrets.DOCKERHUB_RELEASE_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_RELEASE_TOKEN }}
+          username: ${{ secrets.DOCKERHUB_OAT_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_OAT_SECRET }}
 
       - name: Download candidate
         run: aws s3 cp "s3://aws-otel-collector-release-candidate/${{ steps.set-input-vars.outputs.sha }}.tar.gz" candidate.tar.gz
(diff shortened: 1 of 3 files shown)


.github/workflows/CD.yml
```diff
@@ -70,8 +70,8 @@ jobs:
       - name: Login Dockerhub
         uses: docker/login-action@v3
         with:
-          username: ${{ secrets.DOCKERHUB_RELEASE_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_RELEASE_TOKEN }}
+          username: ${{ secrets.DOCKERHUB_OAT_USERNAME }}
```
jefchien, line 73: Do we need to rebase release/v0.48.x on `main` or something?

.github/workflows/CI.yml
```diff
@@ -153,6 +162,18 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: ${{ env.GO_VERSION }}
+        cache: false
+
+    - name: Cache go
+      if: steps.cached_binaries.outputs.cache-hit != 'true'
```
jefchien, line 185: We're trying to use the `steps.cached_binaries.outputs.cache-hit` before the step, so this will always resolve to true. If we want to skip it on cache hit, then we need to move the `cached_binaries` step up.

.github/workflows/CD.yml
jefchien: Looks like we can't rebase, so we'll need to cherry-pick it in at some point. Fine with keeping it. Mention in the PR description that this also cherry-picks in that commit.

jefchien (approved):

PR #3191, opened 2026-05-15
fix: quote batch key for jq compatibility with platform-grouped names

TravisStark:
**Details:**
The batch generator now produces keys with slashes (e.g., EC2/amazonlinux2/3). The jq expression .${{ matrix.BatchKey }} fails on these because jq interprets / as division. 

**Fix:**
Use .["${{ matrix.BatchKey }}"] bracket notation.

Pairs with aws-observability/aws-otel-test-framework#1836 which adds the platform-grouped batch generator and log improvements.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -95,7 +95,7 @@ jobs:
       - name: Set testRef output
         id: setRef
         run: |
-          if [[ ${{ github.ref_name }} == release/v* ]]; then 
+          if [[ ${{ github.ref_name }} == release/v* ]]; then
             echo "ref=${{github.ref_name}}" >> $GITHUB_OUTPUT
           else
             echo "ref=terraform" >> $GITHUB_OUTPUT
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -687,7 +687,7 @@ jobs:
       - name: create test-case-batch file
         run: |
           jsonStr='${{ needs.get-testing-suites.outputs.test-case-batch-value }}'
-          jsonStr="$(jq -r '.${{ matrix.BatchKey }} | join("\n")' <<< "${jsonStr}")"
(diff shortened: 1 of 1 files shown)


okankoAMZ (approved):

.github/workflows/CI.yml
```diff
@@ -95,11 +95,7 @@ jobs:
       - name: Set testRef output
         id: setRef
         run: |
-          if [[ ${{ github.ref_name }} == release/v* ]]; then 
-            echo "ref=${{github.ref_name}}" >> $GITHUB_OUTPUT
-          else
-            echo "ref=terraform" >> $GITHUB_OUTPUT
-          fi
+          echo "ref=${{github.ref_name}}" >> $GITHUB_OUTPUT
```
jefchien: Don't quite agree with this change. This is going to break `main` once this merges back in.

What about something like this where we check to make sure the branch actually exists before using it?

```suggestion
        env:
          GH_TOKEN: ${{ github.token }}
          REF_NAME: ${{ github.ref_name }}
        run: |
          ref=terraform
          encoded_ref="${REF_NAME//\//%2F}"
          if [[ $REF_NAME == release/v* || $REF_NAME == test/* ]] && \
              gh api repos/${{ env.TESTING_FRAMEWORK_REPO }}/branches/$encoded_ref \
              --silent 2>/dev/null; then
            ref=$REF_NAME
          fi
          echo "ref=$ref" >> $GITHUB_OUTPUT
```

jefchien (changes requested):

.github/workflows/CI.yml
TravisStark: good catch, fixed this

jefchien (commented):
Please update the title and description before merging. This PR isn't doing any of that.

jefchien (approved):

PR #3203, opened 2026-05-20
chore: remove BAH (me-south-1) region from SSM release and rollback

TravisStark:
## Summary
Remove me-south-1 region from SSM release and rollback workflows.


alexperez52 (approved):
probably better to comment them out so we remember to add them back eventually but its okay

